PR #2, opened 2015-07-14
Develop

djmiley:


.gitignore
```diff
@@ -0,0 +1,4 @@
+doc
```
jleft: We don't have a `doc` directory, so I don't think we'll need this line.

This [.gitignore](https://github.com/github/gitignore/blob/master/Node.gitignore) is a good place to start. I expect we'll likely want to add `dist` too, but we can leave it for now.

README.md
```diff
@@ -1 +1,64 @@
-# d3fc-showcase
+# d3fc-showcase [![Build Status] travis link]
```
jleft: Remove `[![Build Status] travis link]` for now, we can add the link back in, when we add Travis and have the URL.

README.md
```diff
@@ -1 +1,64 @@
...
+
+A cross-platform financial charting application to showcase the functionality of d3fc components
+
+## Installations
```
jleft: I'd probably remove everything from here and below, except for the _License_ section, as I don't think it's applicable yet. We can always add these sections back in when we have Grunt tasks in place etc., unless there's some value in this already?

README.md
```diff
@@ -1 +1,64 @@
-# d3fc-showcase
+# d3fc-showcase [![Build Status] travis link]
+
+A cross-platform financial charting application to showcase the functionality of d3fc components
```
jleft: Link to d3fc website on d3fc text would be nice. And add full stop.

jleft:
+ [ ] I think it'd be good to remove the `visual-tests` directory, as I don't think we're likely to use this.
+ [ ] I'd also rename `tests` to `test` for consistency (singular vs. plural).

Looks good overall, just a few minor comments :+1:

jleft:
Thanks for the amends - I'll merge this for now, as it will give us a starting structure for the project and we can refine if we need in future :smiley:

PR #6, opened 2015-07-14
Feature/initial files

CBarrett2:
Added gruntfile with default and watch tasks to check the coding style using JShint and JSCS. Also added example files into src, which create a simple chart.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,24 @@
+{
+    "curly": true,
+    "eqeqeq": true,
+    "eqnull": true,
+    "forin": true,
+    "immed": true,
+    "latedef": true,
+    "newcap": true,
+    "noarg": true,
+    "strict": true,
+    "undef": true,
+    "unused": "vars",
+    "validthis": true,
+
+    "browser": true,
+    "devel": true,
+    "jasmine": true,
+
+    "globals": {
+        "d3": false,
+        "fc": false,
+        "require": false
+    }
+}
--- a/.jshintrc
+++ b/.jshintrc
@@ -18,7 +18,6 @@
 
     "globals": {
         "d3": false,
-        "fc": false,
-        "require": false
+        "fc": false
     }
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,8 +1,75 @@
-module.exports = function(grunt) {
+/* Global module */
 
-  grunt.initConfig({
-  });
+module.exports = function (grunt) {
+    'use strict';
+    grunt.initConfig({ 
+        pkg: grunt.file.readJSON('package.json'),
+        
+        meta: {
+            componentsJsFiles: [
+                'src/**/*.js'
+            ],
+            testJsFiles: [
+                'test/**/*Spec.js'
+            ],
+            ourJsFiles: [
+                '<%= meta.componentsJsFiles %>',
+                '<%= meta.testJsFiles %>',
+            ]
+        },
+    
+        jscs: {
+            options: {
+                config: '.jscsrc'
+            },
+            failOnError: {
+                files: {
+                    src: ['<%= meta.ourJsFiles %>']
+                }
+            },
+            warnOnly: {
+                options: {
+                    force: true
+                },
+                files: {
+                    src: ['<%= meta.ourJsFiles %>']
+                }
+            }
+        },
 
-  grunt.registerTask('default', []);
-
+        jshint: {
+            options: {
+                jshintrc: true
+            },
+            failOnError: {
+                files: {
+                    src: ['<%= meta.ourJsFiles %>']
+                }
+            },
+            warnOnly: {
+                options: {
+                    force: true
+                },
+                files: {
+                    src: ['<%= meta.ourJsFiles %>']
+                }
+            }
+        },
+        
+        watch: {
+            files: ['<%= meta.ourJsFiles %>'],
+            tasks: ['jshint', 'jscs']
+        }
+    });
+    
+    grunt.loadNpmTasks('grunt-contrib-jshint');
+    grunt.loadNpmTasks('grunt-jscs');
+    grunt.loadNpmTasks('grunt-contrib-watch');
+    
+    grunt.registerTask('default', ['jshint', 'jscs']);
+    grunt.registerTask('check:failOnError', ['jshint:failOnError', 'jscs:failOnError']);
+    grunt.registerTask('check:warnOnly', ['jshint:warnOnly', 'jscs:warnOnly']);
+    grunt.registerTask('check', ['check:failOnError']);
+    grunt.registerTask('check:failOnError', ['jshint:failOnError', 'jscs:failOnError']);
+    grunt.registerTask('check:warnOnly', ['jshint:warnOnly', 'jscs:warnOnly']);
 };
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -67,7 +67,7 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks('grunt-jscs');
     grunt.loadNpmTasks('grunt-contrib-watch');
 
-    grunt.registerTask('default', ['jshint', 'jscs']);
+    grunt.registerTask('default', ['check']);
     grunt.registerTask('check:failOnError', ['jshint:failOnError', 'jscs:failOnError']);
     grunt.registerTask('check:warnOnly', ['jshint:warnOnly', 'jscs:warnOnly']);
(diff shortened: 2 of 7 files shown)


.jshintrc
```diff
@@ -0,0 +1,24 @@
...
+    "validthis": true,
+
+    "browser": true,
+    "devel": true,
+    "jasmine": true,
+
+    "globals": {
+        "d3": false,
+        "fc": false,
+        "require": false
```
jleft: We're not using `require`, so we can remove this for now.

Gruntfile.js
```diff
@@ -1,8 +1,76 @@
+/* global module */
+
 module.exports = function(grunt) {
+    'use strict';
+    grunt.initConfig({
+        pkg: grunt.file.readJSON('package.json'),
+
+        meta: {
+            componentsJsFiles: [
```
jleft: I think `componentsJsFiles` is a bit misleading, as we're build an app rather than components, maybe `srcJsFiles`?

Gruntfile.js
```diff
@@ -1,8 +1,76 @@
...
+                    src: ['<%= meta.ourJsFiles %>']
+                }
+            }
+        },
 
-  grunt.initConfig({
-  });
+        watch: {
+            files: ['<%= meta.ourJsFiles %>'],
+            tasks: ['jshint', 'jscs']
```
jleft: Can we just call `default` here?

Gruntfile.js
```diff
@@ -1,8 +1,76 @@
...
-  grunt.registerTask('default', []);
+    grunt.loadNpmTasks('grunt-contrib-jshint');
+    grunt.loadNpmTasks('grunt-jscs');
+    grunt.loadNpmTasks('grunt-contrib-watch');
 
+    grunt.registerTask('default', ['jshint', 'jscs']);
+    grunt.registerTask('check:failOnError', ['jshint:failOnError', 'jscs:failOnError']);
+    grunt.registerTask('check:warnOnly', ['jshint:warnOnly', 'jscs:warnOnly']);
+    grunt.registerTask('check', ['check:failOnError']);
+    grunt.registerTask('check:failOnError', ['jshint:failOnError', 'jscs:failOnError']);
```
jleft: `check:warnOnly` and `check:failOnError` have been included twice.

src/index.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE HTML>
+<html lang="en">
+	<div class="container-fluid">
```
jleft: The `head` element shouldn't be surrounded by this `div`. `head` should be first element in the `html`, it only contains meta data about the document, so doesn't need a `div` to position/style it etc. The `head` doesn't need a class either.

http://www.w3.org/TR/html5/document-metadata.html#the-head-element

src/index.html
```diff
@@ -0,0 +1,34 @@
...
+		<head class="page-header">
+			<meta charset="UTF-8">   
+			<meta http-equiv="X-UA-Compatible" content="IE=edge"> <!--Needed for bootstrap-->
+			<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"> <!--Needed for bootstrap-->
+			<meta name="description" content="description">
+			
+			<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script> <!-- JQuery - also necessary for bootstrap -->
```
jleft: Not a big deal, but I think we're safe to remove jQuery, Bootstrap, html5shiv and respond for now? Just to make it a little less cluttered.

src/index.html
```diff
@@ -0,0 +1,34 @@
...
+<html lang="en">
+	<div class="container-fluid">
+		<head class="page-header">
+			<meta charset="UTF-8">   
+			<meta http-equiv="X-UA-Compatible" content="IE=edge"> <!--Needed for bootstrap-->
+			<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"> <!--Needed for bootstrap-->
+			<meta name="description" content="description">
+			
+			<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script> <!-- JQuery - also necessary for bootstrap --> 
+			<script src="bootstrap.min.js"></script> <!-- Bootstrap -->
```
jleft: This file doesn't exist, so I'd remove the tag.

src/index.html
```diff
@@ -0,0 +1,34 @@
...
+			<script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script> <!--These two for ie8 support of bootstrap-->
+			<script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
+			<script src="http://d3js.org/d3.v3.min.js" charset="utf-8"></script> <!-- D3! -->
```
jleft: We don't need to include D3 twice - we can just link to the `node_modules`.

src/index.html
```diff
@@ -0,0 +1,34 @@
...
+			<script src="node_modules/d3fc/node_modules/d3/d3.js"></script> <!-- These 3 for D3FC -->
```
jleft: Fix the link to the `node_modules` as they are up a level. `src="../node_modules/etc`

src/main.js
```diff
@@ -0,0 +1,39 @@
+/* A test function */
+
+// Main function
```
jleft: Could strip out the comments but very minor point.

src/main.js
```diff
@@ -0,0 +1,39 @@
...
+(function() {
```
jleft: For consistency: `(function(d3, fc) {`

src/main.js
```diff
@@ -0,0 +1,39 @@
...
+        .attr('width', width)
+        .attr('height', height);
+
+    var data = fc.dataGenerator()(150);
+
+    var chart = fc.chart.linearTimeSeries()
+        .xDomain(fc.util.extent(data, 'date'))
+        .yDomain(fc.util.extent(data, ['open', 'close']));
+
+    svg.datum(data)
```
jleft: I think this:

```js
svg.datum(data)
    .call(chart);
```

should be called last, otherwise we won't see the multi series.

src/main.js
```diff
@@ -0,0 +1,39 @@
...
+
+    var gridlines = fc.annotation.gridline()
+        .yTicks(5)
+        .xTicks(0);
+
+    var multi = fc.series.multi()
+        .series([gridlines, area, line]);
+
+    chart.plotArea(multi);
+})();
```
jleft: For consistency: `})(d3, fc);`

src/style.css
```diff
@@ -0,0 +1,14 @@
+path.area {
+  fill: url(#area-gradient);
```
jleft: I don't think `#area-gradient` is defined? We could just set a solid fill or use the default and remove this class?

jleft:
Thanks for this @CBarrett2, this looks like a really good start to me, just a few minor comments, but I think it'll be good to merge once the amends have been made!

CBarrett2:
Fixed everything you suggested :)

Gruntfile.js
```diff
@@ -1,8 +1,74 @@
...
+            tasks: ['check']
+        }
+    });
 
-  grunt.registerTask('default', []);
+    grunt.loadNpmTasks('grunt-contrib-jshint');
+    grunt.loadNpmTasks('grunt-jscs');
+    grunt.loadNpmTasks('grunt-contrib-watch');
 
+    grunt.registerTask('default', ['jshint', 'jscs']);
```
jleft: Shall we change `['jshint', 'jscs']` to `['check']`, then we won't be running both `failOnError` and `warnOnly` for both tasks?

src/index.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE HTML>
```
jleft: This file has mixed tabs and spaces - shall we go with 2 spaces for now?

src/style.css
```diff
@@ -0,0 +1,9 @@
+path.line {
+  stroke: rgb(26, 154, 249);
+}
+
+#chart-example {
+    position: absolute;
```
jleft: I think 2 space indentation would be good here too.

jleft:
> Fixed everything you suggested :)

Brilliant, thank you!

Just a couple more pedantic comments, but then this should be good to go.

CBarrett2:
Fixed :D

jleft:
> Fixed :D

Thanks :smiley:

PR #10, opened 2015-07-15
Added grunt build task 

CBarrett2:
now use 'grunt build' to build the src file to the dist file

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -59,17 +59,74 @@ module.exports = function(grunt) {
 
         watch: {
             files: ['<%= meta.ourJsFiles %>'],
-            tasks: ['check']
+            tasks: ['build']
+        },
+
+        clean: {
+            build: {
+                src: ['dist']
+            }
+        },
+
+        copy: {
+            build: {
+                files: [{
+                    cwd: 'src',
+                    src: ['*.html'],
+                    dest: 'dist',
+                    expand: true
+                },
+                {
+                    cwd: 'src',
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -59,17 +59,74 @@ module.exports = function(grunt) {
...
+            build: {
+                src: ['dist']
+            }
+        },
+
+        copy: {
+            build: {
+                files: [{
+                    cwd: 'src',
+                    src: ['*.html'],
```
jleft: Let's update `src` to have the same structure as `dist`, so `src` also has `js` and `css` sub-directories. Then we can then simplify this to copy all the files in `src` to `dist`, so we won't need a separate config for html, css and js.

Gruntfile.js
```diff
@@ -59,17 +59,74 @@ module.exports = function(grunt) {
...
+                    dest: 'dist',
+                    expand: true
+                },
+                {
+                    cwd: 'src',
+                    src: ['*.js'],
+                    dest: 'dist/assets/js',
```
jleft: `dist/assets` is repeated quite a bit, maybe we can pull this out into a variable?

Gruntfile.js
jleft: Can we also remove `placeholder.txt` from `src`?

jleft:
Ah, I was too slow! Happy to see this merged, we can always make the refinements later :smiley:

PR #12, opened 2015-07-15
fixed file structure in src

CBarrett2:


jleft:
Hi @CBarrett2, thanks for this - it all looks good. Any chance that the branch could be updated by using rebase, rather than merge, so we can get cleaner history for the project? Give me a shout if you'd like to go through it :-) thanks

CBarrett2:
Yeah, I'll try in future. Also just added another feature to this branch which really should have been in its own - sorry! Added a grunt serve task which runs builds to the dist file and runs a server where you can access the site.

PR #18, opened 2015-07-17
Added zoom plugin

CBarrett2:
zoomPlugin plugin adds zoom and pan functionality - still needs to fix scaleExtent, so you don't zoom off the chart.

### DIFF
--- a/src/assets/js/chartFramework.js
+++ b/src/assets/js/chartFramework.js
@@ -1,0 +1,58 @@
+(function(fcsc, fc, d3) {
+    'use strict';
+    fcsc.chartFramework = function() {
+        var width = 500;
+        var height = 300;
+        var timeSeries = fc.chart.linearTimeSeries();
+        var multi = fc.series.multi();
+        var plugins = [];
+
+        function chart(selection) {
+            selection.each(function(d, i) {
+                timeSeries.xDomain(fc.util.extent(d, 'date'))
+                    .yDomain(fc.util.extent(d, ['open', 'close']));
+
+                timeSeries.plotArea(multi);
+                selection.call(timeSeries);
+
+                for (var j = 0; j < plugins.length(); j++) {
+                    selection.call(plugins[j], this);
+                }
+            });
+            return chart;
+        }
+
+        chart.width = function(value) {
+            if (!arguments.length) { return width; }
+            width = value;
+            return chart;
+        };
+
+        chart.height = function(value) {
+            if (!arguments.length) { return height; }
+            height = value;
+            return chart;
+        };
+
+        chart.plugins = function(value) {
+            if (!arguments.length) { return plugins; }
+            plugins = value;
+            return chart;
+        };
+
+        d3.rebind(chart, multi, 'series');
+
+        chart.getVisibleData = function(data, dateExtent) {
+            // Calculate visible data for main/volume charts given [startDate, endDate]
+            var bisector = d3.bisector(function(d) { return d.date; });
+            var visibleData = data.slice(
+                // Pad and clamp the bisector values to ensure extents can be calculated
+                Math.max(0, bisector.left(data, dateExtent[0]) - 1),
+                Math.min(bisector.right(data, dateExtent[1]) + 1, data.length)
+            );
+            return visibleData;
+        };
+
+        return chart;
+    };
+})(fcsc, fc, d3);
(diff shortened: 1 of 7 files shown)


src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
+(function(fcsc, fc, d3) {
+    'use strict';
+    fcsc.chartFramework = function() {
+        var width = 500;
+        var height = 300;
+        var timeSeries = fc.chart.linearTimeSeries();
```
jleft: `linearTimeSeries` in more than just a series, so maybe rename `timeSeries` to `chart`?

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+        var multi = fc.series.multi();
+        var plugins = [];
+        var dateScale = fc.scale.dateTime()
```
jleft: Do we need this? `linearTimeSeries` has an `xScale` property; are we effectively duplicating some of its behavior?

For example, if we need the x-scale, could we just do?

```js
timeSeries.xScale();
```

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+            .range([0, width]);
+        var scaleExtent = [];
```
jleft: Is `scaleExtent` used?

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+
+        function chart(selection) {
+            selection.each(function(d, i) {
+                scaleExtent = fc.util.extent(d, ['date']);
+
+                // Y-axis automatically scales
+                timeSeries.xDomain(dateScale.domain())
+                    .yDomain(fc.util.extent(d, ['open', 'close']));
```
jleft: This assumes that the data will have open/close properties, and that the series will be created with y-values of either the open or close -  which may not always be the case.

For example, if the user specifies a series to use the data's `high` property, some of the data may be clipped:

```js
var line = fc.series.line()
    .yValue(function(d) { return d.high; });
```

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+        };
+
+        chart.dateScale = function(value) {
+            if (!arguments.length) { return dateScale; }
+            dateScale = value;
+            return chart;
+        };
+
+        // Perhaps this should be done using d3.rebind
+        chart.initDomain = function(value) {
```
jleft: Do we need a setter for the init domain, or could this be set outside of chartFramework and passed in as a property of the scale?

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+            if (!arguments.length) { return dateScale.domain(); }
+            dateScale.domain(value);
+            return chart;
+        };
+
+        chart.width = function(value) {
+            if (!arguments.length) { return width; }
+            width = value;
+            dateScale.range([0, width]);
```
jleft: Can we set this in `function chart { ...`?

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,83 @@
...
+
+        chart.plugins = function(value) {
+            if (!arguments.length) { return plugins; }
+            plugins = value;
+            return chart;
+        };
+
+        d3.rebind(chart, multi, 'series');
+
+        chart.getVisibleData = function(data, dateExtent) { // currently unused
```
jleft: If this is unused, shall we remove it and add it back in later?

src/assets/js/zoomPlugin.js
```diff
@@ -0,0 +1,17 @@
+(function(fcsc, fc, d3) {
+    'use strict';
+    fcsc.zoomPlugin = function() {
```
jleft: `d3.behavior.zoom()` is effectively a 'plugin', do we need to wrap it?

jleft:
Looks like a really good start for now, thanks @CBarrett2 :)

src/assets/js/chartFramework.js
CBarrett2: We already have a function named chart though - perhaps something else? seriesChart?

src/assets/js/chartFramework.js
CBarrett2: How do you mean? Like, as a default value, or that it reads the size of the parent element?

src/assets/js/zoomPlugin.js
CBarrett2: I've added more functionality in the callback now making me think it's neater - check the new pull request :)

PR #20, opened 2015-07-20
Feature/fixes

CBarrett2:
Fix to zoomPlugin

### DIFF
--- a/src/assets/js/chartFramework.js
+++ b/src/assets/js/chartFramework.js
@@ -8,16 +8,15 @@
         var plugins = [];
         var dateScale = fc.scale.dateTime()
             .range([0, width]);
-        var scaleExtent = [];
 
         function chart(selection) {
             selection.each(function(d, i) {
-                scaleExtent = fc.util.extent(d, ['date']);
-
                 // Y-axis automatically scales
+                var yExtent = fc.util.extent(chart.getVisibleData(d, dateScale.domain()), ['open', 'close']);
+                // Add say 20% on either side of extent
                 timeSeries.xDomain(dateScale.domain())
-                    .yDomain(fc.util.extent(d, ['open', 'close']));
-
+                    //.yDomain(fc.util.extent(d, ['open', 'close'])); non y-scaling
+                    .yDomain(yExtent);
                 timeSeries.plotArea(multi);
                 selection.call(timeSeries);
 
@@ -27,12 +26,6 @@
             });
             return chart;
         }
-
-        chart.scaleExtent = function(value) {
-            if (!arguments.length) { return scaleExtent; }
-            scaleExtent = value;
-            return chart;
-        };
 
         chart.dateScale = function(value) {
             if (!arguments.length) { return dateScale; }
--- a/src/assets/js/chartFramework.js
+++ b/src/assets/js/chartFramework.js
@@ -6,13 +6,13 @@
         var timeSeries = fc.chart.linearTimeSeries();
         var multi = fc.series.multi();
         var plugins = [];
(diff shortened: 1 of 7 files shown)


src/assets/js/main.js
```diff
@@ -1,6 +1,6 @@
 (function(d3, fc) {
```
jleft: Pass in `fcsc`. Let's be consistent with the order  for all files too - how about `d3, fc, fcsc`?

src/assets/js/main.js
```diff
@@ -1,6 +1,6 @@
...
     'use strict';
-
+    // set height/width of container
```
jleft: I don't think we need this comment, with the variable names we know what's going on.

src/assets/js/main.js
```diff
@@ -9,15 +9,11 @@
         .attr('width', width)
         .attr('height', height);
 
-    var data = fc.dataGenerator()(150);
-
-    var chart = fc.chart.linearTimeSeries()
-        .xDomain(fc.util.extent(data, 'date'))
-        .yDomain(fc.util.extent(data, ['open', 'close']));
+    var data = fc.dataGenerator()(5000);
```
jleft: `fc.dataGenerator()` has been renamed to  `fc.data.random.financial()`

src/assets/js/main.js
```diff
@@ -9,15 +9,11 @@
...
 
     var area = fc.series.area()
         .yValue(function(d) { return d.open; })
-        .y0Value(chart.yDomain()[0]);
+        .y0Value(0);
```
jleft: Any reason for `y0Value(0)` over `chart.yDomain()[0]`?

src/assets/js/main.js
```diff
@@ -26,13 +22,18 @@
         .yTicks(5)
         .xTicks(0);
 
-    var multi = fc.series.multi()
-        .series([gridlines, area, line]);
+    // Our zoom plugin implements pan and zoom
+    var zoomPlugin = fcsc.zoomPlugin();
+    //    .scaleExtent(fc.util.extent(data, ['date']));
```
jleft: Remove comments?

src/assets/js/chartFramework.js
```diff
@@ -0,0 +1,46 @@
...
+
+                timeSeries.yDomain(yExtent);
+                timeSeries.plotArea(multi);
+                selection.call(timeSeries);
+
+                for (var j = 0; j < plugins.length; j++) {
+                    selection.call(plugins[j], chart);
+                }
+            });
+            return chart;
```
jleft: Do we need to return `chart`?

src/assets/js/zoomPlugin.js
```diff
@@ -0,0 +1,28 @@
...
+                        var ty = zoomBehavior.translate()[1];
+                        zoomBehavior.translate([tx, ty]);
+                    }
+                    lastX = d3.mouse(this)[0];
+                    // Recalling chart draws the new view, this is a place-holder for drawing/two way binding
+                    chart(selection);
+                });
+
+            selection.call(zoomBehavior);
+            return zoom;
```
jleft: Do we need to return `zoom` here?

src/assets/js/chartFramework.js
CBarrett2: I suppose it's just in case you want to chain operations, it's from the re-useable chart framework for d3. I can remove it if you like though?

jleft: We would still be able to chain operations if `return  chart` is removed - we only need to do this in the getters/setters.

PR #22, opened 2015-07-22
Feature/uifixes

CBarrett2:


src/assets/js/main.js
```diff
@@ -1,38 +1,202 @@
...
+    var rsiHeight = height / 2;
+    var navHeight = height / 3;
 
-    var svg = d3.select('#chart-example')
-        .append('svg')
+
+    // Set SVGs
+    var container = d3.select('#chart-example');
+    var svgMain = container.select('svg.main')
+        .attr('viewBox', function() { return '0 0 ' + width + ' ' + height; })
```
jleft: Do we need `viewBox`?

src/assets/js/main.js
```diff
@@ -1,38 +1,202 @@
...
+        .xTicks(0);
+
+    // Create and apply the Moving Average
+    var movingAverage = fc.indicator.algorithm.movingAverage();
+
+    // Create a line that renders the result
+    var ma = fc.series.line()
+        .yValue(function(d) { return d.movingAverage; });
+
+    var multi = fc.series.multi().series([candlestick, gridlines, ma]);
```
jleft: Can we pass in gridlines first, so they're rendered under the candlesticks and ma?

src/assets/js/main.js
```diff
@@ -1,38 +1,202 @@
...
+
+    function zoomCall(zoom, data, scale) {
```
jleft: I'm still finding some slightly peculiar behaviour with the zoom - we can raise bugs against this.

src/assets/js/main.js
```diff
@@ -1,38 +1,202 @@
...
+    var rsiScale = d3.scale.linear()
+        .domain([0, 100])
+        .range([rsiHeight, 0]);
+    var rsiAlgorithm = fc.indicator.algorithm.relativeStrengthIndex();
+
+    var rsi = fc.indicator.renderer.relativeStrengthIndex()
+        .xScale(timeSeries.xScale())
+        .yScale(rsiScale);
+
+    var rsiChart = function(selection) {
```
jleft: I think it's worth considering if this should have its own x-axis, happy to leave as-is for now though.

src/assets/js/main.js
```diff
@@ -1,38 +1,202 @@
...
+                }
+            });
 
-    svg.datum(data)
-        .call(chart);
+        // Allow to zoom using mouse, but disable panning
+        var zoom = d3.behavior.zoom()
+            .x(timeSeries.xScale())
+            .on('zoom', function() {
+                if (zoom.scale() === 1) {
```
jleft: Should we be making use of `zoomCall()` here, to limit the zoom?

CBarrett2:
Removed the extra features for now, still working on a fix but thought at least without them it might be able to get pulled

jleft:
> Removed the extra features for now, still working on a fix but thought at least without them it might be able to get pulled

Great, thank you :)

Please can we add Bootstrap as a dependency, then this should be good to go:
```sh
npm install bootstrap --save
```

PR #31, opened 2015-07-28
Feature/fix panning

CBarrett2:


src/assets/js/main.js
```diff
@@ -11,6 +11,18 @@
...
     }
 
+    function getExtent(data) {
+        /* Applies the fc.util.extent function to find the extent of dates,
+        but adds a buffer of one day to either side */
+        var buffer = 1000 * 60 * 60 * 24;
+
+        var xExtent = fc.util.extent(data, ['date']);
+        xExtent[0] = new Date(xExtent[0].getTime() - buffer);
+        xExtent[1] = new Date(xExtent[1].getTime() + buffer);
```
jleft: [D3's time API](https://github.com/mbostock/d3/wiki/Time) is useful for this; I think this looks a bit easier to read and should also work if there is a DST transition! What do you think?

```js
var xExtent = fc.util.extent(data, ['date']);
xExtent[0] = d3.time.day.offset(xExtent[0], -1);
xExtent[1] = d3.time.day.offset(xExtent[1], 1);
```

src/assets/js/main.js
```diff
@@ -124,8 +136,9 @@
 
     // Create navigation chart
     var yExtent = fc.util.extent(getVisibleData(data, fc.util.extent(data, 'date')), ['low', 'high']);
+    var xExtent = getExtent(data);
     var navTimeSeries = fc.chart.linearTimeSeries()
-        .xDomain(fc.util.extent(data, 'date'))
+        .xDomain(xExtent)
```
jleft: Should we also update the xExtent for any of the other charts and in zoomCall?

CBarrett2: Everything else is based on the currently visible extent of the data, whereas this is about the absolute extremes of all the data we have :)

jleft: Ah, yeah, it is used in zoomCall, I must have misread the diff!

I suppose one downside to this approach is that there'll be a small gap between the end of the area series in the navigator and the y-axis? I don't think this is a big deal for now though!

src/assets/js/main.js
```diff
@@ -11,6 +11,18 @@
         return visibleData;
     }
 
+    function getExtent(data) {
```
jleft: I think the function name is a bit misleading as it isn't just getting the extent of the data. Maybe something like `padExtentAdditionalDay` or `offsetExtentAdditionalDay`?

jleft:
Are these changes now incorporated in #47?

CBarrett2:
yes :)

PR #32, opened 2015-07-28
Set end bar to end of static data

djmiley:


src/assets/js/main.js
```diff
@@ -29,7 +29,7 @@
 
     // Create main chart and set how much data is initially viewed
     var timeSeries = fc.chart.linearTimeSeries()
-        .xDomain([data[Math.floor(data.length / 2)].date, data[Math.floor(data.length * 3 / 4)].date])
+        .xDomain([data[Math.floor((5 - 1.618) * data.length / 5)].date, data[Math.floor(data.length) - 1].date])
```
jleft: Why `1.618`?

`Math.floor(data.length)` - I don't think we need to do this, as `data.length` should always be an integer.

djmiley:
1.618 is approximately the golden ratio, which is allegedly meant to produce a pleasingly sized rectangle. The choice is relatively arbitrary for now.

jleft:
> 1.618 is approximately the golden ratio, which is allegedly meant to produce a pleasingly sized rectangle. The choice is relatively arbitrary for now.

Nice idea. I think it would be worth assigning the value to a `var` with an appropriate name, so it doesn't seem so arbitrary (at least to me!), or possibly add a comment.

PR #33, opened 2015-07-28
Added zero line to data, noughitified data

djmiley:
Remove or leave noughtify function?


djmiley:
Altering this commit, due to lack of obvious need.

PR #34, opened 2015-07-28
Resize chart in accordance with window size

djmiley:
Introduced padding style & use of scroll size

Width resizes well

Done except for small margin error

Resizing works wihtout padding

Removed console log

Resizes window

Reintroduced padding


djmiley:
Couldn't figure out how to get the Bootstrap padding externally so set it to same value. I must have missed something.

Removed SVG pixel dependence but still tied to ratio dependence.

djmiley:
Doing some work on abstracting layout out of js file, and will recommit

PR #38, opened 2015-07-29
Resize chart in accordance with window size

djmiley:
Introduced padding style & use of scroll size

Width resizes well

Done except for small margin error

Resizing works wihtout padding

Removed console log

Resizes window

Reintroduced padding

Removed some constant dependence

Removed padding dependency in js file

### DIFF
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -2,11 +2,11 @@ path.line {
   stroke: rgb(26, 154, 249);
 }
 
-#chart-example {
-  position: absolute;
-  left: 10px;
-  top: 10px;
-} 
+.main, .rsi, .nav {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+}
 
 .extent {
   stroke: #D1D1E0;
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -11,19 +11,20 @@
         return visibleData;
     }
 
-    var width = 500;
-    var height = 300;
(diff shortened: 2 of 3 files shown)


src/assets/css/style.css
```diff
@@ -2,11 +2,23 @@ path.line {
   stroke: rgb(26, 154, 249);
 }
 
-#chart-example {
-  position: absolute;
-  left: 10px;
-  top: 10px;
-} 
+.main {
```
jleft: We're repeating the same properties/values here, can we either create a single class (maybe `chart`?) and apply this to all or the chart `div`s? (We can set multiple classes on the same element.) Or combine the classes like this:

```css
.main, .rsi, .nav {
  display: block;
  margin-left: auto;
  margin-right: auto;
}
```

src/assets/js/main.js
```diff
@@ -11,20 +11,21 @@
...
-    var rsiHeight = height / 2;
-    var rsiAspect = rsiHeight / width;
-    var navHeight = height / 3;
-    var navAspect = navHeight / width;
-
-    // Set SVGs
+    // Set SVGs & column padding
     var container = d3.select('#chart-example');
+
+    var leftPadding = parseInt(container.selectAll('.col-md-12').style('padding-left'), 10);
```
jleft: Can we just `select`, rather than `selectAll`?

jleft:
Thanks @djmiley - as discussed we'll revisit this soon when working on the layout.

PR #41, opened 2015-07-30
Feature/merge data stream: WIP

CBarrett2:
Just for people to have a look at :) Not completely finished yet - there's a few tweaks that definitely need doing before it's pulled, most tasks are listed at the top of the dataInterface class


CBarrett2:
If you pan to the far left it makes a call for new data, and live data streams in all the time, altering the last candlestick. You can change periods in the main.js file where dataInterface is initialized

mobile/.gitignore
```diff
@@ -1,4 +0,0 @@
-hooks
```
jleft: We'll still need the mobile/* files

src/assets/js/dataInterface.js
```diff
@@ -0,0 +1,147 @@
+(function(fc) {
+    'use strict';
+    fc.data.dataInterface = function() {
```
jleft: Let's create our own namespace so we don't assign to `fc`.

src/assets/js/dataInterface.js
```diff
@@ -0,0 +1,147 @@
...
+                    console.log(currentData)
+                    callback(currentData);
+                }
+                else { console.log("Error getting data from live feed: " + err); }
+            })
+
+            return dataInterface;
+        };
+
+        dataInterface.createNewBasket = function(datum) {
```
jleft: Does this need to be public?

src/assets/js/dataInterface.js
```diff
@@ -0,0 +1,147 @@
...
+                date: datum.date, // round to nearest period, middle
+                open: datum.price,
+                close: datum.price,
+                low: datum.price,
+                high: datum.price,
+                volume: datum.volume
+            };
+        }
+
+        dataInterface.updateBasket = function(datum) {
```
jleft: Does this need to be public?

src/assets/js/main.js
```diff
@@ -11,6 +11,20 @@
         return visibleData;
     }
 
+    function getPaddedExtent(data) { // Change name
+        /* Applies the fc.util.extent function to find the extent of dates,
+        but adds a buffer of one day to either side */
+        var period = dataInterface.period();
+        var buffer = 1000 * period
```
jleft: Unused?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,37 @@
+(function(fc) {
+    'use strict';
+
+    fc.data.websocket = function() {
+        var timeChunk = 2500; // Every 2.5 seconds
```
jleft: Unused?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,37 @@
+(function(fc) {
+    'use strict';
+
+    fc.data.websocket = function() {
```
jleft: If this is tied to coinbase, then I think we should update the component name `sc.data.feed.coinbase.live` or something like that?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,37 @@
...
+        var timeChunk = 2500; // Every 2.5 seconds
+
+        function websocket(cb) {
+            var coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
+            var msg = {
+                type: 'subscribe',
+                product_id: 'BTC-USD'
```
jleft: Worth having a getter/setter for this?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,37 @@
+(function(fc) {
+    'use strict';
+
+    fc.data.websocket = function() {
+        var timeChunk = 2500; // Every 2.5 seconds
+
+        function websocket(cb) {
+            var coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
```
jleft: We should also handle errors with [`coinbaseSocket.onerror`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#Attributes)

Having the ability to [close](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#close()) the connection makes sense too.

src/assets/js/dataInterface.js
jleft: This component seems to have two distinct responsibilities:

  1. Handling historic data
  2. Handling live data

It could be good to separate these concerns into separate components.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,37 @@
...
+            };
+
+            coinbaseSocket.onmessage = function(event) {
+                var jMsg = JSON.parse(event.data);
+                if (jMsg.type === 'received' && jMsg.size > 1) {
+                    var datum = {};
+                    datum.date = new Date(jMsg.time);
+                    datum.price = parseFloat(jMsg.price);
+                    datum.volume = parseFloat(jMsg.size);
+                    cb(null, datum);
```
jleft: `null`?

src/assets/js/dataInterface.js
CBarrett2: Yeah I see your point about that. So it then be the chart's responsibility to join these two data sources?

jleft: I think the chart's responsibility should be to appropriately render the chart, given data (and some configuration properties). I think we could have another module, responsible for joining the data sources. Does that sound reasonable?

PR #42, opened 2015-07-30
Added naive reset button

djmiley:


src/assets/js/main.js
```diff
@@ -28,6 +28,16 @@
 
     var data = fc.data.random.financial()(250);
 
+    // Set Reset button event
+    function resetToLive() {
+        timeSeries.xDomain([data[Math.floor((1 - navAspect * goldenRatio) * data.length)].date,
```
jleft: It's probably worth pulling out this value and assigning it to a `var` for reuse, as it's also [used here](https://github.com/djmiley/d3fc-showcase/blob/feature/resetbutton/src/assets/js/main.js#L46). It should be easier to maintain, should we choose to change the default position of the navigator in the future.

src/assets/js/main.js
```diff
@@ -28,6 +28,16 @@
...
+            data[data.length - 1].date]);
+        render();
+    }
+
+    var resetButton = container.select('#reset-button')[0][0];
+    resetButton.addEventListener('click', resetToLive);
```
jleft: The D3 way to add an event listener is using [selection.on()](https://github.com/mbostock/d3/wiki/Selections#on)
```js
container.select('#reset-button')
    .on('click', resetToLive);
```

src/assets/js/main.js
```diff
@@ -214,8 +224,10 @@
     }
 
     function resize() {
+        var resetRowHeight = parseInt(container.select('#reset-row').style('height'), 10);
+
         var useableScreenWidth = window.innerWidth - (leftPadding + rightPadding);
-        var useableScreenHeight = window.innerHeight;
+        var useableScreenHeight = window.innerHeight - resetRowHeight;
```
jleft: I think we should get the height of the chart's container, rather than the window, then we wouldn't have to worry about this calculation. This is ok for now though and we can reconsider this approach when working on the new layout.

Issue #39, opened 2015-07-29
Remove Linear Time Series


simonterrywatson:
Use lower level components to compose the charts, which will allow for more flexibility as we add more functionality later on.

djmiley:
#96

PaulScho:
Not required

PR #48, opened 2015-07-31
Removed linear time series and componentised chart

djmiley:


djmiley:
There is a lot of refactoring that can be done here, but I hope it is on the right track.

djmiley:
Have fixed panning bug. Will look into some more refactoring.

jleft:
#39

djmiley:
There is a bug in here that needs fixing, rsi not rendering correctly.

PR #55, opened 2015-08-06
Added websocket and OHLC converter

CBarrett2:


src/assets/js/sc.js
```diff
@@ -0,0 +1,17 @@
+(function() {
+    'use strict';
+
+    // Crazyness to get a strict mode compliant reference to the global object
+    var global = null;
+    /* jshint ignore:start */
+    global = (1, eval)('this');
+    /* jshint ignore:end */
+
+    global.sc = {
```
jleft: Need to define `sc` as a global in .jshintrc

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
+(function(sc) {
+    'use strict';
+    sc.data.feed.coinbase.websocket = function() {
+        // can get product list from /products
```
jleft: Remove comment.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+        var product = 'BTC-USD';
+        var productList = [];
```
jleft: Do we need to cache this? I don't think this cache can/will ever be invalidated?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+        var msgType = 'match';
+        var coinbaseSocket = null;
+        var callback = null; // not sure we want to be storing things like this?
```
jleft: I don't think we should - unless there's a reason to?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+
+        function websocket(cb) {
+            var coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
```
jleft: Should `coinbaseSocket ` be re-declared here, or should we be using the var declared on line 8?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+
+            callback = cb;
+
+            coinbaseSocket.onopen = function() {
+                coinbaseSocket.send(JSON.stringify(msg));
+            };
+
+            coinbaseSocket.onmessage = function(event) {
+                var jMsg = JSON.parse(event.data);
+                if (jMsg.type === msgType) {
```
jleft: Will we only ever want just one message type? I think this is probably okay for now.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            };
+
+            callback = cb;
+
+            coinbaseSocket.onopen = function() {
+                coinbaseSocket.send(JSON.stringify(msg));
+            };
+
+            coinbaseSocket.onmessage = function(event) {
+                var jMsg = JSON.parse(event.data);
```
jleft: Maybe rename to `messageData`?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+                if (jMsg.type === msgType) {
+                    var datum = {};
+                    datum.date = new Date(jMsg.time);
```
jleft: Do we need to check if the properties exist on `jMsg` or will we always have the appropriate properties?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+                    datum.price = parseFloat(jMsg.price);
+                    datum.volume = parseFloat(jMsg.size);
+                    console.log(datum);
```
jleft: Remove.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+                    // Pass errors back also
+                    cb(datum);
+                }
+            };
+
+            coinbaseSocket.onerror = function(err) {
+                console.log('Error loading data from coinbase websocket: ' + err);
```
jleft: Let's pass the error to the callback, so it can be handled in the calling code.

The callback could then expect and error and data:

```js
cb(error, data);
```

If there is no error (e.g. in `onmessage`), we can just pass in `null`:

```js
cb(null, data);
```

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            };
+        }
```
jleft: I think it's worth handling the `coinbaseSocket.onclose` event too.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+
+        websocket.close = function() {
+            if (coinbaseSocket) {
+                coinbaseSocket.close();
```
jleft: What if the connection is not open yet, is closing, or has already been closed? Should we only close if it's already open?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            }
+        };
+
+        websocket.msgType = function(msgType) {
+            if (!arguments.length) { return msgType; }
```
jleft: Stylistically, I think it would be good to be consistent across getters/setters like this:

```js
if (!arguments.length) {
    return msgType;
}
```

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            msgType = msgType;
```
jleft: In don't think it will affect the property on the class, as `msgType` will refer to the parameter passed into the function. Maybe pass in a var named `value` to all getters/setters?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            return websocket;
+        };
+
+        websocket.product = function(product) {
+            if (!arguments.length) { return product; }
+            product = product;
+            if (coinbaseSocket) {
+                coinbaseSocket.close();
+                websocket(callback);
+            }
```
jleft: Should we close the websocket and reopen a different one when the product changes, or should we leave this responsibility to the calling code?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+            return websocket;
+        };
+
+        websocket.getProductList = function(cb) {
+            if (productList) { cb(productList); }
+            d3.json('https://api.exchange.coinbase.com/products/', function(err, data) {
+                // jscs:disable
```
jleft: Can we disable the specific rule, or use bracket notation?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+                productList = data.map(function(currValue) { return currValue.display_name; });
+                // jscs:enable
+                cb(productList);
+            });
+        };
+
+        return websocket;  
+    };
+    
+    sc.data.OHLC = function() {
```
jleft: Put into a separate file.

jleft: Probably just go all lowercase too:

```js
sc.data.ohlc
```

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+        // Expects transactions with a price, volume and date and organizes them into candles of given periods
+        var currentBasket = null;
+        var period = 60 * 60 * 24;
```
jleft: Which unit of time is this in - seconds?

src/index.html
```diff
@@ -15,7 +15,7 @@
     <script src="assets/js/Layout.js"></script>
     <script src="assets/js/d3fc.js"></script>
 
-    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
+    <script src="assets/js/jquery.min.js"></script>
```
jleft: Does this exist?

src/index.html
```diff
@@ -41,7 +53,12 @@
           </div>
         </div>
       </div>
+      
+      
     </div>
+    <script src="assets/js/sc.js"></script>
+    <script src="assets/js/websocket.js"></script>
+    <script src="assets/js/dataInterface.js"></script>
```
jleft: Does this exist?

src/index.html
```diff
@@ -23,6 +23,18 @@
       <div class="row" id="reset-row">
         <div class="col-md-12">
           <button type="button" id="reset-button">Reset</button>
+          Period
```
jleft: Do we these options in this pull request?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+        var liveFeed = null;
```
jleft: Any reason not to set the default to: `sc.data.feed.coinbase.websocket()`?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,144 @@
...
+        }
+        
+        OHLC.period = function(x) {
+            if (!arguments.length) { return period; }
+            period = x;
+            currentBasket = null;
+            return OHLC;
+        };
+
+        OHLC.basket = function(x) {
```
jleft: Don't need the parameter.

src/assets/js/websocket.js
CBarrett2: It's so we can just re-initialize the websocket when product is changed - would it work to just send a second subscription message from the same websocket? right now I close the original/open a new one

src/assets/js/websocket.js
CBarrett2: What would we do in this case? give an error to the callback also?

src/assets/js/websocket.js
CBarrett2: This is consistent with d3fc, I can change it though :)

jleft: I think all on the same line isn't consistent with the getter/setters (e.g. [area series](https://github.com/ScottLogic/d3fc/blob/master/src/series/area.js#L49-L83)), but it is used when declaring `vars`, chaining methods etc. Arrow functions would be nice! :)

src/assets/js/websocket.js
jleft: How about being more generic and passing the event through:

```js
cb(event, data);
```

src/assets/js/websocket.js
jleft: I suppose this depends on https://github.com/ScottLogic/d3fc-showcase/pull/55/files#r36411667. I'm not sure about this one - @tlsim, any thoughts?

src/assets/js/websocket.js
CBarrett2: I think it's pretty unintuitive to think you have to re-open the websocket when you change the product, but not when you change the period. I'll have a look at the websocket api and see if there's a way of changing subscription without re-starting the whole thing

jleft: > I think it's pretty unintuitive to think you have to re-open the websocket when you change the product, but not when you change the period. 

There's no way to change the period when using the websocket though?

src/assets/js/websocket.js
tlsim: Sending another subscribe message won't work unfortunately - https://docs.exchange.coinbase.com/#subscribe "Any other messages, or a mal-formed message, will result in your connection being terminated."
I'm not a fan of having reconnect logic in a getter/setter. Simpler to just have the caller re-initialize sc.data.feed.coinbase.weboskcet with a new callback.

jleft: > I'm not a fan of having reconnect logic in a getter/setter. Simpler to just have the caller re-initialize sc.data.feed.coinbase.weboskcet with a new callback.

That was my initial feeling - I think it's more flexible this way too. I wonder if there is ever a use case to change the product and not re-initialise though?

CBarrett2: Ah yeah, you're right! Okay, in that case I think it makes sense to have ohlc handle restarting the websocket

tlsim: > I wonder if there is ever a use case to change the product and not re-initialise though

Yeah probably not. Could initialize with product string and remove the product getter/setter?

Gruntfile.js
```diff
@@ -204,12 +210,31 @@ module.exports = function(grunt) {
                     platforms: ['android']
                 }
             }
+        },
+
+        jasmine: {
```
jleft: If possible, can we leave the jasmine config out for now? `git add -p` should be useful

Gruntfile.js
```diff
@@ -204,12 +210,31 @@ module.exports = function(grunt) {
...
+                }
+            }
         }
 
     });
 
     require('load-grunt-tasks')(grunt);
 
+    grunt.registerTask('test', ['jasmine:test']);
+    grunt.registerTask('serve', ['build', 'connect:dist']);
```
jleft: I think this task already exists at the bottom of this file?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
+(function(sc) {
+    'use strict';
+    sc.data.feed.coinbase.websocket = function() {
+        var product = 'BTC-USD';
+        var msgType = 'match';
+        var coinbaseSocket = null;
+        var callback = null; // not sure we want to be storing things like this?
```
jleft: Remove.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
...
+
+        function websocket(cb) {
+            coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
+            var msg = {
+                type: 'subscribe',
+                'product_id': product
+            };
+
+            callback = cb;
```
jleft: Remove.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
...
+            };
+
+            coinbaseSocket.onclose = function(event) {
+                cb(event, null);
+            };
+
+        }
+
+        websocket.close = function() {
+            if (coinbaseSocket) {
```
jleft:
```js
if (coinbaseSocket && coinbaseSocket.readyState === 1)
```

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
...
+            product = x;
+            return websocket;
+        };
+
+        websocket.getProductList = function(cb) {
+            d3.json('https://api.exchange.coinbase.com/products/', function(err, data) {
+                // jscs:disable
+                var productList = data.map(function(currValue) { return currValue.display_name; });
+                // jscs:enable
+                cb(productList);
```
jleft: Should we pass `err` to the callback?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
...
+                return product;
+            }
+            product = x;
+            return websocket;
+        };
+
+        websocket.getProductList = function(cb) {
+            d3.json('https://api.exchange.coinbase.com/products/', function(err, data) {
+                // jscs:disable
+                var productList = data.map(function(currValue) { return currValue.display_name; });
```
jleft: Is `currValue` a product? If so, rename to `product`?

src/assets/js/websocket.js
```diff
@@ -0,0 +1,82 @@
...
+            if (!arguments.length) {
+                return product;
+            }
+            product = x;
+            return websocket;
+        };
+
+        websocket.getProductList = function(cb) {
+            d3.json('https://api.exchange.coinbase.com/products/', function(err, data) {
+                // jscs:disable
```
jleft: Can we disable the specific rule, or use bracket syntax?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,85 @@
...
+
+        ohlc.product = function(x) {
+            if (!arguments.length) {
+                return liveFeed.product();
+            }
+            liveFeed.product(x);
+            liveFeed.close();
+            ohlc();
+            currentBasket = null;
+        };
```
jleft: `return ohlc;`

What happens if we want to set the product before opening the connection/calling `ohlc`?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,85 @@
...
+            });
+        }
+
+        ohlc.liveFeed = function(x) {
+            if (!arguments.length) {
+                return liveFeed;
+            }
+            liveFeed = x;
+            currentBasket = null;
+            // This requires liveFeed to be set up first, before changing prodcut/etc on liveFeed
```
jleft: Does this belong here?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,85 @@
...
+            period = x;
+            currentBasket = null;
+            return ohlc;
+        };
+
+        ohlc.basket = function() {
+            return currentBasket;
+        };
+
+        function updateBasket(datum) {
```
jleft: I think messages can arrive [out-of-order](https://docs.exchange.coinbase.com/#sequence-numbers), should we handle this here? e.g. we might have to update the open if an earlier message arrives after a later one?

CBarrett2: That would be something we could do, however if a message arrives earlier than the start of the current basket, there's currently no way of going back and editing it (would involve storing state of previous baskets)

CBarrett2: How about the case of 'close'? Should we be storing the latest time in order to check that?

src/assets/js/ohlc.js
CBarrett2: Ah, good point. I'm not really sure what to do about this - it doesnt cause much harm atm as there's a dummy callback. perhaps we could set ohlc() to restart the websocket also? then on the live() call it would re-initialize as wanted. hesitant to pass the responsibility on to caller for the reasons I mentioned before

src/assets/js/websocket.js
CBarrett2: Do you want to disable the specific rule throughout the whole project? Using product[display_name] gives a style error also.

jleft: > Do you want to disable the specific rule throughout the whole project?

No, I was thinking we could just disable the specific rule (not sure which rule it is!) here:

```js
// jscs:disable specificRule
var productList = data.map(function(product) { return product.display_name; });
// jscs:enable specificRule

Comment (CBarrett2) [inline src/assets/js/websocket.js]: Ah, okay! I can do that :)

File: src/assets/js/websocket.js [hunk 35]
```diff
@@ -0,0 +1,77 @@
...
+
+        function websocket(cb) {
+            coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
+            var msg = {
+                type: 'subscribe',
+                'product_id': product
+            };
+
+            coinbaseSocket.onopen = function() {
+                coinbaseSocket.send(JSON.stringify(msg));
```
jleft: If think it would be good to pass the `onopen` event to the callback too.

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,68 @@
+(function(sc) {
+    'use strict';
+    sc.data.ohlc = function() {
```
jleft: How about `sc.data.feed.coinbase.ohlcWebSocketAdapter`?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,68 @@
...
+        // Expects transactions with a price, volume and date and organizes them into candles of given periods
+        // Re-call OHLC whenever you want to start collecting for a new period/product
+        // In seconds
+        var period = 60 * 60 * 24;
+        var liveFeed = sc.data.feed.coinbase.websocket();
+
+        function ohlc(cb) {
+            liveFeed.close();
```
jleft: Should closing be the responsibility of the WebSocket component?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,68 @@
...
+            var currentBasket = null;
```
jleft: Maybe just `basket`?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,68 @@
...
+
+        ohlc.period = function(x) {
+            if (!arguments.length) {
+                return period;
+            }
+            period = x;
+            return ohlc;
+        };
+
+        d3.rebind(ohlc, liveFeed, 'product');
```
jleft: Any reason not to rebind the other getters/setter? Message type, product list?

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,68 @@
...
+            var startTime = basket.date.getTime();
+            var msPeriod = period * 1000;
+            if (latestTime > startTime + msPeriod) {
+                basket = createNewBasket(datum);
+            } else {
+                // Update current basket
+                basket.high = Math.max(basket.high, datum.price);
+                basket.low = Math.min(basket.low, datum.price);
+                basket.volume += datum.volume;
+                // Messages can arrive out of order, so in this case we might want to update the open
```
jleft: As discussed, let's not worry about this for now.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,77 @@
+(function(sc) {
+    'use strict';
+    sc.data.feed.coinbase.websocket = function() {
```
jleft: `websocket` -> `webSocket`

src/assets/js/websocket.js
jleft: >```js
>if (coinbaseSocket && coinbaseSocket.readyState === 1)
>```

This still applies.

src/assets/js/websocket.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        websocket.close = function() {
+            if (coinbaseSocket) {
+                coinbaseSocket.close();
+            }
+            return websocket;
+        };
+
+        websocket.msgType = function(x) {
```
jleft: `msgType` -> `messageType`

src/assets/js/websocket.js
```diff
@@ -0,0 +1,77 @@
...
+
+        websocket.product = function(x) {
+            if (!arguments.length) {
+                return product;
+            }
+            product = x;
+            return websocket;
+        };
+
+        websocket.getProductList = function(cb) {
```
jleft: `getProductList` -> `productList`

src/assets/js/websocket.js
```diff
@@ -0,0 +1,77 @@
...
+            d3.json('https://api.exchange.coinbase.com/products/', function(err, data) {
+                // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+                var productList = data.map(function(product) { return product.display_name; });
```
jleft: Actually, should we return only the display name, or all of the properties?

src/assets/js/websocket.js
jleft: Should we close if the connection is opening too?

CBarrett2: I looked up the API and close() does nothing if the websocket's already closed, so didn't think we needed to check if it was open first. I think it should close no matter what, as that's presumably what the user's asking for when they call close()

src/assets/js/websocket.js
CBarrett2: The other properties on the object are: 
base_currency: "BTC"
base_max_size: "10000"
base_min_size: "0.01"
display_name: "BTC/USD"
id: "BTC-USD"
quote_currency: "USD"
quote_increment: "0.01"
not really sure any of these are useful except display_name, for setting the messageType. But could easily pass the whole objects back? I feel maybe just passing the display_name property back makes for a more generic interface

CBarrett2: In actual fact, I just realized we should be passing the id back and not the display_name!

jleft: I think if we pass the whole object back, then we can let the calling code use it as desired - I can imagine quite a few of the fields being useful, especially `base_currency`, `display_name`, `id` and `quote_currency`.

CBarrett2:
New update :)

Gruntfile.js
```diff
@@ -80,6 +80,12 @@ module.exports = function(grunt) {
                     expand: true
                 },
                 {
+                    cwd: 'node_modules/jquery/dist',
```
jleft: I don't think we have jQuery as a dependency? It will need to be.

src/assets/js/websocket.js
jleft: `sc.data.feed.coinbase.websocket` -> sc.data.feed.coinbase.webSocket`

Update file name too.

src/assets/js/websocket.js
jleft: > I looked up the API and close() does nothing if the websocket's already closed, so didn't think we needed to check if it was open first. I think it should close no matter what, as that's presumably what the user's asking for when they call close()

It feels nicer to me only closing if the WebSocket isn't already closing or closed, but I can't see any harm in doing it this way :).

src/assets/js/websocket.js
```diff
@@ -0,0 +1,76 @@
...
+            };
+
+            coinbaseSocket.onmessage = function(event) {
+                var messageData = JSON.parse(event.data);
+                if (messageData.type === msgType) {
+                    var datum = {};
+                    datum.date = new Date(messageData.time);
+                    datum.price = parseFloat(messageData.price);
+                    datum.volume = parseFloat(messageData.size);
+                    console.log(datum);
```
tlsim: Can remove this logging.

src/assets/js/ohlc.js
```diff
@@ -0,0 +1,62 @@
...
+
+        function updateBasket(basket, datum) {
+            if (basket == null) {
+                basket = createNewBasket(datum);
+            }
+            var latestTime = datum.date.getTime();
+            var startTime = basket.date.getTime();
+            var msPeriod = period * 1000;
+            if (latestTime > startTime + msPeriod) {
+                basket = createNewBasket(datum);
```
tlsim: It might be that using the trade time as basket time will produce oddly spaced candles if trades aren't happening too regularly (as is the case with coinbase).

tlsim:
Looking much better now :) I still think moving the 'get product list' functionality out of websocket would be worthwhile. Would also be nice to see the ohlc adaptor used in the chart (if it isn't already?)

src/assets/js/websocket.js
jleft: > `sc.data.feed.coinbase.websocket -> sc.data.feed.coinbase.webSocket`
>
> Update file name too.

Still needs to be done.

jleft:
> I still think moving the 'get product list' functionality out of websocket would be worthwhile

Agreed - I'll create an issue for this.

> Would also be nice to see the ohlc adaptor used in the chart (if it isn't already?)

Definitely - I'll add an issue for this too :)

src/assets/js/main.js
```diff
@@ -11,6 +11,17 @@
         return visibleData;
     }
 
+    function padExtent(extent) {
+        /* Applies the fc.util.extent function to find the extent of dates,
```
tlsim: Misleading comment - there is no call to fc.util.extent here

src/assets/js/main.js
```diff
@@ -18,6 +29,7 @@
     var svgRSI = container.select('svg.rsi');
     var svgNav = container.select('svg.nav');
 
+    var navAspect;
     function calculateDimensions() {
         var leftPadding = parseInt(container.select('.col-md-12').style('padding-left'), 10);
```
tlsim: Could be clearer about what this is meant to be selecting - maybe add classes for the container elements.

CBarrett2: This isn't my code, I don't really know what it's doing

src/assets/js/main.js
```diff
@@ -18,6 +29,7 @@
     var svgRSI = container.select('svg.rsi');
     var svgNav = container.select('svg.nav');
 
+    var navAspect;
     function calculateDimensions() {
```
tlsim: Not just *calculate*, but *set* too. There's a lot going on here and it would benefit from being split into smaller pieces. This would also help remove the code duplication around setting attributes.

src/assets/js/webSocket.js
```diff
@@ -0,0 +1,68 @@
+(function(sc) {
+    'use strict';
+    sc.data.feed.coinbase.webSocket = function() {
+        var product = 'BTC-USD';
+        var msgType = 'received';
```
tlsim: Need to request 'match' messages somewhere to get actual trades. Are we sure 'received' messages should be the default?

src/assets/js/main.js
```diff
@@ -56,32 +68,95 @@
...
+                resetToLive();
+                loading(false, '');
+            } else if (currData[currData.length - 1].date.getTime() !== latestBasket.date.getTime()) {
+                currData.push(latestBasket);
+            }
+            render();
+        } else if (event.type === 'open') {
+            loading(true, 'Connected, waiting for data...');
+        } else if (event.type === 'close') {
+            // I don't think there's any need for a message on successful close
```
tlsim: What about unsuccessful close? Useful error info is present in the event received in onclose. https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
onerror on the other hand only receives 'a simple event named "error"'.

src/assets/js/main.js
```diff
@@ -251,14 +331,43 @@
         selection.call(navTimeSeries);
     };
 
+    function loading(bool, text) {
+        if (bool) {
```
tlsim: If what? `bool` isn't a helpful name.

src/assets/js/webSocket.js
CBarrett2: Usually we should use match, I agree, but for demo purposes there are too few matches IMO to get interesting data

src/assets/js/main.js
CBarrett2: I can have a look at fixing this, but think it might be best saved for a separate PR, as this isn't code I'm adding just now.

src/assets/js/main.js
```diff
@@ -251,14 +331,43 @@
         selection.call(navTimeSeries);
     };
 
+    function loading(bool, text) {
```
tlsim: This function is setting loading/status text, but also changing the visibility of chart pieces. Consider doing these in separate functions with more 'intention revealing' names.

src/assets/js/main.js
djmiley: Should be making navAspect redundant by recalculating it in resetToLive()

src/assets/js/main.js
```diff
@@ -251,14 +331,43 @@
...
+            svgMain.style('visibility', 'visible');
+            svgRSI.style('visibility', 'visible');
+            svgNav.style('visibility', 'visible');
+            d3.select('#loading-text')
+                .style('visibility', 'hidden');
+        }
+
+        var type = d3.select('#type-selection').property('value');
+        if (type === 'live') {
+            svgRSI.style('visibility', 'hidden');
```
tlsim: Is hiding these for live data a requirement?

CBarrett2: At the minute there's only one or very few data points, so it looks a lot better without.

src/assets/js/webSocket.js
tlsim: There's more data yes, but the chart will not tell you much about the price of bitcoin. Let's add a fake live feed as a new task for demo purposes.

jleft: > Let's add a fake live feed as a new task for demo purposes. #66

:+1:

CBarrett2:
See #77

PR #64, opened 2015-08-10
Feature/historic

CBarrett2:
Superseded pull request for websocket. Adds and integrates historic data with live data.


CBarrett2:
See #77 first

PR #68, opened 2015-08-11
Closing price tag implemented

djmiley:
Get rid of open/close labels/ lines now?

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -88,15 +88,41 @@
         .yTicks(5)
         .xTicks(0);
 
-    var startPriceLine = fc.annotation.line()
-        .orient('horizontal')
-        .value(function(d) { return d.open; })
-        .label(function(d) { return 'OPEN'; });
+    function calculateCloseAxisTagPath(width, height) {
+        var h2 = height / 2;
+        return [
+            [0, 0],
+            [h2, -h2],
+            [width, -h2],
+            [width, h2],
+            [h2, h2],
+            [0, 0]
+        ];
+    }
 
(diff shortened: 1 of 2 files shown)


jleft:
Look really good, nice work :-)

> Get rid of open/close labels/ lines now?

Yeah, I think that'd be a good idea :+1:

src/assets/js/main.js
```diff
@@ -88,16 +88,52 @@
...
+
+    var priceFormat = d3.format('.2f');
+
     var startPriceLine = fc.annotation.line()
         .orient('horizontal')
         .value(function(d) { return d.open; })
         .label(function(d) { return 'OPEN'; });
 
-    var endPriceLine = fc.annotation.line()
+    var closePriceLine = fc.annotation.line()
```
jleft: I can't see much value in both `closePriceLine` and `closeAxisAnnotation`, so I think we're good to remove `closePriceLine`?

djmiley:
`openPriceLine` feels a bit redundant to me, especially with respect to moving the data back further in time?

djmiley:
Keep the drawn close price dotted line?

jleft:
> openPriceLine feels a bit redundant to me, especially with respect to moving the data back further in time?

That's been mentioned before, so let's go with the consensus and remove it.

> Keep the drawn close price dotted line?

Yes, I think so.

jleft:
I expect, as `linearTimeSeries` plots the axis over the plot area, that the axis labels will overlap, above, the annotation's text?

jleft:
Resolves #69

PR #63, opened 2015-08-10
Feature/multi-series

RuthC-C:
Buttons added to change what series the chart is showing (e.g. Candlestick, OHLC, Line etc.)
Tiny bit of adjustment to css to include the buttons and line things up a little. 
Resolves #57

Issue #70, opened 2015-08-11
Modularise app


jleft:
- [x] Primary chart (#115)
- [x] Secondary chart (#116)
- [x] Navigator (#117)
- [x] Menu (#118)
- [x] Utilities (#119)
- [x] Resize (#120)
- [x] Zoom? (#121)
- [x] Data (#122)
- ~~Render?~~ (no longer relevant)

PR #78, opened 2015-08-11
Allow user to choose primary chart series type

### LINKED ISSUE #57: Change primary chart series types
Allow the user to toggle between different series types on the primary chart:

+ Candlestick
+ OHLC
+ Line
+ Area
+ Point

jleft:
Resolves #57 

(Resolved merge conflicts in #63)

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -100,7 +100,7 @@
                 currentSeries = candlestick;
                 break;
         }
-        multi.series([gridlines, ma, currentSeries, startPriceLine, endPriceLine]);
+        multi.series([gridlines, ma, currentSeries, closeAxisAnnotation]);
         render();
     }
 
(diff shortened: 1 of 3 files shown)


src/assets/js/main.js
```diff
@@ -71,6 +79,41 @@
 
     calculateDimensions();
 
+    function changeSeries(seriesTypeString) {
+        switch (seriesTypeString) {
+            case 'ohlc':
+                currentSeries = ohlc;
```
CBarrett2: Maybe could make this a little more concise by initializing ohlc on this line too, rather than before-hand? 
ie. currentSeries = fc.series.ohlc(); and similar for the others.
Aside from that, this looks perfect :)

jleft: Nice idea, but I'm not sure if we'd want to reinitialise the series each time we change the series?

jleft:
Let's revisit this approach in #70 and consider factoring into a component

src/assets/js/main.js
CBarrett2: Yeah, I think you're right :)

jleft: Ok, thanks for the feedback :)

Issue #59, opened 2015-08-10
Use Coinbase OHLC adapter as a data feed for the chart


jleft:

CBarrett2:
This can be closed now :)

PR #79, opened 2015-08-11
Feature/integrate live data

CBarrett2:
Integrates live data option into chart

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -30,8 +30,7 @@
             .attr('height', mainHeightRatio * targetHeight / totalHeightRatio);
         svgRSI.attr('width', targetWidth)
             .attr('height', rsiHeightRatio * targetHeight / totalHeightRatio);
-        svgNav.attr('width', targetWidth)
-            .attr('height', navHeightRatio * targetHeight / totalHeightRatio);
+        svgNav.attr('width', targetWidth);
     }
 
     var candlestick = fc.series.candlestick();
@@ -368,7 +367,8 @@
         var data = selection.datum();
 
         var xExtent = fc.util.extent(data, 'date');
-        var yExtent = fc.util.extent(getVisibleData(data, fc.util.extent(data, 'date')), ['low', 'high']);
+        var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data,
+            fc.util.extent(data, 'date')), ['low', 'high']);
         navTimeSeries.xDomain(xExtent)
             .yDomain(yExtent);
         area.y0Value(yExtent[0]);
@@ -407,8 +407,6 @@
         navTimeSeries.plotArea(navMulti);
         selection.call(navTimeSeries);
     };
-
-
 
     function render() {
         if (!currData.length) {
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -1,6 +1,16 @@
 (function(d3, fc, sc) {
     'use strict';
 
+    function padExtent(extent) {
+        // Adds a buffer of current period to either side of data
+        var period = ohlc.period();
+
+        extent[0] = d3.time.second.offset(new Date(extent[0]), -period);
+        extent[1] = d3.time.second.offset(new Date(extent[1]), +period);
+
+        return extent;
+    }
+
     // Set SVGs & column padding
     var container = d3.select('#chart-example');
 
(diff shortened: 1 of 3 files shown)


CBarrett2:
Resolves #89 and resolves #59

src/assets/js/main.js
```diff
@@ -11,6 +11,16 @@
         return visibleData;
     }
 
+    function padExtent(extent, period) {
```
jleft: Feels more like a util to me? Consider factoring this out into its own component?

CBarrett2: Like in a seperate file or as a seperate PR? I implemented this here so a single data point would display properly in the live data demo

jleft: > Like in a seperate file or as a seperate PR?

Both, as I believe this relates to more than just a single data point. e.g. the first and last data points are cut off in the current implementation #89

CBarrett2: Cool, I'll do that

src/assets/js/main.js
```diff
@@ -114,23 +102,133 @@
...
+        d3.select('#period-span').style('visibility', visibility);
+        d3.select('#product-span').style('visibility', visibility);
+    }
+
+    d3.select('#type-selection')
+        .on('change', function() {
+            var type = d3.select(this).property('value');
+            if (type === 'live') {
+                toggleLiveFeedUI(true);
+                currData = [];
```
tlsim: These 4 lines are repeated elsewhere - maybe extract into a function?

src/assets/js/main.js
```diff
@@ -337,14 +402,68 @@
         selection.call(navTimeSeries);
     };
 
+    function hideOrShowChartsAndSetLoadingText(hidden, text) {
```
tlsim: No need for the inner function - just call setLoadingText and toggleVisibility separately.

src/assets/js/main.js
```diff
@@ -114,23 +102,133 @@
...
 
     var gridlines = fc.annotation.gridline()
         .yTicks(5)
         .xTicks(0);
 
+    // Create and apply the Moving Average
+    var movingAverage = fc.indicator.algorithm.movingAverage();
+
+    // Create a line that renders the result
+    var ma = fc.series.line()
```
tlsim: movingAverageLine

src/index.html
```diff
@@ -48,8 +49,37 @@
...
+            <option value="fake">Data generator</option>
+            <option value="live">Live data</option>
+          </select>
+        </div>
+        <div class="col-md-2">
+          <span id="period-span">
+            <select id="period-selection" class="form-control">
+              <option value="43200">12hr</option>
+              <option value="3600">1hr</option>
+              <option value="50">1min</option>
```
tlsim: 60 seconds?

src/assets/js/main.js
```diff
@@ -114,23 +102,133 @@
...
         });
 
-    // Set Reset button event
+    var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
+    // Default to generated data
+    var currData = fc.data.random.financial()(250);
+
+    function liveCallback(event, latestBasket) {
+        if (!event && latestBasket) {
+            if (!currData.length) {
```
tlsim: Extract function for 'new basket received'.

src/index.html
```diff
@@ -48,8 +49,37 @@
...
+          
+          <select id="type-selection" selected="type" class="form-control">
+            <option value="fake">Data generator</option>
+            <option value="live">Live data</option>
+          </select>
+        </div>
+        <div class="col-md-2">
+          <span id="period-span">
+            <select id="period-selection" class="form-control">
+              <option value="43200">12hr</option>
```
tlsim: Might be worth extracting period selection into a component (future task).

CBarrett2:
Thanks for the feedback! :)

src/assets/js/main.js
```diff
@@ -239,13 +285,27 @@
...
+                    return data;
+            }
+        });
+
         // Scale y axis
         var yExtent = fc.util.extent(getVisibleData(data, timeSeries.xDomain()), ['low', 'high']);
         // Add 10% either side of extreme high/lows
         var variance = yExtent[1] - yExtent[0];
+        if (variance === 0) {
+            // Bit of a hack to get single-value candlesticks to show nicely.
```
tlsim: Looks like padExtent but for a non-time domain.

tlsim:
Good job - the 1 min live updated candles look nice!
All comments are fairly minor code style things.
Worth thinking about the data part of #70 after this I think.

src/assets/js/main.js
CBarrett2: That's a great idea

PR #81, opened 2015-08-11
Added dataInterface class

CBarrett2:
After this and the live websocket integration, it should be a short step to full live/historic integration

### DIFF
--- a/src/assets/js/dataInterface.js
+++ b/src/assets/js/dataInterface.js
@@ -1,0 +1,63 @@
+(function(sc) {
+    'use strict';
+
+    sc.data.dataInterface = function() {
+        // In seconds - maybe take this out and have chart pass in a period whenever needed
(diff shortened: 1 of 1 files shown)


src/assets/js/dataInterface.js
```diff
@@ -0,0 +1,63 @@
...
+            historicFeed(function(err, newData) {
+                if (id < n) { return; }
+                if (!err) {
+                    // To order from oldest to newest
+                    newData = newData.reverse();
+                    callback(null, newData);
+                } else { callback(err, null); }
+            });
+        };
+
```
CBarrett2: Might be worth keeping the ohlc-converter/live-feed seperate, and keep this as just a historic data class - then we can link them up in main.js or another class. Since all the class does with the live feed atm is rebind it!

CBarrett2:
Resolves #83

PR #82, opened 2015-08-11
Refactored getVisibleData into util class

djmiley:
#70 Is this the kind of thing we want to be doing?

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -1,15 +1,5 @@
-(function(d3, fc) {
+(function(d3, fc, sc) {
     'use strict';
-    function getVisibleData(data, dateExtent) {
-        // Calculate visible data, given [startDate, endDate]
-        var bisector = d3.bisector(function(d) { return d.date; });
-        var visibleData = data.slice(
-        // Pad and clamp the bisector values to ensure extents can be calculated
-            Math.max(0, bisector.left(data, dateExtent[0]) - 1),
-            Math.min(bisector.right(data, dateExtent[1]) + 1, data.length)
-        );
-        return visibleData;
-    }
 
     // Set SVGs & column padding
     var container = d3.select('#chart-example');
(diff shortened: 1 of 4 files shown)


jleft:
> Is this the kind of thing we want to be doing?

Yes, definitely! The utils are a great place to start too.

If/when this is ready for more formal feedback, let me know and I'll have a look over it.

CBarrett2:
perfect!

src/assets/js/util.js
```diff
@@ -0,0 +1,15 @@
+(function(d3, fc) {
```
CBarrett2: fc isn't used here

djmiley: should we assume it probably will be, for ease and clarity? it can be changed easily also.

src/assets/js/util.js
```diff
@@ -0,0 +1,15 @@
...
+    'use strict';
+
+    sc.util.returnDataExtent = function(dataValues, dateExtent) {
```
jleft: Maybe `filterDataInDateRange`?

src/assets/js/util.js
```diff
@@ -0,0 +1,15 @@
...
+        // Calculate visible data, given [startDate, endDate]
+        var bisector = d3.bisector(function(d) { return d.date; });
+        var visibleData = dataValues.slice(
```
jleft: `visibleData` -> `filteredData`? Does _filter_ seem like the appropriate verb to use?

jleft:
I think moving the component to its own file (i.e. not having all utils in the same file) and having the namespaces mirror the directory structure would be good.

`src/assets/js/util.js` could change to `src/assets/js/util/componentName.js`

djmiley:
Filter seems a good verb.

djmiley:
Don't merge!

CBarrett2:
Can we pull this in please?

PR #84, opened 2015-08-12
Set up framework for unit tests

CBarrett2:


src/assets/js/ohlc.js
```diff
@@ -29,7 +37,7 @@
 
         function updateBasket(basket, datum) {
             if (basket == null) {
```
djmiley: What's happening here?

src/assets/js/ohlc.js
```diff
@@ -29,7 +37,7 @@
...
-                basket = createNewBasket(datum, datum.date);
+                return createNewBasket(datum, datum.date);
             }
```
djmiley: What's happening here?

djmiley:
Looks good to me otherwise.

src/assets/js/ohlc.js
CBarrett2: returning a new basket based on the single datum.

djmiley: Ok. I see what is happening now.

Gruntfile.js
```diff
@@ -210,8 +210,25 @@ module.exports = function(grunt) {
...
+        jasmine: {
+            options: {
+                specs: '<%= meta.testJsFiles %>',
+                vendor: ['node_modules/jquery/dist/jquery.min.js',
+                        'node_modules/d3fc/node_modules/d3/d3.js',
+                        'node_modules/d3fc/node_modules/css-layout/src/Layout.js',
+                        'node_modules/d3fc/dist/d3fc.js']
+            },
+            test: {
+                src: ['src/assets/js/sc.js',
```
janakerman: Rather than listing your source files explicitly, could you use `src/**/*.js`? It looks like this is already defined as variable on line 9 so I think you could just do use `'<%= meta. srcJsFiles %>'`.

janakerman:
I think it would be good to add an entry into the projects README about the test task.

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    describe('sc.data.feed.coinbase.ohlcWebSocketAdaptor', function() {
```
janakerman: Picky one - I wouldn't use the components full path - I'd just use the name of the component you are testing, so in this case 'ohlc'. Considering that though, it looks as though `ohlcWebSocketAdaptor` would be a better name for the component itself!

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+                callback(event, null);
+            };
+
+            return feed;
+        }
+
+        var ohlc;
+
+        beforeEach(function() {
+            ohlc = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
```
janakerman: The context of `this` refers to a blank object in Jasmine beforeEach and it functions. This means you can attach things to it and remove variables like `ohlc` on line 23.

    beforeEach(function() {
        this.ohlc = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
    };

    it('...', function() {
        this.ohlc.period ....;
    });

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+        });
+
+        it('should have a working getter/setter for period', function() {
+            ohlc.period(60 * 60 * 24);
+            expect(ohlc.period()).toEqual(60 * 60 * 24);
+        });
+
+        it('should have working getter/setter for liveFeed', function() {
+            var feed = testFeed();
```
janakerman: Anything I create in my unit tests that isn't a 'real' component I tend to prefix with the word 'mock' so it's really clear at a glance that the object is only used for the test scenario.

test/ohlcSpec.js
janakerman: A note on the ohlcWebSocketAdaptor construction - rather than creating the `liveFeed` getters and setters, I think it would be a good idea to pass the feed in as a parameter to it's initialisation. As we discussed, it doesn't seem likely to change, and creating a new one would probably be nicer anyway.

An addition to what we discussed though - could the period be passed in too? I don't see any point in having the setter unless it's needed.

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+            ohlc.liveFeed(feed);
+            expect(ohlc.liveFeed()).toEqual(feed);
+        });
+
+        it('should pass events to callback when one is triggered in the live feed', function() {
```
janakerman: 'should pass events to callback when **one** is triggered in the live feed' - might be better to specify what one is.

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+            var feed = testFeed();
+            ohlc.liveFeed(feed);
+            var eventType = '';
```
janakerman: I'd just define eventType like so:
    var eventType;

It's pretty easy to see what's happening in this usage, but you can imagine how in a more complex block of code, the empty string could become confusing. At a glance you may wonder if it has any significance.

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+            ohlc(function(event, basket) {
+                eventType = event.type;
+            });
+
+            feed.sendEvent('open');
+            expect(eventType).toEqual('open');
+        });
+
+        it('should pass the latest basket of data collected to callback' +
```
janakerman: I wouldn't worry about the line length here, just let it flow on! (Unless JSHint moans then leave it as is - it's probably not worth setting up different JSHint configs for tests and source.)

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+            'whenever a new datum is pushed by the live feed', function() {
+            var feed = testFeed();
+            ohlc.liveFeed(feed);
+            var called = false;
+            ohlc(function(event, basket) {
```
janakerman: I'd test the specific value here rather than just testing it is called. If you think that doesn't make sense, then I'd just assign the basket and check that it's truthy:

        var basket = false;
        ohlc(function(event, basket) {
            basket = true;
        });

        expect(basket).toBeTruthy();

test/ohlcSpec.js
```diff
@@ -0,0 +1,178 @@
...
+            };
+
+            feed.sendDatum(datum);
+            expect(called).toBeTruthy();
+        });
+
+        it('should initialize a new basket when the first datum is pushed by live feed', function() {
+            var feed = testFeed();
+            ohlc.liveFeed(feed);
+            var currBasket = {};
```
janakerman: Same again here `var currBasket;`

Gruntfile.js
CBarrett2: It seems important that the files are loaded in a specific order, unfortunately. Not sure how best to solve this

janakerman: Leave this for now - I'll raise an issue in general about the ordering dependancy, it's a bigger issue than this change.

CBarrett2: Thanks

janakerman:
Give me a shout when you've had a think about my comments and we can look at getting this merged.

PR #86, opened 2015-08-12
Removed ticks from nav

djmiley:
Removed ticks from navigator, in response to Rob's comments.


jleft:
:+1: 

Resolves #93

CBarrett2:
+1

Issue #89, opened 2015-08-12
Add padding to extent of data domain


jleft:
The first and last bars of the chart are cut off - we should add some padding, so they can be fully displayed.

![extent-padding](https://cloud.githubusercontent.com/assets/2017615/9224960/3c49b26a-40fe-11e5-9473-9a01867b90fb.PNG)

djmiley:
We need to pad for candlestick, OHLC and area series.

PR #99, opened 2015-08-12
fixed bug, added padExtent to utils

CBarrett2:


CBarrett2:
Resolves #89

jleft:
Do we need to consider other scale types? What if we wanted to pad a linear scale? Maybe not for now...

Any thoughts on a function like this, with an, optional, padding param? (Could use `fc.util.extent` internally.)
```js
// sc.util.extent(data, fields, padding); // For example:
sc.util.extent(data, ['date'], 60 * 60 * 24);
```

CBarrett2:
I like both of those ideas @jleft, maybe a new issue once this is pulled? Also perhaps an option to pad by percentage rather than an absolute value?

src/assets/js/main.js
```diff
@@ -189,7 +190,7 @@
             var tx = zoom.translate()[0];
             var ty = zoom.translate()[1];
 
-            var xExtent = fc.util.extent(data, ['date']);
+            var xExtent = sc.util.findPaddedExtent(fc.util.extent(data, ['date']), 60 * 60 * 24);
```
jleft: Will there be excess whitespace at the start and end of the chart when using a line/area/point series?

src/assets/js/main.js
```diff
@@ -262,8 +263,9 @@
 
     // Create navigation chart
     var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data, fc.util.extent(data, 'date')), ['low', 'high']);
+    var xExtent = sc.util.findPaddedExtent(fc.util.extent(data, 'date'), 60 * 60 * 24);
```
jleft: Will this create a gap between the extremes of the series and the edge of the navigator?

src/index.html
```diff
@@ -69,8 +69,11 @@
     </div>
     <script src="assets/js/sc.js"></script>
     <script src="assets/js/util/filterDataInDateRange.js"></script>
+    <script src="assets/js/util/findPaddedExtent.js"></script>
     <script src="assets/js/data/feed/coinbase/webSocket.js"></script>
     <script src="assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js"></script>
+    <script src="assets/js/webSocket.js"></script>
```
jleft: Remove.

src/index.html
```diff
@@ -69,8 +69,11 @@
...
+    <script src="assets/js/ohlc.js"></script>
```
jleft: Remove.

src/assets/js/util/findPaddedExtent.js
```diff
@@ -0,0 +1,10 @@
+(function(d3, fc) {
+    'use strict';
+    sc.util.findPaddedExtent = function(extent, period) {
```
jleft: I'm not sure if this is really _finding_ the padded extent, maybe it's _offsetting_ the extent? I quite liked _padExtent_ too... maybe I'm being too pedantic!

src/assets/js/main.js
CBarrett2: There's some excess whitespace on the nav bar, at either end, but none on the main chart

src/assets/js/main.js
```diff
@@ -280,8 +292,18 @@
...
+                    var extent = fc.util.extent(data, ['date']);
+                    var earliest = brush.extent()[0][0];
+                    var latest = brush.extent()[1][0];
+                    /*if (brush.extent()[1][0].getTime() === extent[1].getTime()) {
+                        earliest = new Date(earliest.getTime() - bufferPeriod);
+                    }
+                    if (brush.extent()[0][0].getTime() === extent[0].getTime()) {
+                        latest = new Date(latest.getTime() + bufferPeriod);
+                    }*/
+
```
djmiley: Removed bracketed code.

src/assets/js/main.js
```diff
@@ -262,8 +273,9 @@
 
     // Create navigation chart
     var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data, fc.util.extent(data, 'date')), ['low', 'high']);
+    var xExtent = fc.util.extent(data, 'date');
     var navTimeSeries = fc.chart.linearTimeSeries()
-        .xDomain(fc.util.extent(data, 'date'))
+        .xDomain(xExtent)
         .yDomain(yExtent)
```
djmiley: What do we gain with this? Also maybe `xDomain` before `yDomain`

src/assets/js/main.js
```diff
@@ -92,7 +100,8 @@
 
     // Set Reset button event
     function resetToLive() {
-        timeSeries.xDomain(standardDateDisplay);
+        var extent = sc.util.paddedExtent(standardDateDisplay, bufferPeriod);
+        timeSeries.xDomain(extent);
         render();
```
djmiley: `extent` -> `paddedExtent`?

src/assets/js/main.js
```diff
@@ -100,7 +109,6 @@
 
     // Create main chart and set how much data is initially viewed
     var timeSeries = fc.chart.linearTimeSeries()
-        .xDomain(standardDateDisplay)
         .xTicks(6);
```
djmiley: Why remove this line?

CBarrett2: It paves the way for removing standardDateDisplay as a global variable.

djmiley: Ok. Yes, I see what's happening here.

Issue #62, opened 2015-08-10
Resize chart to parent container


jleft:
- [ ] Define layout of page in CSS
- [ ] Update resizing algorithm

PR #102, opened 2015-08-13
Altered max aspect ratio

### LINKED ISSUE #95: The chart should fill the screen on mobile devices
The chart should fill to the edge of the screen and excess white space should be removed:

excess-whitespace

djmiley:
Resolves #95

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -29,22 +29,8 @@
         var useableScreenHeight = window.innerHeight - headRowHeight - navHeight -
             2 * fc.chart.linearTimeSeries().xAxisHeight();
 
-        var targetWidth;
-        var targetHeight;
-
-        var maxWidthToHeightRatio = 1.5;
-        var maxHeightToWidthRatio = 1.5;
-
-        if (useableScreenWidth > maxWidthToHeightRatio * useableScreenHeight) {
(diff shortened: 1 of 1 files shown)


djmiley:
Changed max aspect ratio from 1.5 to 16 / 9, which seems to be maximum aspect ratio for mobile devices.

src/assets/js/main.js
```diff
@@ -30,8 +30,8 @@
         var targetWidth;
         var targetHeight;
 
-        var maxWidthToHeightRatio = 1.5;
-        var maxHeightToWidthRatio = 1.5;
+        var maxWidthToHeightRatio = 16 / 9;
+        var maxHeightToWidthRatio = 16 / 9;
```
jleft: Rather than assuming this aspect ratio, could we calculate it from  `useableScreenWidth` and `useableScreenHeight`?

djmiley: I possibly ought to rename this, as this refers to the maximum allowed aspect ratio for the charts (although ideally this would be for the container).

Investigation in progress...

djmiley:
Removed `maxHeightToWidthRatio` and `maxWidthToHeightRatio`. Works well without them, as Bootstrap holds the page together well.

djmiley:
Don't merge. Got a bug in here

jleft:
Looks good!

(Will continue this in #62)

PR #106, opened 2015-08-13
Moved data files

djmiley:
#70 Moved data files to support suggested file structure.


RuthC-C:
Good idea.

jleft:
LGTM.

Will merge when conflicts are resolved.

PR #107, opened 2015-08-13
Alternative live data integration

CBarrett2:
This is a bit of an experiment, but in consideration of @tlsim 's comment about refactoring the data out being the next task after the live integration is pulled in, I tried taking as much of the data code out as I could. The button logic is still there, but the ugliest callback stuff is pulled out into a class with onError, onOpen, onBasketReceived etc methods exposed. It would be very easy to then integrate historic data into this set up too.

### DIFF
--- a/src/assets/js/dataInterface.js
+++ b/src/assets/js/dataInterface.js
@@ -1,0 +1,70 @@
+(function(sc) {
+    'use strict';
+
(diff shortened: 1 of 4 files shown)


CBarrett2:
This doesn't want pulling in before the paddedExtent stuff though

jleft:
Alternative to #79

PR #109, opened 2015-08-13
Fixed bug of area white view

### LINKED ISSUE #91: Area series y0 value (baseline) can appear above the x-axis
This leaves a gap, as shown below (highlighted in yellow); this gap should be removed, and the bottom of the area series should meet the x-axis.

area-series-baseline

djmiley:
Resolves #91.


RuthC-C:
Looks fine

CBarrett2:
This would also have the effect of *only just* fitting the data in on the y-axis, I'm not sure this is desirable?

PR #112, opened 2015-08-13
Modularized three main charts

CBarrett2:
This isn't to pull quite yet, but spent some time splitting the charts up into different files. I've used events to communicate changes in view and when to render. Thoughts?

### DIFF
--- a/src/assets/js/chart/navChart.js
+++ b/src/assets/js/chart/navChart.js
@@ -1,0 +1,72 @@
+(function(sc) {
+    'use strict';
+    // Helper functions
+    sc.chart.navChart = function() {
+
+        var navTimeSeries = fc.chart.linearTimeSeries()
+            .yTicks(0);
+
+        var area = fc.series.area()
+            .yValue(function(d) { return d.open; });
+
+        var line = fc.series.line()
+            .yValue(function(d) { return d.open; });
+
+        var brush = d3.svg.brush();
+
+        var navMulti = fc.series.multi().series([area, line, brush]);
+
+        var viewScale = fc.scale.dateTime();
+
+        function navChart(selection) {
+            var data = selection.datum();
+
+            viewScale.domain(data.viewDomain)
+                .range([0, selection.attr('width')]);
+
+            var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data,
+                fc.util.extent(data, 'date')), ['low', 'high']);
+
+            navTimeSeries = fc.chart.linearTimeSeries()
+                .xDomain(fc.util.extent(data, 'date'))
+                .yDomain(yExtent);
+
+            brush.on('brush', function() {
+                if (brush.extent()[0][0] - brush.extent()[1][0] !== 0) {
+                    // Control the shared view scale's domain
+                    sc.dispatch.viewChange([brush.extent()[0][0], brush.extent()[1][0]]);
+                }
+            });
+
+            // Allow to zoom using mouse, but disable panning
+            var zoom = d3.behavior.zoom();
+            zoom.x(viewScale)
+                .on('zoom', function() {
+                    if (zoom.scale() === 1) {
+                        zoom.translate([0, 0]);
+                    } else {
+                        // Usual behavior
+                        sc.zoomCall(zoom, data, viewScale)();
+                    }
+                });
+            selection.call(zoom);
+
+
+            navMulti.mapping(function(series) {
+                if (series === brush) {
+                    brush.extent([
+                        [viewScale.domain()[0], navTimeSeries.yDomain()[0]],
+                        [viewScale.domain()[1], navTimeSeries.yDomain()[1]]
+                    ]);
+                }
+                return data;
+            });
+
+            navTimeSeries.plotArea(navMulti);
+            selection.call(navTimeSeries);
+        }
+
+        return navChart;
+    };
+
+})(sc);
--- a/src/assets/js/chart/navChart.js
+++ b/src/assets/js/chart/navChart.js
@@ -46,7 +46,7 @@
                         zoom.translate([0, 0]);
                     } else {
                         // Usual behavior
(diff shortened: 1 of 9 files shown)


djmiley:
#70

CBarrett2:
Resolves #115 #116 and #117 plus sets up an events system that should make some other modularizations easier

CBarrett2:
Would appreciate another look-over of this :)

src/assets/js/sc.js
```diff
@@ -13,6 +13,7 @@
                 coinbase: {}
             }
         },
-        util: {}
+        util: {},
+        chart: {}
     };
```
djmiley: Alphabetical order?

src/assets/js/main.js
```diff
@@ -61,75 +69,25 @@
                 .node()
                 .value;
             changeSeries(seriesTypeString);
+            render();
         });
```
djmiley: Is there a reason you've changed `changeSeries`?

src/assets/js/main.js
```diff
@@ -12,23 +12,30 @@
     var ohlc = fc.series.ohlc();
     var point = fc.series.point();
     var line = fc.series.line();
+    line.isLine = true;
```
djmiley: My menu request should solve this hopefully.

src/assets/js/main.js
CBarrett2: It no longer has access to the multi inside primaryChart, so it has to go through primaryChart.changeSeries now.

src/assets/js/main.js
CBarrett2: I saw your solution to that, I like it!

src/assets/js/main.js
djmiley: Yes, that's good.

src/assets/js/chart/navChart.js
```diff
@@ -0,0 +1,78 @@
...
+                return data;
+            });
+
+            navTimeSeries.plotArea(navMulti);
+            selection.call(navTimeSeries);
+        }
+
+        navChart.onViewChange = function(func) {
+            dispatch.on('viewChange', func);
+        };
```
djmiley: `viewChange.navChart` ?

src/assets/js/chart/primaryChart.js
```diff
@@ -0,0 +1,122 @@
...
+            dispatch.on('viewChange.primaryChart', func);
+        };
+
+        primaryChart.changeSeries = function(series) {
+            multi.series([gridlines, movingAverageLine, series, closeAxisAnnotation]);
+            return primaryChart;
+        };
+
+        d3.rebind(primaryChart, timeSeries, 'xDomain');
+
```
djmiley: Is this needed? Have a check.

djmiley:
'viewChange' -> 'viewChanged'? Other than that, looks good.

src/assets/js/chart/primaryChart.js
```diff
@@ -0,0 +1,122 @@
...
+            });
+
+            // Scale y axis
+            var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data, timeSeries.xDomain()), ['low', 'high']);
+            // Add 10% either side of extreme high/lows
+            var variance = yExtent[1] - yExtent[0];
+            yExtent[0] -= variance * 0.1;
+            yExtent[1] += variance * 0.1;
+            timeSeries.yDomain(yExtent);
+
```
djmiley: Also remove lines 88 to 91. #91 #109

src/assets/js/chart/navChart.js
```diff
@@ -0,0 +1,78 @@
...
+                }
+                return data;
+            });
+
+            navTimeSeries.plotArea(navMulti);
+            selection.call(navTimeSeries);
+        }
+
+        navChart.onViewChange = function(func) {
+            dispatch.on('viewChange.navChart', func);
```
tlsim: Does the namespace make any difference here?

CBarrett2: I don't think so

src/assets/js/main.js
```diff
@@ -12,23 +12,30 @@
...
+    var navChart = sc.chart.navChart();
+
+    function onViewChange(domain) {
+        data.viewDomain = [domain[0], domain[1]];
+        render();
+    }
 
-    var standardDateDisplay = [data[Math.floor((1 - navAspect * goldenRatio) * data.length)].date,
-        data[data.length - 1].date];
+    primaryChart.onViewChange(onViewChange);
```
tlsim: prefer `primaryChart.on('viewChange' onViewChanged)` so it's clear that it's using d3 events.

src/assets/js/main.js
```diff
@@ -12,23 +12,30 @@
...
+    line.isLine = true;
     var area = fc.series.area();
 
-    var currentSeries = candlestick;
-
     var data = fc.data.random.financial()(250);
 
-    // Using golden ratio to make initial display area rectangle into the golden rectangle
-    var goldenRatio = 1.618;
+    data.viewDomain = [];
```
tlsim: Would be better not to store this on the data array itself. Bind a model with both `data` and `viewDomain` instead.

tlsim:
Definitely moving in the right direction. Seeing it split up certainly makes things easier to follow! I'm wondering if rather than each chart creating a new 'viewScale' and using events to communicate the domain change, they could instead share one instance of an xScale? It's also surprising to see 3 d3.behavior.zoom components, but we can work on that later in #121.

CBarrett2:
viewScale is only created in navChart, and I think that requires the range being navChart's width, as opposed to the width of any other chart. The rest just use the domain passed to them. It's possible for all of them to share the same scale, but I thought we agreed to do it this way?

tlsim:
@CBarrett2 Ok we can think about if shared scales make any sense later.

PR #113, opened 2015-08-14
Refactored zoomControl to util class

djmiley:
#70

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -184,32 +184,9 @@
             }
         });
 
-    function zoomCall(zoom, data, scale) {
+    function zoomCall(zoom, selection, data, scale) {
         return function() {
-            var tx = zoom.translate()[0];
-            var ty = zoom.translate()[1];
-
-            var xExtent = fc.util.extent(data, ['date']);
-            var min = scale(xExtent[0]);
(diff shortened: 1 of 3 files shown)


src/assets/js/main.js
```diff
@@ -184,32 +184,9 @@
...
-            if (zoom.scale() !== 1) {
-                if ((min >= 0) && (max - width) <= 0) {
-                    scale.domain(xExtent);
-                    zoom.x(scale);
-                    tx = scale(xExtent[0]);
-                }
-            }
-
-            zoom.translate([tx, ty]);
+            sc.util.zoomControl(zoom, selection, data, scale);
```
tlsim: This returns a function - when is it called?

src/assets/js/util/zoomControl.js
```diff
@@ -0,0 +1,33 @@
+(function(d3, fc) {
+    'use strict';
+
+    sc.util.zoomControl = function(zoom, selection, data, scale) {
+        return function() {
```
tlsim: Do this need to be wrapped in a function?

src/assets/js/main.js
djmiley: Unwrapped it from function in external file.

PR #123, opened 2015-08-14
Refactored resizing method into own file

djmiley:
#120


djmiley:
Passed in container, rather than container and the three SVGs.

djmiley:
i'll sort this out

Issue #118, opened 2015-08-14
Modularise menu


jleft:

djmiley:
Is this possible without #65 ?

jleft:
> Is this possible without #65 ?

Yes, I think we can modularise the existing menu. For example, we could handle the `click` events in a menu component and provide a more abstract event for them and propagate this to the top level - such as a _series changed_ event.

However, as #65 is in progress, I think it makes sense to do this after #65? :smiley:

djmiley:
I misspoke, you're right. :) I think I meant to say it is conceptually easier for me to do this after #65.

jleft:
> I misspoke, you're right. :) I think I meant to say it is conceptually easier for me to do this after #65.

Cool, no problem, let's do this after #65 is closed.

janakerman:
@djmiley resolved with #125.

jleft:
Continued in #157 #194

PR #125, opened 2015-08-17
Refactored menu into own namespace

djmiley:
#118 

Set initial active series in javascript. 

Moved switch statement into external class.


CBarrett2:
This looks good

RuthC-C:
great - just wondering if there's a way we can also move the buttons? From html as well maybe?

djmiley:
I'll work on doing so.

djmiley:
WIP

src/assets/js/chart/primaryChart.js
```diff
@@ -99,23 +104,24 @@
 
         d3.rebind(primaryChart, dispatch, 'on');
 
-        primaryChart.changeSeries = function(series, indicator) {
-            if (indicator == null) {
-                multi.series([gridlines, series, closeAxisAnnotation]);
+        function updateMultiSeries() {
+            if (currentIndicator === null) {
```
RuthC-C: I would leave this as == then it can work for 'void/undefined(?)' too

src/assets/js/menu/selectIndicator.js
```diff
@@ -0,0 +1,48 @@
+(function(d3, fc) {
+    'use strict';
+
+    var IndicatorType = function(displayString, valueString, indicator) {
+        this.displayString = displayString;
+        this.valueString = valueString;
+        this.indicator = indicator;
+    };
+
+    var ma = fc.series.line()
```
RuthC-C: change 'ma' to 'movingAverage' for consistency

RuthC-C:
Great work, really helpful thanks Drew :+1:

src/assets/js/menu/selectSeries.js
```diff
@@ -0,0 +1,44 @@
...
+    var ohlc = new SeriesType('OHLC', 'ohlc', fc.series.ohlc());
+    var line = new SeriesType('Line', 'line', fc.series.line());
+    var point = new SeriesType('Point', 'point', fc.series.point());
+    var area = new SeriesType('Area', 'area', fc.series.area());
+
+    var seriesTypeArray = [candlestick, ohlc, line, point, area];
+
+    var seriesButtons = d3.select('#series-buttons');
+
+    seriesButtons.selectAll('label')
```
janakerman: You can chain this with the declaration above (line 18).

djmiley: Yes, good suggestion! :+1:

src/assets/styles/chartLayout.less
```diff
@@ -0,0 +1,22 @@
+.plot-area, .background {
```
djmiley: Delete this file.

src/assets/js/menu/selectSeries.js
```diff
@@ -0,0 +1,43 @@
...
+    var point = new SeriesType('Point', 'point', fc.series.point());
+    var area = new SeriesType('Area', 'area', fc.series.area());
+
+    var seriesTypeArray = [candlestick, ohlc, line, point, area];
+
+    d3.select('#series-buttons')
+        .selectAll('label')
+        .data(seriesTypeArray)
+        .enter()
+        .append('label')
```
djmiley: var selectSeries = sc.menu.selectSeries();

var seriesTypeArray = [..];

var seriesButtonsContainer = d3.select('#series-buttons')
  .series(seriesTypesArray)
  .call(selectSeries)

src/assets/js/chart/primaryChart.js
janakerman: You can just check that it's falsey: `if (!currentIndicator) { ... }`. I believe that null & undefined and coerced into booleans for boolean comparisons & logical statements.

src/assets/js/menu/selectSeries.js
```diff
@@ -0,0 +1,44 @@
+(function(d3, fc) {
```
janakerman: I think it would be nice to follow the D3 component pattern here. For example, `sc.menu.selectSeries` be a function that can be ran on a selection. Your selection would still be `d3.select('#series-buttons')` but it would be done outside of this file and you'd call this component on your selection:

Related to this, you could have a D3 pattern getter/setter on the component for the series, and you could also create the series objects outside of this file. This would result in a more flexible and reusable component.

The usage of the component would look something like this (maybe in `main.js`):

    ...

    sc.menu.selectSeries()
        .series([candlestick, ohlc, line, point, area]);

    d3.select('#series-buttons')
        .call(selectSeries);

src/assets/js/menu/selectIndicator.js
```diff
@@ -0,0 +1,48 @@
+(function(d3, fc) {
```
janakerman: Same comments here as for `selectSeries.js` below.

src/assets/js/main.js
```diff
@@ -45,70 +28,27 @@
...
-        }
-        primaryChart.changeSeries(currentSeries, currentIndicator);
-    }
-
-    changeSeries('candlestick');
-
-    d3.select('#series-buttons')
+    container.select('#series-buttons')
         .selectAll('.btn')
         .on('click', function() {
```
janakerman: As we discussed, I think pulling the event handler creation into the component functions would be a good idea. You'll always want a selection button to do something when it is clicked so the buttons are almost useless in isolation.

This bring on the question - how would you get the chart to re-render? Would the `seriesSelect` & `indicatorSelect` components need a reference to the chart? They would, and that's not very nice either as they are then coupled to the chart. Instead, we could use D3 events within the `seriesSelect` & `indicatorSelect` components that basically just notify anyone who cares that the selected series / indicator has updated, and pass the relevant `IndicatorType` or `SeriesType` as extras to the event.

Then, in `main.js`, or where ever appropriate, the events can be listened for, and the chart can be re-rendered with the new series/indicators.

janakerman:
Great first attempt at pulling out the series / indicators. Have a go at the changes I highlighted in my comments, we can then see if you like the approach and take it from there.

src/assets/js/main.js
```diff
@@ -8,34 +8,20 @@
...
-
-    var bollinger = fc.indicator.renderer.bollingerBands();
-
     sc.util.calculateDimensions(container);
 
     var primaryChart = sc.chart.primaryChart();
     var rsiChart = sc.chart.rsiChart();
     var navChart = sc.chart.navChart();
 
+    var seriesOptions = sc.menu.optionGenerator();
```
janakerman: Minor point, and I don't mind that you're creating two objects here, but you could reuse the same options generator! They have no differences in configuration (are you're using `.datum()`) and the component has no state.

src/assets/js/main.js
```diff
@@ -45,72 +31,54 @@
...
-        .selectAll('.btn')
-        .on('click', function() {
-            var indicatorType = d3.select(this)
-                .select('input')
-                .node()
-                .value;
-            changeIndicator(indicatorType);
-            render();
-        });
+    seriesOptions.on('optionChange', onSeriesChanged);
```
janakerman: I think it would be nice to set these on events in the same location that you create the components. Also, since you don't re-use the `onIndicatorChanged` and `onSeriesChanged` functions I'd just have them as anonymous functions. Due to the D3 syntax it would be more readable and all the set up is in one place:

    var seriesOptions = sc.menu.optionGenerator()
        .on('optionChange', function() {
        // Do things!
    });

src/assets/js/main.js
```diff
@@ -45,72 +31,54 @@
...
+    var ohlc = new SeriesType('OHLC', 'ohlc', fc.series.ohlc());
+    var line = new SeriesType('Line', 'line', fc.series.line());
+    var point = new SeriesType('Point', 'point', fc.series.point());
+    var area = new SeriesType('Area', 'area', fc.series.area());
+
+    container.select('#series-buttons')
+        .datum([candlestick, ohlc, line, point, area])
+        .call(seriesOptions);
+
+    var IndicatorType = function(displayString, valueString, indicator) {
```
janakerman: Note for the future - we may want to pull our model objects into their own files. Same for line 34. I may raise this as an improvement issue. Don't worry about it on this PR.

src/assets/js/chart/primaryChart.js
```diff
@@ -53,12 +56,14 @@
             });
 
         var multi = fc.series.multi()
-            .series([gridlines, closeAxisAnnotation])
+            .series([gridlines, fc.series.candlestick(), closeAxisAnnotation])
```
janakerman: Is this setting it up the multi with a default series? You could use `currentSeries ` and you'd only need to change it in one place if the default series changed at any point in the future.

janakerman:
This is looking great by the way. Just a few small opinionated comments to discuss and then I think this is good to be merged.

I may raise an improvement to pull the menu set up out of `main.js` and in to a component of it's own. I think this would really tidy up the main file.

PR #126, opened 2015-08-17
Removed id from container as it seems misleading

djmiley:
Using the `chart-example` id for the container seems misleading now.

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -2,7 +2,7 @@
     'use strict';
 
     // Set SVGs & column padding
-    var container = d3.select('#chart-example');
+    var container = d3.select('#app-container');
 
     var svgMain = container.select('svg.primary');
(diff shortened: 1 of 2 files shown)


CBarrett2:
This is good

src/assets/js/main.js
```diff
@@ -2,7 +2,7 @@
     'use strict';
 
     // Set SVGs & column padding
-    var container = d3.select('#chart-example');
```
tlsim: The old way makes sense because the `container` class is bootstrap specific, and we want a singleton chart. Better to rename 'chart-example' if it is misleading.

djmiley:
The container contains everything though, not just the chart?

djmiley:
Something like `#app-container`?

tlsim:
Yeah sorry, I meant the whole thing including menu for 'chart'. `app-container` is a good name for it!

PR #127, opened 2015-08-17
Removed svg selection from main.js

djmiley:
To synchronise with #124, #62


CBarrett2:
Not sure I see why this is necessary?

djmiley:
You're right, not strictly relevant at moment.

PR #128, opened 2015-08-17
Fixed latest price annotation mismatch

### LINKED ISSUE #92: Does the latest price on the area series and annotation align properly?
Sometimes it appears that the latest price annotation and the latest price shown on the area series don't match:

area-series-latest-price

djmiley:
On line and area series

Resolves #92


djmiley:
Resolved by #112 ?

Issue #83, opened 2015-08-11
Use Coinbase Historic feed as a data feed for the chart


jleft:
Allow the user to select:
- product
- period

CBarrett2:
This can be closed now :)

PR #131, opened 2015-08-18
Integrate historic data

CBarrett2:
Added an option for basic historic data, no period/product selectors atm but they would be simple to add afterwards. This also makes it pretty simple to add in support for a live stream. There is a button to switch between 'live' and generated data, and a class for the historic feed which adds the ability to cancel a request that's been made (invalidateCallback)

### DIFF
--- a/src/assets/js/data/feed/coinbase/historicFeed.js
+++ b/src/assets/js/data/feed/coinbase/historicFeed.js
@@ -2,7 +2,6 @@
     'use strict';
 
     sc.data.feed.coinbase.historicFeed = function() {
-        // In seconds - maybe take this out and have chart pass in a period whenever needed
         var feed = fc.data.feed.coinbase();
         var n = 0;
 
--- a/src/assets/js/data/feed/coinbase/historicFeed.js
+++ b/src/assets/js/data/feed/coinbase/historicFeed.js
@@ -1,0 +1,30 @@
+(function(sc) {
+    'use strict';
+
+    sc.data.feed.coinbase.historicFeed = function() {
+        // In seconds - maybe take this out and have chart pass in a period whenever needed
+        var feed = fc.data.feed.coinbase();
+        var n = 0;
+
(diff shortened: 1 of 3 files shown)


CBarrett2:
Resolves #83

src/assets/js/data/feed/coinbase/historicFeed.js
```diff
@@ -0,0 +1,29 @@
+(function(sc) {
+    'use strict';
+
+    sc.data.feed.coinbase.historicFeed = function() {
```
tlsim: Not sure we need this extra class. Is `invalidateCallback` intended to stop a slow to return coinbase callback rendering if the user switches back to generated data? Could main.js `historicCallback` check a boolean instead?

CBarrett2: Yes, that's right :) Possibly, although I don't think that's quite as sturdy - what if someone tries to load live data, changes back to generated, then back to live again, and for some reason the first call comes back after the second? Unlikely, perhaps. We could probably move all this to main.js for now if you like?

CBarrett2: Also a boolean value doesn't cover if you want to potentially make multiple historic requests - as in when we get to changing period/instrument - we'll want the option to change quickly between periods/instruments etc, so I think the integer is better.

src/assets/js/main.js
```diff
@@ -121,9 +121,37 @@
...
+    d3.select('#type-selection')
+        .on('change', function() {
+            var type = d3.select(this).property('value');
+            if (type === 'live') {
+                historicFeed(historicCallback);
+            } else if (type === 'generated') {
+                historicFeed.invalidateCallback();
+                dataModel.data = fc.data.random.financial()(250);
+                resetToLive();
+                render();
```
CBarrett2: Pull out as method

PR #132, opened 2015-08-18
Integrate Live data with historic

### LINKED ISSUE #53: Coinbase Data aggregation
Aggregate data from historic and live Coinbase data sources.

CBarrett2:
Adds live data integration to the previous PR #131. Has a period of one minute.
Resolves #53

### DIFF
--- a/src/assets/js/data/feed/coinbase/historicFeed.js
+++ b/src/assets/js/data/feed/coinbase/historicFeed.js
@@ -2,7 +2,6 @@
     'use strict';
 
     sc.data.feed.coinbase.historicFeed = function() {
-        // In seconds - maybe take this out and have chart pass in a period whenever needed
         var feed = fc.data.feed.coinbase();
         var n = 0;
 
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -52,6 +52,13 @@
             </label>
           </div>
 
+          <div class="col-md-2">
+            <select id="type-selection" selected="type" class="form-control">
+              <option value="generated">Data generator</option>
+              <option value="live">Live data</option>
```
CBarrett2: change to bitcoin data

src/assets/js/data/feed/coinbase/historicFeed.js
```diff
@@ -0,0 +1,29 @@
...
+                    newData = newData.reverse();
+                    callback(null, newData);
+                } else { callback(err, null); }
+            });
+        }
+
+        d3.rebind(historicFeed, feed, 'granularity', 'start', 'end', 'product');
+
+        historicFeed.invalidateCallback = function() {
+            n++;
```
CBarrett2: could invalidate future callbacks

CBarrett2: Actually, I'm not sure if there is a problem with this? Since any 'future callbacks' would use the new n to generate their id, thus never being generated invalid.

PR #133, opened 2015-08-18
Bug(?) -Deleted reinitialisation in navChart

djmiley:
Previous variable is overwritten, fixed this. #112


CBarrett2:
Good catch

PR #134, opened 2015-08-18
Pad the xDomain of the primary chart so full candles can be seen at either extreme

CBarrett2:
The primary chart now shows slightly more than the 'viewDomain', so the full candles can fit on the screen. The bufferPeriod can be set outside the chart and also turned since it's only needed for certain series types.

### DIFF
--- a/src/assets/js/chart/macdChart.js
+++ b/src/assets/js/chart/macdChart.js
@@ -11,7 +11,7 @@
         function macdChart(selection) {
             var data = selection.datum().data;
             var domain = selection.datum().domain;
-            var currentBufferPeriod = selection.datum().displayBuffer ? selection.datum().dataPeriod : 0;
+            var currentBufferPeriod = selection.datum().displayBuffer ? selection.datum().period : 0;
             var paddedDomain = sc.util.paddedExtent(domain, currentBufferPeriod);
 
             macdAlgorithm(data);
--- a/src/assets/js/chart/macdChart.js
+++ b/src/assets/js/chart/macdChart.js
@@ -10,7 +10,9 @@
 
         function macdChart(selection) {
             var data = selection.datum().data;
-            var viewDomain = selection.datum().viewDomain;
+            var domain = selection.datum().domain;
+            var currentBufferPeriod = selection.datum().displayBuffer ? selection.datum().dataPeriod : 0;
+            var paddedDomain = sc.util.paddedExtent(domain, currentBufferPeriod);
 
             macdAlgorithm(data);
 
@@ -19,21 +21,14 @@
             });
 
             macd.xScale()
-                .domain(viewDomain)
+                .domain(paddedDomain)
                 .range([0, parseInt(selection.style('width'), 10)]);
             macd.yScale()
                 .domain([-maxYExtent, maxYExtent])
                 .range([parseInt(selection.style('height'), 10), 0]);
 
+            sc.util.applyZoom(selection, data, dispatch, domain);
 
-            var zoom = d3.behavior.zoom();
-            zoom.x(macd.xScale())
-                .on('zoom', function() {
-                    sc.util.zoomControl(zoom, selection, data, macd.xScale());
-                    dispatch.viewChange(macd.xScale().domain());
-                });
-
-            selection.call(zoom);
             selection.datum(data)
                 .call(macd);
         }
--- a/src/assets/js/chart/macdChart.js
+++ b/src/assets/js/chart/macdChart.js
@@ -27,7 +27,7 @@
                 .domain([-maxYExtent, maxYExtent])
                 .range([parseInt(selection.style('height'), 10), 0]);
 
-            sc.util.applyZoom(selection, data, dispatch, domain);
+            selection.call(sc.util.boundedZoom, dispatch);
 
             selection.datum(data)
                 .call(macd);
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -43,8 +43,8 @@
             .yTicks(5)
             .xTicks(0);
 
-        var currentSeries = sc.series.candlestick();
-        var currentIndicator;
+        var currentSeries = sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick(), true);
+        var currentIndicator = sc.menu.option('None', 'no-indicator', null);
 
         // Create and apply the Moving Average
         var movingAverage = fc.indicator.algorithm.movingAverage();
@@ -63,21 +63,23 @@
                 }
                 return series;
             })
-            .series([gridlines, currentSeries, closeLine]);
+            .series([gridlines, currentSeries.option, closeLine]);
 
         function updateMultiSeries() {
-            if (currentIndicator) {
-                multi.series([gridlines, currentSeries, closeLine, currentIndicator]);
+            if (currentIndicator.option) {
+                multi.series([gridlines, currentSeries.option, closeLine, currentIndicator.option]);
             } else {
-                multi.series([gridlines, currentSeries, closeLine]);
+                multi.series([gridlines, currentSeries.option, closeLine]);
             }
         }
 
         function primaryChart(selection) {
             var data = selection.datum().data;
-            var viewDomain = selection.datum().viewDomain;
+            var domain = selection.datum().domain;
 
-            timeSeries.xDomain(viewDomain);
+            var currentBufferPeriod = currentSeries.buffer ? selection.datum().dataPeriod : 0;
+            var paddedDomain = sc.util.paddedExtent(domain, currentBufferPeriod);
+            timeSeries.xDomain(paddedDomain);
 
             // Scale y axis
             var yExtent = fc.util.extent(sc.util.filterDataInDateRange(data, timeSeries.xDomain()), ['low', 'high']);
@@ -121,15 +123,7 @@
             timeSeries.plotArea(multi);
             selection.call(timeSeries);
 
-            // Behaves oddly if not reinitialized every render
-            var zoom = d3.behavior.zoom();
-            zoom.x(timeSeries.xScale())
-                .on('zoom', function() {
-                    sc.util.zoomControl(zoom, selection.select('.plot-area'), data, timeSeries.xScale());
-                    dispatch.viewChange(timeSeries.xDomain());
-                });
-
-            selection.call(zoom);
+            sc.util.applyZoom(selection, data, dispatch, domain);
         }
 
         d3.rebind(primaryChart, dispatch, 'on');
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -77,7 +77,7 @@
             var data = selection.datum().data;
             var domain = selection.datum().domain;
 
-            var currentBufferPeriod = currentSeries.buffer ? selection.datum().dataPeriod : 0;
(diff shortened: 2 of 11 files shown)


src/assets/js/main.js
```diff
@@ -40,21 +41,27 @@
         switch (seriesTypeString) {
             case 'ohlc':
                 currentSeries = ohlc;
+                primaryChart.toggleBuffer(true);
```
CBarrett2: If #125 is pulled, this can be checked for inside primaryChart, which might be better

CBarrett2:
Resolves #89

CBarrett2:
image
This introduces a lesser issue as seen above with the bollinger bands, when viewed with an OHLC or other padded series, where they stop short of the end. I don't know what a sensible solution to this could be - perhaps it's a matter of removing the fill/colouring between the lines, so it looks acceptable that they stop short?

CBarrett2:
I think it might still be worth pulling in, as it gets rid of a worse bug and introduces what is I suppose expected behaviour?

src/assets/js/chart/primaryChart.js
```diff
@@ -61,10 +61,19 @@
                 return series;
             });
 
+        var viewScale = fc.scale.dateTime();
+        var bufferPeriod = 60 * 60 * 24;
+        var bufferOn = true;
```
djmiley: I'd probably move these further up in the function, they're more important than the closing line annotation for example

src/assets/js/chart/primaryChart.js
```diff
@@ -61,10 +61,19 @@
...
         function primaryChart(selection) {
             var data = selection.datum().data;
             var viewDomain = selection.datum().viewDomain;
-            timeSeries.xDomain(viewDomain);
+
+            var currBufferPeriod = bufferOn ? bufferPeriod : 0;
```
djmiley: No need to shorten current

src/assets/js/chart/primaryChart.js
```diff
@@ -61,10 +61,19 @@
...
+            var paddedDomain = sc.util.paddedExtent(viewDomain, currBufferPeriod);
+            timeSeries.xDomain(paddedDomain);
+
+            viewScale.domain(viewDomain)
+                .range([0, parseInt(selection.attr('width'), 10)]);
```
djmiley: Yes, `linearTimeSeries` does this internally which is why it looked a little odd to me.

src/assets/js/chart/primaryChart.js
```diff
@@ -107,6 +116,15 @@
...
         };
+        primaryChart.toggleBuffer = function(toggle) {
+            bufferOn = toggle;
+            return primaryChart;
+        };
+
+        primaryChart.bufferPeriod = function(period) {
+            bufferPeriod = period;
+            return primaryChart;
+        };
```
djmiley: Maybe add return for `!arguments.length`?

src/assets/js/main.js
```diff
@@ -32,7 +32,8 @@
 
     sc.util.calculateDimensions(container);
 
-    var primaryChart = sc.chart.primaryChart();
+    var primaryChart = sc.chart.primaryChart()
+        .bufferPeriod(60 * 60 * 24);
```
djmiley: Is this is default value, you don't need to set it here.

src/assets/js/chart/primaryChart.js
```diff
@@ -107,6 +116,15 @@
             }
             return primaryChart;
         };
+        primaryChart.toggleBuffer = function(toggle) {
+            bufferOn = toggle;
+            return primaryChart;
+        };
+
```
djmiley: Not really a `toggleBuffer`, more a `displayBuffer`?

src/assets/js/chart/primaryChart.js
```diff
@@ -29,6 +29,10 @@
     sc.chart.primaryChart = function() {
         var dispatch = d3.dispatch('viewChange');
 
+        var viewScale = fc.scale.dateTime();
+        var bufferPeriod = 60 * 60 * 24;
```
jleft: Should this be adjusted for the period of the data which is being displayed?

E.g. I expect with the smaller period currently used with the bitcoin data source, there would be a disproportionately large amount of padding either side of the series?

CBarrett2: Yes, it should be :) Must have forgotten to update it after the live data got pulled in.

CBarrett2:
Squashed the commits and rebased

src/assets/js/chart/primaryChart.js
```diff
@@ -115,9 +125,21 @@
...
+            currentSeries = seriesType.series;
             updateMultiSeries();
+            if (seriesType.valueString === 'line' || seriesType.valueString === 'area') {
+                displayBuffer = false;
+            } else { displayBuffer = true; }
+
+            return primaryChart;
+        };
+
+        primaryChart.bufferPeriod = function(period) {
```
jleft: More general question - why have a setter for `bufferPeriod`, rather than have it as a property on the data model (which is bound to a selection passed to this component)? Just curious!

CBarrett2: Only because it's specific to timeSeries, I suppose! Either could work

CBarrett2:
Just noticed a bug with this, where they secondary chart will move relative to the primary chart, especially at high zoom levels. This is probably just a side-effect of the way I've done this, will look into fixes

src/assets/js/chart/macdChart.js
```diff
@@ -11,6 +11,8 @@
         function macdChart(selection) {
             var data = selection.datum().data;
             var viewDomain = selection.datum().viewDomain;
+            var currentBufferPeriod = selection.datum().displayBuffer ? selection.datum().dataPeriod : 0;
```
jleft: If `displayBuffer` is only used to decide if we should use `dataPeriod` or `0`, could we just have a property on the model, maybe `xScaleDomainPeriodPadding` (maybe something a bit more concise...), where we pass in the period to pad (it could be set to `0` for the current instances where `displayBuffer` is `false`)?

Then we'd only need something like this:

```js
var currentBufferPeriod = selection.datum().xScaleDomainPeriodPadding;
```

src/assets/js/chart/macdChart.js
```diff
@@ -19,21 +21,14 @@
...
 
             macd.xScale()
-                .domain(viewDomain)
+                .domain(paddedDomain)
                 .range([0, parseInt(selection.style('width'), 10)]);
             macd.yScale()
                 .domain([-maxYExtent, maxYExtent])
                 .range([parseInt(selection.style('height'), 10), 0]);
 
+            sc.util.applyZoom(selection, data, dispatch, viewDomain, macd.xScale().range());
```
jleft: Is `viewDomain` an appropriate name? It seems like `paddedDomain` is the domain which is _viewed_?

src/assets/js/main.js
```diff
@@ -28,6 +30,9 @@
     var mainMenu = sc.menu.main()
         .on('primaryChartSeriesChange', function(series) {
             primaryChart.changeSeries(series.option);
+            if (series.valueString === 'line' || series.valueString === 'area') {
+                dataModel.displayBuffer = false;
+            } else { dataModel.displayBuffer = true; }
```
jleft: `dataModel.displayBuffer = true;` on a new line

src/assets/js/main.js
```diff
@@ -28,6 +30,9 @@
     var mainMenu = sc.menu.main()
         .on('primaryChartSeriesChange', function(series) {
             primaryChart.changeSeries(series.option);
+            if (series.valueString === 'line' || series.valueString === 'area') {
```
jleft: What if we add more series types? Could this become difficult to maintain?

src/assets/js/util/paddedExtent.js
```diff
@@ -0,0 +1,13 @@
+(function(d3, fc) {
+    'use strict';
+    sc.util.paddedExtent = function(extent, period) {
```
jleft: Looks like a nice one to unit test.

src/assets/js/chart/macdChart.js
CBarrett2: This could work, but I think we'll always want to use the dataPeriod as the padding value and that will also be used when we add the period change button, so maybe it's better to have dataPeriod and bufferDisplay rather than dataPeriod and xScaleDomainPeriodPadding?

src/assets/js/main.js
CBarrett2: the alternative I guess would be having a property of the seriesType object as 'padding = true/false', which might be nice?

src/assets/js/chart/macdChart.js
CBarrett2: I think you're right but struggling to think what to call it. Maybe navigationDomain? Not sure if that's misleading, as if it's specifically to do with the navChart

src/assets/js/chart/primaryChart.js
```diff
@@ -68,7 +68,9 @@
             var data = selection.datum().data;
             var viewDomain = selection.datum().viewDomain;
```
CBarrett2: viewDomain is still potentially misleading, but unsure what a better name would be!

src/assets/js/chart/macdChart.js
```diff
@@ -19,21 +21,14 @@
...
 
             macd.xScale()
-                .domain(viewDomain)
+                .domain(paddedDomain)
                 .range([0, parseInt(selection.style('width'), 10)]);
             macd.yScale()
                 .domain([-maxYExtent, maxYExtent])
                 .range([parseInt(selection.style('height'), 10), 0]);
 
+            sc.util.applyZoom(selection, data, dispatch, domain);
```
djmiley: You need to do this for `xAxisChart` ?

djmiley: `data = selection.datum().data` , `domain = selection.datum().domain`

So don't need to pass them in?

djmiley: Also, `selection.call...` would look a lot nicer

CBarrett2: xAxis shouldn't need a zoom component at all imo, so I'll remove that. Good catch though, and good suggestions

CBarrett2:
There's a slight issue in this where the period is changed before the get request comes back with the new data, but might be better fixed in the loading screen issue, rather than checking button states in the dataLoaded event handler?

djmiley:
I like what is going on here.

src/assets/js/menu/primaryChart/series.js
```diff
@@ -5,12 +5,12 @@
...
-        var ohlc = sc.menu.option('OHLC', 'ohlc', fc.series.ohlc());
-        var line = sc.menu.option('Line', 'line', fc.series.line());
+        var candlestick = sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick(), true);
+        var ohlc = sc.menu.option('OHLC', 'ohlc', fc.series.ohlc(), true);
+        var line = sc.menu.option('Line', 'line', fc.series.line(), false);
         line.option.isLine = true;
-        var point = sc.menu.option('Point', 'point', fc.series.point());
-        var area = sc.menu.option('Area', 'area', fc.series.area());
+        var point = sc.menu.option('Point', 'point', fc.series.point(), true);
+        var area = sc.menu.option('Area', 'area', fc.series.area(), false);
```
djmiley: There should be something more inherent we can do this on? As in, I feel there ought to be.

CBarrett2:
Superseded by #222

PR #140, opened 2015-08-19
Optimised candlestick series from d3fc

CBarrett2:
Before every candlestick was in its own group, so I removed that to optimize performance, following the example from http://blog.scottlogic.com/2015/08/06/an-adventure-in-svg-filter-land.html. I believe the candlestick series works faster on mobile now.


CBarrett2:
Begins to work with #76

src/assets/js/series/optimisedCandlestick.js
```diff
@@ -0,0 +1,59 @@
+(function(d3, fc, sc) {
```
jleft: It would be good to credit the original author and link to the blog post which this is based upon - can do this in a comment.

src/assets/js/series/optimisedCandlestick.js
```diff
@@ -0,0 +1,59 @@
...
+    'use strict';
+    sc.series.optimisedCandlestick = function() {
```
jleft: `optimisedCandlestick` -> `candlestick` (file name too)

src/index.html
```diff
@@ -72,6 +72,10 @@
     <script src="assets/js/util/callbackInvalidator.js"></script>
     <script src="assets/js/data/feed/coinbase/webSocket.js"></script>
     <script src="assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js"></script>
+    <script src="assets/js/series/optimisedCandlestick.js"></script>
+    <script src="assets/js/chart/primaryChart.js"></script>
```
jleft: Already included - same for `rsiChart` and `navChart`

src/assets/styles/candlestick.less
```diff
@@ -0,0 +1,8 @@
+path.down {
```
jleft: Can we move this to `plotArea.less`? (When integrated which the primary chart, it would make sense to have a parent container with a `candlestick` class and use this class to apply the styles only to the candlestick series.)

jleft:
Looks good - any reason not to use this series in place of the current d3fc candlestick series?

CBarrett2:
The downside is that you can no longer interact with each individual candlestick using decorate, I think, but we don't seem to need that at the minute.

jleft:
> The downside is that you can no longer interact with each individual candlestick using decorate, I think, but we don't seem to need that at the minute.

If decorate isn't being used, then I think it would be good to integrate this with the chart.

src/assets/js/main.js
```diff
@@ -45,7 +45,7 @@
         this.series = series;
     };
 
-    var candlestick = new SeriesType('Candlestick', 'candlestick', fc.series.candlestick());
+    var candlestick = new SeriesType('Candlestick', 'candlestick', sc.series.candlestick());
```
jleft: Will also need to update `currentSeries` in primary chart?

src/assets/styles/plotArea.less
```diff
@@ -20,4 +20,13 @@
     }
   }
 
+}
+
```
jleft: Move `path.up` and `path.down`  into `.plotArea` above, so these styles will only be applied within the plot area

src/assets/styles/plotArea.less
```diff
@@ -20,4 +20,13 @@
...
+path.down {
+  stroke: #000;
```
jleft: I think it's worth factoring these colours out into variables.

src/assets/styles/plotArea.less
CBarrett2: Ah, okay, I see what you mean. I am a CSS n00b

src/assets/js/main.js
```diff
@@ -45,13 +45,15 @@
...
     var ohlc = new SeriesType('OHLC', 'ohlc', fc.series.ohlc());
     var line = new SeriesType('Line', 'line', fc.series.line());
     line.series.isLine = true;
     var point = new SeriesType('Point', 'point', fc.series.point());
     var area = new SeriesType('Area', 'area', fc.series.area());
 
+    primaryChart.changeSeries(candlestick.series);
```
jleft: If possible, I'd avoid this, and just set the candlestick as the default in primary chart.

jleft:
One final comment and when the merge conflicts are resolved, I think this is good to go - thanks @CBarrett2.

jleft:
:+1:

PR #141, opened 2015-08-19
Ability to toggle secondary chart

### LINKED ISSUE #98: Hide secondary chart by default
By default, the secondary chart should be hidden. The user should have the option to turn RSI on/off, which will show/hide the secondary chart.

djmiley:
Resolves #98

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -4,8 +4,8 @@
     // Set SVGs & column padding
     var container = d3.select('#app-container');
 
-    var svgMain = container.select('svg.primary');
-    var svgRSI = container.select('svg.rsi');
+    var svgPrimary = container.select('svg.primary');
+    var svgSecondary = container.select('svg.secondary');
     var svgNav = container.select('svg.nav');
 
     var dataModel = {
@@ -16,7 +16,7 @@
     sc.util.calculateDimensions(container);
 
     var primaryChart = sc.chart.primaryChart();
-    var rsiChart = sc.chart.rsiChart();
+    var secondaryChart = null;
     var navChart = sc.chart.navChart();
 
     var seriesOptions = sc.menu.optionGenerator()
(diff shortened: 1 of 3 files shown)


djmiley:
oops copuple of things to do here.

djmiley:
Blocked by #125

src/assets/js/util/calculateDimensions.js
```diff
@@ -13,18 +13,16 @@
...
-            .attr('height', mainHeightRatio * targetHeight / totalHeightRatio);
-        container.select('svg.rsi').attr('width', targetWidth)
-            .attr('height', rsiHeightRatio * targetHeight / totalHeightRatio);
-        container.select('svg.nav').attr('width', targetWidth);
+        container.select('svg.primary').attr('width', useableScreenWidth)
+            .attr('height', primaryHeightRatio * useableScreenHeight / totalHeightRatio);
+        container.select('svg.secondary').attr('width', useableScreenWidth)
+            .attr('height', secondaryHeightRatio * useableScreenHeight / totalHeightRatio);
+        container.select('svg.nav').attr('width', useableScreenWidth);
     };
```
djmiley: Just renaming of variables, except for truthy on `secondaryChartShown`

src/assets/js/main.js
```diff
@@ -92,18 +112,22 @@
...
 
         svgNav.datum(dataModel)
             .call(navChart);
     }
 
     function resize() {
-        sc.util.calculateDimensions(container);
+        sc.util.calculateDimensions(container, secondaryChart);
         render();
     }
```
djmiley: Not certain about my treatment of initialisation and `null` values here.

janakerman: I'm not actually sure what the best practises are with initialising things with no value. I normally simply do:

    var myVar;

I think it's fine to set it to null too. I suppose a bit of googling would answer that question, but I'm happy with what's here for the PR.

What do you mean about you 'treatment of initialisation'?

PR #143, opened 2015-08-19
Integrate coinbase alternative - pulled out invalidateCallback as util

CBarrett2:
Sorry for the extra PR, but wasn't sure if this was what you meant @tlsim :)


src/assets/js/data/feed/coinbase/historicFeed.js
```diff
@@ -0,0 +1,23 @@
+(function(sc) {
+    'use strict';
+
+    sc.data.feed.coinbase.invalidator = function() {
```
CBarrett2: Unsure about the naming of this

src/assets/js/main.js
```diff
@@ -86,9 +86,43 @@
...
+    }
+
+    function historicCallback() {
+        return callbackGenerator(onHistoricDataLoaded);
+    }
+
+    d3.select('#type-selection')
+        .on('change', function() {
+            var type = d3.select(this).property('value');
+            if (type === 'bitcoin') {
```
janakerman: Tidy up as discussed.

src/assets/js/data/feed/coinbase/historicFeed.js
janakerman: callbackInvalidator? Doesn't seem critical either way

src/assets/js/main.js
CBarrett2: done :)

CBarrett2:
Resolves #83

janakerman:
Looks like you've forgotten to delete `historicFeed.js`?

janakerman:
I think a real improvement to this would be to add promises. Using promises would tidy up a lot of the async behaviour here and **may** even give you the ability to cancel callbacks. Hoever, as we discussed, it's fine to leave this for the future if we find ourselves adding more async behaviour.

CBarrett2:
Okay :) I'll have a look into them

Issue #136, opened 2015-08-18
Navbar reset button changes default viewDomain to different thing on mobile and desktop


CBarrett2:
Also IMO the default is too small on desktop now the navbar has shrunk. I believe this issue is due to the use of the golden ratio to size the rectangle, rather than just using a multiple of the current period (I would say to use a percentage of the loaded data, but it's possible we will add functionality to load variable amounts of data in future)

CBarrett2:
Resolves #144

jleft:
Resolved in #144

PR #144, opened 2015-08-19
Reset button changes view domain to the last 50 datums (or fewer if there are fewer than 50 data points)

CBarrett2:


djmiley:
I'm happy to let the golden ratio go, for reasons of consistency.

CBarrett2:
The end of an era :(

djmiley:
It's worth thinking whether a number of data points related to period is a better measure to show?

So is a 1/4 or 50 data points better for now?

djmiley:
But `navAspect` is beginning to look more and more like an ugly variable.

CBarrett2:
Yes, number of data points would be better actually

djmiley:
I agree. :)

CBarrett2:
Resolves #136

djmiley:
Looks good.

Issue #135, opened 2015-08-18
There should be more space between the highest and lowest points of the data and the top/bottom of the screen in the main chart. 


CBarrett2:
Sort of a 10% is padding of the yExtent in timeSeries, like there was before

jleft:
Resolved in #145

PR #145, opened 2015-08-20
More space at top and bottom of data

CBarrett2:


CBarrett2:
Resolves #135

djmiley:
Looks good, as long as it definitely works- which I think it does.

CBarrett2:
I can't reproduce the bug it was removed for in the first place :) Think it was likely something to do with setting y0Value of the area series rather than this

PR #147, opened 2015-08-20
Bug fix for moving average/ line

djmiley:


janakerman:
As discussed, this isn't a nice fix but it's just reverting to the old behaviour. @djmiley is going to look into this keying function and see if he can find an improved approach.

PR #148, opened 2015-08-20
Altered bug fix for line from boolean (See #154)

djmiley:
Low-res solution for isLine boolean. Compare to #154


src/assets/js/chart/primaryChart.js
```diff
@@ -105,9 +102,9 @@
 
         function updateMultiSeries() {
             if (currentIndicator) {
-                multi.series([gridlines, currentSeries, closeAxisAnnotation, currentIndicator]);
+                multi.series([gridlines, closeAxisAnnotation, currentSeries, currentIndicator]);
             } else {
-                multi.series([gridlines, currentSeries, closeAxisAnnotation]);
+                multi.series([gridlines, closeAxisAnnotation, currentSeries]);
             }
```
djmiley: A more sensible reordering imo.

src/assets/js/main.js
```diff
@@ -93,6 +92,8 @@
     container.select('#reset-button').on('click', resetToLive);
 
     function render() {
+        svgMain.selectAll('*').remove();
```
CBarrett2: It seems like it would be really inefficient to remove then re-add every single element (200+ candlesticks and more) every time we want to re-draw

CBarrett2:
Could we close this?

djmiley:
Go on then.

Issue #73, opened 2015-08-11
Configure indicator parameters


jleft:

djmiley:
Need to be able to configure the indicator visible on the Primary Chart. For example, Moving Average or Bollinger Bands. Configuration possibilities include the period length of the indicator #150 or the price used in the calculation of the indicator #149

djmiley:
Some work on changing the period length of the indicators
https://github.com/djmiley/d3fc-showcase/tree/feature/indicatorperiod

PR #152, opened 2015-08-20
added indicator period length/window size

### LINKED ISSUE #150: Primary Chart Indicator period length
Change the period length of the primary chart indicators

RuthC-C:
Added ability to change the window size of both bollinger bands and moving average.
Part of #73 
resolves #150

### DIFF
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -148,6 +148,12 @@
             return primaryChart;
         };
 
+        primaryChart.changeWindowSize = function(windowSize) {
+            movingAverage.windowSize(windowSize);
+            bollingerAlgorithm.windowSize(windowSize);
+            return primaryChart;
+        };
+
         return primaryChart;
     };
 })(d3, fc, sc);
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -118,6 +118,10 @@
         .on('periodChange', function(period) {
             dataModel.period = period;
             dataInterface(dataModel.period);
+        })
+        .on('windowSizeChanged', function(windowSize) {
+            primaryChart.changeWindowSize(windowSize);
+            render();
         });
 
     container.select('.menu')
@@ -130,4 +134,5 @@
     dataInterface.generateData();
     sc.util.calculateDimensions(container, secondaryChart);
     resetToLive();
+    resize();
 })(d3, fc, sc);
--- a/src/assets/js/menu/indicatorWindowSize.js
+++ b/src/assets/js/menu/indicatorWindowSize.js
@@ -1,0 +1,65 @@
+(function(d3, fc) {
+    'use strict';
+
+    sc.menu.indicatorWindowSize = function() {
+
+        var dispatch = d3.dispatch('windowSizeChanged');
+        var windowValue = 10;
+        var indicators;
+
+        function indicatorPeriodLength(sel) {
+            var label = sel.selectAll('label')
+                .data(sel.datum())
+                .enter()
+                .append('label')
+                .text(function(d, i) { return d.displayString + ' Period Length'; });
+            label.append('input')
+                .attr({
+                    type: 'range',
+                    min: 1,
+                    max: 100,
+                    value: function(d, i) { return d.valueString; },
+                    id: 'window-slider'
+                })
+                .property('value', windowValue);
+            label.append('input')
+                .attr({
+                    type: 'number',
+                    min: 1,
+                    max: 100,
+                    value: function(d, i) { return d.valueString; },
+                    id: 'window-number'
+                })
+                .property('value', windowValue);
+        }
+
+        function indicatorWindowSize(selection) {
+            selection.datum(indicators)
+                .call(indicatorPeriodLength);
+
+            selection.selectAll('input')
+                .on('input', function() {
+                    windowValue = +this.value;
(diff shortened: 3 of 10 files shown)


src/assets/js/main.js
```diff
@@ -92,6 +92,19 @@
 
     container.select('#reset-button').on('click', resetToLive);
 
+    d3.select('#windowSize').on('input', function() {
```
jleft: Is it worth componentising this? The component could take in a selection and add the slider. This event listener, for `input`, could be be added in the component and a more abstract event, maybe `windowSizeChanged`, broadcast to main?

src/assets/js/chart/primaryChart.js
```diff
@@ -123,6 +123,12 @@
             return primaryChart;
         };
 
+        primaryChart.changeWindowSize = function(value) {
```
jleft: `value` -> `windowSize`?

src/assets/js/main.js
```diff
@@ -92,6 +92,19 @@
...
+        update(parseInt(this.value, 10));
+    });
+
+    function update(windowSize) {
```
jleft: `update` -> `updateWindowSize`? or `updateIndicatorWindowSize`?

src/assets/js/chart/primaryChart.js
```diff
@@ -123,6 +123,12 @@
             return primaryChart;
         };
 
+        primaryChart.changeWindowSize = function(windowSize) {
```
jleft: If we take this approach, I think I'd lean towards storing the window size as a property of the component, then this could be a more conventional getter/setter, and we wouldn't have to worry about calling `whateverIndicator.windowSize` here. This would allow the window size to be inspected (using the getter) or set (using the setter).

An alternative would be to set the window size as a property on the data model which we pass to the primary chart.

I'd probably have a slight preference to the latter in this case, but I wonder if anyone else has any thoughts on this topic?

src/assets/js/menu/indicatorWindowSize.js
```diff
@@ -0,0 +1,27 @@
...
+        function indicatorWindowSize() {
+            d3.select('#windowSize').on('input', function() {
+                var windowValue = +this.value;
+                dispatch.windowSizeChanged(windowValue);
+                updateIndicatorWindowSize(windowValue);
+            });
+        }
+
+        d3.rebind(indicatorWindowSize, dispatch, 'on');
+        updateIndicatorWindowSize(10);
```
jleft: Do we need to do this?

src/index.html
```diff
@@ -34,6 +34,10 @@
           </div>
           <div id="secondary-chart-buttons" class="btn-group" data-toggle="buttons">
           </div>
+           <div id="window-slider" class="btn-group" for="windowSize" >
```
jleft: This feels like something which could be pulled out into a component, but I don't think it's worth it at the moment.

There's a mix of `camelCase` and `hyphen-separated` variables. Let's be consistent and use `hypen-separated`.

src/assets/js/chart/primaryChart.js
CBarrett2: is attaching it to the dataModel is pretty much the same as making it a global variable though, as maybe the other charts could then change it?

src/assets/js/menu/indicatorWindowSize.js
```diff
@@ -0,0 +1,27 @@
...
+
+        var dispatch = d3.dispatch('windowSizeChanged');
+
+        var updateIndicatorWindowSize = function(windowSize) {
+            d3.select('#windowSize-value').text(windowSize);
+            d3.select('#windowSize').property('value', windowSize);
+        };
+
+        function indicatorWindowSize() {
+            d3.select('#windowSize').on('input', function() {
```
jleft: This is tightly-coupled to a specific element in the HTML - i.e. the element with ID `windowSize`. Perhaps this will be problematic when adding more configuration for the indicators?

Typically, a component would accept a selection to allow one or more elements to be manipulated by the component - if this isn't an issue now, I expect we'll need to revisit when support for multiple simultaneous indicators is added?

src/assets/js/chart/primaryChart.js
jleft: > is attaching it to the dataModel pretty much the same as making it a global variable though, as maybe the other charts could then change it? Unsure if they all have copies of the same data or - in a more typically JS way - have references to the same object

I don't think so. At the moment, very little is stored is stored on the model, so they all happen to share the same model.

Each chart/navigator/menu etc. could have its own data model. For example, the navigator doesn't need to know about window size, so window size wouldn't be on its model. The models would have some shared properties though, e.g. data.

> Perhaps it's worth considering taking the indicator algorithms out into a seperate class/main

I'd encourage not taking it out into main (due to the maintenance issues we had before); a separate class could work though.

> since basically what they do is attach new properties onto the data - which isn't really the responsibility of primaryChart at all IMO. Then they could just be called before the data is bound onto the selections

This isn't actually done in primary chart though - this is done within the indicator algorithm components, for example `fc.indicator.algorithm.movingAverage()`. I think it's reasonable if primary chart to have some control over its indicators, otherwise what would primary chart be responsible for?

Perhaps primary chart could be composed of a component which manages the indicators? This could then be reused for the secondary chart(s)?

djmiley:
Superseded by #258

PR #154, opened 2015-08-20
Component approach to solving line issue

djmiley:
Created `sc.indicator.renderer.movingAverage` Compare to #148


djmiley:
I prefer this one.

src/assets/js/indicator/renderer/movingAverage.js
```diff
@@ -0,0 +1,64 @@
...
+    'use strict';
+
+    sc.indicator.renderer.movingAverage = function() {
+
+        var xScale = d3.time.scale();
+        var yScale = d3.scale.linear();
+        var yValue = function(d, i) { return d.close; };
+        var xValue = function(d, i) { return d.date; };
+
+        var averageLine = fc.series.line()
```
RuthC-C: Call <code>line</code> instead of <code>averageLine</line>?
Nothing important just something that bugs me personally (<code>averageLine</code> sounds a little like there is more than one line?)

RuthC-C:
Looks great - tested and working. :+1:

src/assets/js/indicator/renderer/movingAverage.js
```diff
@@ -0,0 +1,64 @@
...
+            .yValue(function(d, i) {
+                return d.movingAverage;
+            });
+
+        function movingAverage(selection) {
+
+            var multi = fc.series.multi()
```
CBarrett2: Can we just render the line series directly, rather than through multi? Since it's just a single series

djmiley: I should have a look, thanks.

jleft:
If I understand the problem correctly, we were keying the data based upon the series; ultimately these keys are strings. If we have more than one series of the same type, they will have the same key - so only one would be rendered.

We also can't key on index, as this would involve reusing containers which could result in the wrong output.

If we took this approach, of wrapping a component in another component, we'd never be able to have more than one series of the same type working with multi. We'd also have to create a new component each time this scenario arises - the new components would essentially be doing an identical job too, which seems somewhat redundant?

I actually quite like the basis of this idea - wrapping the rendering component in another component, and keying on its unique property. Something simple like this _might_ work, where we key on the ID, although I think the render function from the component if what multi expects, not an object like this:

```js
{ renderer: sc.some.renderer(), id: 0 }
```

One idea would be to add an ID property to the series and assign it a GUID and key on it:

```js
var series = sc.some.renderer();
series.id = sc.util.guid(); // would assign a new GUID
```

Any thoughts?

jleft:
Given the above comment, can this be closed?

djmiley:
Yes, makes sense.

PR #155, opened 2015-08-20
Integrate live data with historic

### LINKED ISSUE #53: Coinbase Data aggregation

CBarrett2:
Resolves #53

### DIFF
--- a/src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
+++ b/src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
@@ -7,8 +7,8 @@
         var period = 60 * 60 * 24;
         var liveFeed = sc.data.feed.coinbase.webSocket();
 
-        function ohlc(cb) {
-            var basket = null;
+        function ohlc(cb, initialBasket) {
+            var basket = initialBasket || null;
             liveFeed(function(err, datum) {
                 if (datum) {
                     basket = updateBasket(basket, datum);
@@ -35,7 +35,9 @@
             var startTime = basket.date.getTime();
             var msPeriod = period * 1000;
             if (latestTime > startTime + msPeriod) {
-                basket = createNewBasket(datum, new Date(startTime + msPeriod));
+                var timeIntoCurrentPeriod = (latestTime - startTime) % msPeriod;
+                var newTime = latestTime - timeIntoCurrentPeriod;
+                basket = createNewBasket(datum, new Date(newTime));
             } else {
                 // Update current basket
                 basket.high = Math.max(basket.high, datum.price);
--- a/src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
+++ b/src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
@@ -8,7 +8,7 @@
         var liveFeed = sc.data.feed.coinbase.webSocket();
 
         function ohlcWebSocketAdaptor(cb, initialBasket) {
-            var basket = initialBasket || null;
+            var basket = initialBasket;
             liveFeed(function(err, datum) {
                 if (datum) {
                     basket = updateBasket(basket, datum);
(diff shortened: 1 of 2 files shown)


src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
```diff
@@ -7,8 +7,8 @@
         var period = 60 * 60 * 24;
         var liveFeed = sc.data.feed.coinbase.webSocket();
 
-        function ohlc(cb) {
-            var basket = null;
+        function ohlc(cb, initialBasket) {
```
janakerman: Could you rename this to ohlcWebSocketAdaptor? I know you didn't change it in this commit, but it'll be a quick change and it doesn't feel worth raising as an issue to fix later.

CBarrett2: Yeah, not a problem!

src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
```diff
@@ -7,8 +7,8 @@
...
+            var basket = initialBasket || null;
```
janakerman: `null == undefined` is `true`, I don't think there's any need to differentiate between the two in this scenario. I'd just put
    
    var basket = intialBasket;

src/assets/js/main.js
```diff
@@ -119,6 +119,32 @@
...
+        if (data[data.length - 1].date.getTime() !== basket.date.getTime()) {
+            data.push(basket);
+        } else {
+            data[data.length - 1] = basket;
+        }
+        render();
+    }
+
+    function liveCallback(event, latestBasket) {
+        if (!event && latestBasket) {
```
janakerman: Is the lack of an event parameter being used to specify that there was an update event? Might be nicer to add an event type, unless these events are defined by the socket protocol or something, then I'd rename this parameter `socketEvent` or something the like. Just to make it a little more clear that it's an event relating to sockets, and not just to your adaptor.

src/assets/js/main.js
```diff
@@ -119,6 +119,32 @@
...
+            newBasketReceived(latestBasket);
+        } else if (event.type === 'open') {
```
janakerman: I'd just delete the empty branches of this if statement.

src/assets/js/data/feed/coinbase/ohlcWebSocketAdaptor.js
```diff
@@ -7,8 +7,8 @@
         var period = 60 * 60 * 24;
         var liveFeed = sc.data.feed.coinbase.webSocket();
 
-        function ohlc(cb) {
-            var basket = null;
+        function ohlcWebSocketAdaptor(cb, initialBasket) {
```
CBarrett2: Should I pass in period here, rather than hold it as a state? re: our discussion about unit tests?

src/assets/js/main.js
CBarrett2: I'd have to then check that event !== 'open && !(event === 'close' && event.code === 1000) in the last check then? Not sure if there's a better way

src/assets/js/main.js
```diff
@@ -119,6 +119,34 @@
...
+        }
+        render();
+    }
+
+    function liveCallback(socketEvent, latestBasket) {
+        if (latestBasket) {
+            newBasketReceived(latestBasket);
+        }
+
+        if (socketEvent) {
```
janakerman: I wouldn't nest the IF..  (socketEvent && !sucessfulOpen && !successfulClose) just because I always find deep logic nesting hard to follow (and these things always tend to grow!) but now I'm being super picky!

CBarrett2: The problem is that sometimes socketEvent is null so it will throw an error if I try to check socketEvent.type and socketEvent is null. I could use, say `var successfulClose = socketEvent && socketEvent.type === 'close' && socketEvent.code === 1000; perhaps

PR #157, opened 2015-08-20
Split out series definitions from main

CBarrett2:
Thought it might be worth taking what we can out of main into another file. Just an experiment

#118


CBarrett2:
Unsure about naming of the files/appropriate namespaces for the functions

djmiley:
LGTM

jleft:
Really nice, great to see the code in main being modularised, definitely a step in the right direction. I've continued this in #194 to change the structure slightly :)

PR #158, opened 2015-08-21
Feature/indicator price used

### LINKED ISSUE #149: Price Used for Indicator Calculation
Choice between close/low/high/open for the moving algorithm and bollinger bands.

RuthC-C:
Added buttons to change the price used to calculate the moving average and the bollinger bands - only able to do open/high/low/close right now.
resolves #149

### DIFF
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -123,6 +123,18 @@
             return primaryChart;
         };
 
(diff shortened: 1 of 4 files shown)


jleft:
This looks like something which could be modularised into its own component (or at least its own file), rather than residing in main - if that seems reasonable?

djmiley:
Utilised in #250

Issue #146, opened 2015-08-20
Series and indicators are drawn on top of eachother depending on when they're changed


CBarrett2:
ie. if you select bollinger bands after candlestick, they are drawn on top of the candlestick series, which is not desirable imo

djmiley:
Should have been closed in #160

PR #160, opened 2015-08-24
Series/indicators shown in constant order

CBarrett2:
Series always shown in front of indicators. Perhaps debatable which way round we want it - we might want the lines (eg moving average line) on top of the main series, but the fill/area of bollinger bands definitely behind. Either way, at least this is consistent.

### DIFF
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -62,7 +62,8 @@
                     return index;
                 }
                 return series;
-            });
+            })
+            .series([gridlines, currentSeries, closeAxisAnnotation]);
 
         function primaryChart(selection) {
             var data = selection.datum().data;
--- a/src/assets/js/chart/primaryChart.js
+++ b/src/assets/js/chart/primaryChart.js
@@ -63,7 +63,7 @@
                 }
                 return series;
             })
-            .series([gridlines, currentSeries, closeAxisAnnotation]);
+            .series([gridlines, currentSeries, closeLine]);
 
         function primaryChart(selection) {
             var data = selection.datum().data;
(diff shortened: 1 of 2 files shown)


CBarrett2:
Resolves #146

src/assets/js/chart/primaryChart.js
```diff
@@ -61,18 +61,18 @@
...
         function primaryChart(selection) {
             var data = selection.datum().data;
             var viewDomain = selection.datum().viewDomain;
+
             timeSeries.xDomain(viewDomain);
 
             movingAverage(data);
             bollingerAlgorithm(data);
 
-            updateMultiSeries();
```
djmiley: #188

src/assets/js/main.js
```diff
@@ -22,12 +22,15 @@
     var seriesOptions = sc.menu.optionGenerator()
         .on('optionChange', function(seriesType) {
             primaryChart.changeSeries(seriesType.series);
+            // Remove already drawn series/indicators, so they're redraw in the correct order
+            svgPrimary.selectAll('.multi').remove();
```
jleft: Do we have to re-render all of the series? Can we make use of multi's key function to maintain the correct ordering?

CBarrett2: The order of rendering is determined by the order the series appear in the html. I don't know if multi's key function will account for this? It may do, I'm not really familiar with it. But removing them and re-drawing them definitely works, and I don't think efficiency matters too much in this case as it only happens when a main series is changed

jleft: Ok. Ideally we wouldn't remove all and then re-render, but there is a limitation of d3fc data join/multi series which means we can't keep the correct ordering of elements (easily).

An enhancement will be in soon for this, so the ordering of the elements and their corresponding data will be maintained - resolving this issue.

For now, let's add a comment explaining why we have to do this, as multi can't guarantee the correct ordering, and we can merge this, then revert when the enhancement lands in a d3fc release.

`.remove();` on a new line.

jleft:
Thanks @CBarrett2

PR #162, opened 2015-08-24
Added button to change period of live data

CBarrett2:


tlsim:
I'm having issues with the chart (candlestick width and navchart line don't look right, latest price line does not update) when switching between bitcoin and generated when 1min or 5min intervals are set. Can you reproduce this?

CBarrett2:
This is because I didnt rebase my changes on top of the hotfix I did yesterday - it's the same bug, I believe :) rebased now, so should be fixed when I can push my changes

CBarrett2:
This isn't ready to review yet

PR #163, opened 2015-08-24
fixed major bug

CBarrett2:
where websocket returns latestBasket along with a closeEvent, which gets added to the new fake data when it shouldn't be!

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -133,11 +133,10 @@
     }
 
     function liveCallback(socketEvent, latestBasket) {
-        var successfulOpen = socketEvent.type === 'open';
-        var successfulClose = socketEvent.type === 'close' && socketEvent.code === 1000;
         if (socketEvent.type === 'message' && latestBasket) {
             newBasketReceived(latestBasket);
-        } else if (!successfulOpen && !successfulClose) {
+        } else if (socketEvent.type === 'error' ||
+            (socketEvent.type === 'close' && socketEvent.code !== 1000)) {
             console.log('Error loading data from coinbase websocket: ' +
                 socketEvent.type + ' ' + socketEvent.code);
         }
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -135,13 +135,12 @@
     function liveCallback(socketEvent, latestBasket) {
         if (socketEvent.type === 'message' && latestBasket) {
             newBasketReceived(latestBasket);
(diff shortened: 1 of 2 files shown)


src/assets/js/main.js
```diff
@@ -133,17 +133,13 @@
...
-        }
-
-        if (socketEvent) {
-            var successfulOpen = socketEvent.type === 'open';
-            var successfulClose = socketEvent.type === 'close' && socketEvent.code === 1000;
-            if (!successfulOpen && !successfulClose) {
-                console.log('Error loading data from coinbase websocket: ' +
-                    socketEvent.type + ' ' + socketEvent.code);
-            }
+        } else if (!successfulOpen && !successfulClose) {
```
jleft: I think this is trying to catch a WebSocket error, or a WebSocket closure which is not normal?

Maybe we should check for these, to increase readability (can assign to vars too, if it's easier to understand)?

`socketEvent.type === "error" || (socketEvent.type === 'close' && socketEvent.code !== 1000)`

jleft:
Good catch - please could an issue be created for this too (steps to reproduce, expected behaviour and actual behaviour)?

src/assets/js/main.js
CBarrett2: Okay, but we also need to ignore usual open events

jleft: > Okay, but we also need to ignore usual open events

Yeah, this should do that.

CBarrett2: Sorry, I misunderstood the code. You're right, I'll change this

CBarrett2:
Resolves #164

CBarrett2:
Made the changes :)

jleft:
> Resolves #164

Thanks :smiley: 

---

Just FYI on branch names - [hotfixes](http://nvie.com/posts/a-successful-git-branching-model/#hotfix-branches) usually fix critical bugs which are found in production (i.e. code which has been merged into the master branch).

Gitflow does [lack the concept of a distinct 'bugfix' branch](https://github.com/nvie/gitflow/issues/24), but that's because the branches are categorised by how they're _used_. As this this issue was found in the development branch, the workflow is exactly the same as a feature (we're branching off develop and merging back into develop). In future, I'd probably name the branch `feature/fix-some-bug` or `bugfix/some-bug`, but this is fine for now :-).

---

> Made the changes :)

:+1:

PR #166, opened 2015-08-24
fixed panning bug

### LINKED ISSUE #165: Add Loading Screen for Bitcoin data loading
Panning while a historic data callback comes back means that any subsequent movement from the same pan will move the new data according to the previous data's scale. 

To reproduce: change from generated data to live data then quickly start to pan on the generated data and hold the mouse down. When the new data loads, if you move the mouse (while still held down), the domain will shift to something similar to the domain of the previous set of data.

This could be fixed by adding a loading screen/hiding the charts - it depends how we want to deal with it.

CBarrett2:
Resolves #165 reasonably well by only changing domain if it's in the bounds of the data, but there are other possible solutions eg removing the charts when new data is being loaded

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -32,7 +32,12 @@
         });
 
     function onViewChanged(domain) {
-        dataModel.viewDomain = [domain[0], domain[1]];
+        if (domain[0].getTime() < dataModel.data[0].date.getTime() ||
+            domain[1].getTime() > dataModel.data[dataModel.data.length - 1].date.getTime()) {
+            resetToLive();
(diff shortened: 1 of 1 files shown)


jleft:
Nice idea, but there still seems to be an issue where the extent rect will only snap back into place when the mouse is moved and not when the data is loaded (very minor issue though).

I think I'm leaning towards showing a loading indicator and hiding the charts.

CBarrett2:
I agree, I think we should add a loading screen - will comment on the bug - perhaps we should create an issue for this too?

PR #171, opened 2015-08-25
Update historic data feed start and end date when selecting Bitcoin data

### LINKED ISSUE #170: Bug: current time not updated when historic calls are made
The start and end dates for historic data are only set on initialization, meaning that 5 minutes into loading the program, historic calls will no longer get up-to-date data, while the live feed is, leaving a gap in the data

CBarrett2:
Resolves #170

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -109,13 +109,8 @@
         onViewChanged(standardDateDisplay);
     }
 
-    var currDate = new Date();
-    var startDate = d3.time.minute.offset(currDate, -200);
-
     var historicFeed = fc.data.feed.coinbase()
-        .granularity(60)
-        .start(startDate)
-        .end(currDate);
+        .granularity(60);        
 
     var callbackGenerator = sc.util.callbackInvalidator();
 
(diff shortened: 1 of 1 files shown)


src/assets/js/main.js
```diff
@@ -159,10 +154,20 @@
         return callbackGenerator(onHistoricDataLoaded);
     }
 
+    function setHistoricDates() {
+        var currDate = new Date();
+        var startDate = d3.time.minute.offset(currDate, -200);
+        historicFeed.start(startDate)
+        .end(currDate);
```
jleft: Indent.

src/assets/js/main.js
```diff
@@ -159,10 +154,20 @@
         return callbackGenerator(onHistoricDataLoaded);
     }
 
+    function setHistoricDates() {
```
jleft: `updateHistoricFeedDateRangeToPresent`?

src/assets/js/main.js
```diff
@@ -159,10 +154,20 @@
...
+        var currDate = new Date();
+        var startDate = d3.time.minute.offset(currDate, -200);
+        historicFeed.start(startDate)
+        .end(currDate);
+    }
+
+    setHistoricDates();
```
jleft: Do we need this? Won't this always be done in the change event handler for type selection?

PR #174, opened 2015-08-26
Added button to change period of live data

CBarrett2:
There seems to be a bug where sometimes while loading 12 hour data no data is returned between ~3 and 15th august. I am sure the right request is being made though! 

Also wondering if now is a good time to split the data stuff out into a separate class, which contains ohlcWebsocketConverter and historicFeed, and deals with invalidating callbacks/changing both periods at once/ setting date start/ends automatically

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -56,13 +56,24 @@
         onViewChanged(standardDateDisplay);
     }
 
-    var historicFeed = fc.data.feed.coinbase()
-        .granularity(60);
+    function changePeriod(period) {
+        historicFeed.granularity(period);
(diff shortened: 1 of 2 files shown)


CBarrett2:
The aforementioned bug seems to be something which only occurs between those dates, at whatever period, but only intermittently - sometimes the exact same steps will end in that data being loaded, and other times not.

CBarrett2:
Resolves #179

jleft:
Could any of this be pulled out of main, into a new components? Especially the interaction logic.

CBarrett2:
Created  #206, where I tried to pull this out of main, not sure if that's what you had in mind?

PR #175, opened 2015-08-26
Concat JavaScript source files into a single file and add tests for filterDataInDateRange

jleft:
Just a simple starting point to get the ball rolling for an improved build and test process...

- Adds a very simple step to the build process to concatenate the source JavaScript files into a single file; this should tidy the HTML and (hopefully!) be a bit easier to manage as there's now a single point to update the source paths for both the build and the unit tests (as discussed we'll also look into dependency management options, at a later date, to help with this).

- Adds some tests for `filterDataInDateRange`, just to demonstrate that this works - I'd really appreciate if someone would confirm that the tests cover the expected behaviour!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 
 # Built files
 dist
+
+# Jasmine spec runner
+_SpecRunner.html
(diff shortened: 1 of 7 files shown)


djmiley:
LGTM

CBarrett2:
LGTM2

jleft:
Thanks @djmiley and @CBarrett2

Issue #71, opened 2015-08-11
Create mobile-friendly menu


jleft:

djmiley:
Closed in #176

PR #172, opened 2015-08-25
Removed svg dependence in dimension calculating

djmiley:
Resolves #62

PR #176, opened 2015-08-26
Simple slideout

djmiley:
Utilises #172 

 - Option buttons moved to slideout
 - Utilises container, rather than svg, dependence
 - Transitions employed for secondary chart toggle

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -87,7 +87,17 @@
             }
         });
 
-    var mainMenu = sc.menu.main()
+    var headMenu = sc.menu.head()
+        .on('resetToLive', resetToLive)
+        .on('toggleSlideout', function() {
+            container.selectAll('.row-offcanvas-right').classed('active',
+                !container.selectAll('.row-offcanvas-right').classed('active'));
+        });
+
+    container.selectAll('.head-menu')
+        .call(headMenu);
+
+    var sideMenu = sc.menu.side()
         .on('primaryChartSeriesChange', function(series) {
             primaryChart.changeSeries(series);
             /* Elements are drawn in the order they appear in the HTML - at this minute,
@@ -126,10 +136,8 @@
             dataInterface(dataModel.period);
         });
 
-    container.select('.menu')
-        .call(mainMenu);
-
-    container.select('#reset-button').on('click', resetToLive);
+    container.selectAll('.sidebar-menu')
+        .call(sideMenu);
 
     d3.select(window).on('resize', resize);
 
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -38,7 +38,7 @@
     }
 
     function resize() {
-        sc.util.calculateDimensions(container, secondaryChart);
+        sc.util.dimensions.layout(container, secondaryChart);
         render();
     }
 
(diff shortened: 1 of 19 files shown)


djmiley:
Resolves #71

src/assets/js/main.js
```diff
@@ -39,12 +39,22 @@
...
         .call(mainMenu);
 
+    // Set Toggle menu event
+    function toggleMenu() {
+        container.selectAll('.row-offcanvas-right').classed('active',
+            !container.selectAll('.row-offcanvas-right').classed('active'));
+    }
+
+    container.select('#toggle-button').on('click', toggleMenu);
+
```
djmiley: Left this in, as `container.select('#reset-button').on('click', resetToLive);` is still in `main.js`

src/index.html
```diff
@@ -29,27 +46,27 @@
...
+        <div class="col-xs-12 col-sm-12 col-md-9">
           <svg class="chart nav"></svg>
         </div>
+        <div class="col-xs-6 col-sm-6 col-md-3 sidebar-offcanvas">
+          NAV
+        </div>
       </div>
+
     </div>
+
```
djmiley: Just movement of menu into `sidebar-offcanvas` of `#primary-row`, and addition of bootstrap columns for all screen widths.

djmiley:
Deployed this to gh-pages so that people can see how it looks.

http://djmiley.github.io/d3fc-showcase/

src/assets/js/main.js
jleft: I think this should be moved into `sc.menu.main`. The reset button should be moved in due course too (separate issue).

src/assets/js/main.js
```diff
@@ -138,8 +146,14 @@
     }
 
     function resize() {
-        sc.util.calculateDimensions(container, secondaryChart);
-        render();
+        // Re-render for smooth transitions
```
jleft: Did you consider using [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) or D3's [transitions](https://github.com/mbostock/d3/wiki/Transitions)?

jleft: If we're not throttling/denouncing the calls on window resize, isn't this likely to be called a lot and likely stutter? Especially when dragging the edge of the window to resize.

Would we want to animate during a window resize?

src/assets/js/main.js
djmiley: It can't go in `sc.menu.main` as this renders it inaccessible. I should have realised that earlier. Renaming to `sc.menu.slideout` and creating `sc.menu.head` for the reset and toggle button is something I could do?

jleft: > It can't go in sc.menu.main as this renders it inaccessible

Could we not do this to attach the low-level `click` event handling in `sc.menu.main`, then dispatch an event (which we'd expose) so main could listen for it?

> Renaming to sc.menu.slideout and creating sc.menu.head for the reset and toggle button is something I could do?

Sounds like a good idea to me :)

jleft:
:+1: 

We can look at the styling/layout of buttons etc. within the menu separately #240.

PR #189, opened 2015-08-26
Track live data

### LINKED ISSUE #178: Pan chart automatically to track new streaming data
Number of visible data points stays the same

CBarrett2:
Resolves #178

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -41,8 +41,14 @@
         render();
     }
 
+    var trackingLiveData = true;
     function onViewChanged(domain) {
         dataModel.viewDomain = [domain[0], domain[1]];
+        var latestViewedTime = dataModel.viewDomain[1].getTime();
+        var lastDatumTime = dataModel.data[dataModel.data.length - 1].date.getTime();
+        if (latestViewedTime === lastDatumTime) {
+            trackingLiveData = true;
+        }
         render();
     }
 
@@ -68,6 +74,10 @@
                 socketEvent.type + ' ' + socketEvent.code);
             } else if (socketEvent.type === 'message') {
                 dataModel.data = data;
+                if (trackingLiveData) {
+                    var newDomain = sc.util.returnDomainShiftedToEndOfData(dataModel.viewDomain, dataModel.data);
+                    onViewChanged(newDomain);
+                }
             }
             render();
         })
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -48,6 +48,8 @@
         var lastDatumTime = dataModel.data[dataModel.data.length - 1].date.getTime();
         if (latestViewedTime === lastDatumTime) {
             trackingLiveData = true;
+        } else {
+            trackingLiveData = false;
         }
         render();
     }
--- a/src/assets/js/util/returnDomainShiftedToEndOfData.js
+++ b/src/assets/js/util/returnDomainShiftedToEndOfData.js
@@ -1,0 +1,9 @@
+(function(d3, fc, sc) {
+    'use strict';
+    sc.util.returnDomainShiftedToEndOfData = function(domain, data) {
+        var extentTime = domain[1].getTime() - domain[0].getTime();
+        var latestDate = data[data.length - 1].date;
+        return [new Date(latestDate.getTime() - extentTime), latestDate];
+    };
+
+})(d3, fc, sc);
--- a/test/util/returnDomainShiftedToEndOfDataSpec.js
+++ b/test/util/returnDomainShiftedToEndOfDataSpec.js
@@ -1,0 +1,30 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    describe('sc.util.returnDomainShiftedToEndOfData', function() {
+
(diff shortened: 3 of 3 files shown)


src/assets/js/main.js
```diff
@@ -31,8 +31,15 @@
...
         });
 
+    var trackingLive = true;
     function onViewChanged(domain) {
         dataModel.viewDomain = [domain[0], domain[1]];
+        if (domain[1].getTime() === dataModel.data[dataModel.data.length - 1].date.getTime()) {
+            trackingLive = true;
+        } else {
+            trackingLive = false;
+        }
```
djmiley: `trackingLive = (domain[1].getTime() === dataModel.data[dataModel.data.length - 1].date.getTime())`?

Also, do you need `.getTime()` ?

CBarrett2: Sadly it does need getTime() as otherwise === compares whether the objects are the same instance, and == seems to not be liked by jshint or jscs. Good suggestion for the assignment though

djmiley: I thought that may be the case, but thought i'd check!

src/assets/js/main.js
```diff
@@ -67,18 +67,25 @@
...
-        .period(60);
+        .period(period);
 
     function newBasketReceived(basket) {
         var data = dataModel.data;
         if (data[data.length - 1].date.getTime() !== basket.date.getTime()) {
-            data.push(basket);
+            if (dataModel.viewDomain[1].getTime() === dataModel.data[dataModel.data.length - 1].date.getTime()) {
+                data.push(basket);
+                var newDomain = sc.util.moveDomainToEnd(dataModel.domain, data);
```
tlsim: Should the first argument be `dataModel.viewDomain` ?

src/assets/js/util/moveDomainToEnd.js
```diff
@@ -0,0 +1,12 @@
+(function(d3, fc, sc) {
+    'use strict';
+    sc.util.moveDomainToEnd = function(domain, data) {
+        var extentTime = domain[1].getTime() - domain[0].getTime();
+        var latestDate = data[data.length - 1].date;
+        var movedDomain = [domain[0], domain[1]];
```
tlsim: No need to initialize like this - you modify both items in the next 2 lines.

CBarrett2: It's to avoid changing the domain passed in by reference

src/assets/js/main.js
```diff
@@ -67,18 +67,25 @@
...
 
     var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor()
-        .period(60);
+        .period(period);
 
     function newBasketReceived(basket) {
         var data = dataModel.data;
         if (data[data.length - 1].date.getTime() !== basket.date.getTime()) {
-            data.push(basket);
+            if (dataModel.viewDomain[1].getTime() === dataModel.data[dataModel.data.length - 1].date.getTime()) {
```
tlsim: Both sides of `===` are good candidates for variables with descriptive names.

src/assets/js/util/moveDomainToEnd.js
```diff
@@ -0,0 +1,12 @@
+(function(d3, fc, sc) {
+    'use strict';
+    sc.util.moveDomainToEnd = function(domain, data) {
```
tlsim: The name kind of suggests that the function modifies `domain`, but it actually returns a new array. It's also not obvious what 'End' means. Maybe `calculateLatestDomain` or similar?

test/util/moveDomainToEndSpec.js
```diff
@@ -0,0 +1,30 @@
...
+        function obj(val) {
+            return {
+                date: new Date(val)
+            };
+        }
+
+        it('should keep the extent size the same', function() {
+            var extent = [new Date(1000), new Date(6000)];
+            var data = [obj(1000), obj(10000)];
+            var newExtent = sc.util.moveDomainToEnd(extent, data);
```
tlsim: Decide on one of 'extent' or 'domain' for consistency.

test/util/moveDomainToEndSpec.js
```diff
@@ -0,0 +1,30 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    describe('sc.util.moveDomainToEnd', function() {
+
+        beforeEach(function() {});
+
+        function obj(val) {
```
tlsim: Making datum objects is likely to be common to many tests (see filterDataInDateRangeSpec). Can you factor it out (with a nicer name)?

src/assets/js/util/moveDomainToEnd.js
tlsim: This wouldn't mutate `domain`, right?:
```
return [
    new Date(latestDate.getTime() - extentTime),
    new Date(latestDate.getTime())
];
```

tlsim:
This seems to be forcing shifting the domain to view live data even when panned away from the latest data point. Is this the intended behaviour?

djmiley:
Continued in #228

PR #192, opened 2015-08-27
Bollinger Band moving average line hidden

### LINKED ISSUE #183: Bollinger band should not automatically show moving average

RuthC-C:
Moving average line for Bollinger Bands has been hidden in CSS.
resolves #183


jleft:
Looks good, I think this belongs in `plotArea.less` though, rather than `base.less`.

jleft:
:+1: Nice fix, I'm pleased there was a simple solution for this!

PR #194, opened 2015-08-28
Modularise menu

jleft:
A continuation of #157. Also adds:

- A top-level menu component, to handle the main menu
- Use of option generators moved out of main and into separate modules for each menu section
- A more generic 'option' class for use with the option generator

Still TODO: all of the logic for changing the data source is still in main - I think this should be pulled out into its own module - if we're in agreement, another issue can be created for this.


src/assets/js/menu/option/generator.js
```diff
@@ -0,0 +1,40 @@
...
+                        .datum();
+                    dispatch.optionChange(selectedOption);
+                });
+        }
+
+        d3.rebind(optionGenerator, dispatch, 'on');
+
+        return optionGenerator;
+    };
+
```
djmiley: In light of the `menu/option.js` file, and an option being a more generic structure, could this be renamed to 'menu/option/groupGenerator.js` or similar. For #187, I'm looking at the idea of a `toggleGenerator`, so this isn't the only generator.

src/assets/js/menu/option.js
```diff
@@ -0,0 +1,11 @@
+(function(d3, fc) {
+    'use strict';
+    sc.menu.option = function(displayString, valueString, option) {
+        return {
+            displayString: displayString,
+            valueString: valueString,
+            option: option
+        };
+    };
```
djmiley: Was just thinking about this last night! :+1:

djmiley:
I like this.

src/assets/js/menu/option/generator.js
jleft: Nice idea - how about `optionGroupGenerator`? Could then add `toggleGroupGenerator` later too?

djmiley: SGTM

djmiley: In the `menu` or `menu/option` directory?

jleft: I was thinking `menu/option/group/generator.js` for the path and `sc.menu.option.group.generator` for the component name?

djmiley: Yes, I like the sound of that. :+1:

jleft: Done (882c997)

PR #197, opened 2015-08-28
Introduced MACD chart

### LINKED ISSUE #186: MACD multi-part series plotted in another panel

djmiley:
Resolves #186

### DIFF
--- a/src/assets/js/chart/macdChart.js
+++ b/src/assets/js/chart/macdChart.js
@@ -1,6 +1,6 @@
-(function(sc) {
+(function(d3, fc, sc) {
     'use strict';
-    // Helper functions
+
     sc.chart.macdChart = function() {
         var dispatch = d3.dispatch('viewChange');
 
@@ -43,4 +43,4 @@
         return macdChart;
     };
 
-})(sc);
+})(d3, fc, sc);
--- a/src/assets/js/chart/macdChart.js
+++ b/src/assets/js/chart/macdChart.js
@@ -1,0 +1,46 @@
+(function(sc) {
+    'use strict';
+    // Helper functions
+    sc.chart.macdChart = function() {
+        var dispatch = d3.dispatch('viewChange');
+
+        var macdAlgorithm = fc.indicator.algorithm.macd();
+
+        var macd = fc.indicator.renderer.macd();
+
+        function macdChart(selection) {
+            var data = selection.datum().data;
+            var viewDomain = selection.datum().viewDomain;
+
+            macdAlgorithm(data);
+
+            var maxYExtent = d3.max(data, function(d) {
+                return Math.abs(d.macd.macd);
+            });
+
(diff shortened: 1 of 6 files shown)


src/assets/js/main.js
```diff
@@ -127,11 +127,10 @@
         svgPrimary.datum(dataModel)
             .call(primaryChart);
 
+        svgSecondary.selectAll('*').remove();
```
jleft: Is there a way to avoid this?

src/assets/js/chart/macdChart.js
```diff
@@ -0,0 +1,46 @@
+(function(sc) {
+    'use strict';
+    // Helper functions
```
jleft: Remove comment

src/assets/js/chart/macdChart.js
```diff
@@ -0,0 +1,46 @@
+(function(sc) {
```
jleft: Also pass in `d3` and `fc`

src/assets/js/chart/macdChart.js
```diff
@@ -0,0 +1,46 @@
...
+    'use strict';
+    // Helper functions
+    sc.chart.macdChart = function() {
```
jleft: This is consistent, so we'll leave it for now, but I think I'd consider removing `Chart` from the end:

`sc.chart.macdChart ` - > `sc.chart.macd`
`sc.chart.primaryChart` -> `sc.chart.primary` 
etc.

djmiley: Very happy to do so in this request if wanted?

jleft: I'd leave it for now and solve it in another PR #199

src/assets/js/chart/macdChart.js
```diff
@@ -0,0 +1,46 @@
...
+        var dispatch = d3.dispatch('viewChange');
+
+        var macdAlgorithm = fc.indicator.algorithm.macd();
+
+        var macd = fc.indicator.renderer.macd();
+
+        function macdChart(selection) {
```
jleft: Would we also expect this to create a Y-axis?

djmiley: I'll work on this.

djmiley: So the RSI component in d3fc comes 'free' with annotations but the MACD component does not.

jleft: If the y-axis/annotations are going to be tricky to add, we could pull it out as a separate issue?

djmiley: I think it's non-trivial so it would be a good idea.

jleft: Okay, it can be dealt with separately - #200

jleft:
:+1:

PR #203, opened 2015-08-29
Fixed panning bug

### LINKED ISSUE #202: Bug- can pan/zoom past end of data when using the secondary chart to do so

djmiley:
Resolves #202


jleft:
Good spot!

PR #206, opened 2015-08-31
Added change period button, refactors/modularized

### LINKED ISSUE #179: Change Intra day period

CBarrett2:
Resolves #179
Created dataInterface class to contain data logic. 
Not sure if this is like what you had in mind @jleft 
Tried to keep it as stateless as possible. Does remove a reasonable amount of code from main, and definitely tidies it up.

### DIFF
--- a/src/assets/js/data/dataInterface.js
+++ b/src/assets/js/data/dataInterface.js
@@ -8,8 +8,7 @@
         var dispatch = d3.dispatch('messageReceived', 'historicDataLoaded');
 
         function dataInterface(period) {
-            callbackGenerator.invalidateCallback();
-            ohlcConverter.close();
+            dataInterface.invalidate();
             historicFeed.granularity(period);
             ohlcConverter.period(period);
             updateHistoricFeedDateRangeToPresent(period);
--- a/src/assets/js/data/dataInterface.js
+++ b/src/assets/js/data/dataInterface.js
@@ -7,6 +7,7 @@
         var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
         var dataGenerator = fc.data.random.financial();
         var dispatch = d3.dispatch('messageReceived', 'dataLoaded');
+        var candlesOfData = 200;
 
         function dataInterface(period) {
             dataInterface.invalidate();
@@ -25,7 +26,7 @@
 
         dataInterface.generateData = function() {
             dataInterface.invalidate();
-            dispatch.dataLoaded(null, dataGenerator(200));
+            dispatch.dataLoaded(null, dataGenerator(candlesOfData));
             return dataInterface;
         };
 
@@ -39,7 +40,7 @@
 
         function updateHistoricFeedDateRangeToPresent(period) {
             var currDate = new Date();
-            var startDate = d3.time.second.offset(currDate, -200 * period);
+            var startDate = d3.time.second.offset(currDate, -candlesOfData * period);
             historicFeed.start(startDate)
                 .end(currDate);
         }
--- a/src/assets/js/data/dataInterface.js
+++ b/src/assets/js/data/dataInterface.js
@@ -5,7 +5,8 @@
         var historicFeed = fc.data.feed.coinbase();
         var callbackGenerator = sc.util.callbackInvalidator();
         var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
-        var dispatch = d3.dispatch('messageReceived', 'historicDataLoaded');
+        var dataGenerator = fc.data.random.financial();
+        var dispatch = d3.dispatch('messageReceived', 'dataLoaded');
 
         function dataInterface(period) {
             dataInterface.invalidate();
@@ -18,9 +19,15 @@
                     currentData = data.reverse();
                     ohlcConverter(liveCallback(currentData), currentData[currentData.length - 1]);
                 }
-                dispatch.historicDataLoaded(err, currentData);
+                dispatch.dataLoaded(err, currentData);
             }));
         }
+
+        dataInterface.generateData = function() {
+            dataInterface.invalidate();
+            dispatch.dataLoaded(null, dataGenerator(200));
+            return dataInterface;
+        };
 
         dataInterface.invalidate = function() {
             ohlcConverter.close();
--- a/src/assets/js/data/dataInterface.js
+++ b/src/assets/js/data/dataInterface.js
@@ -1,0 +1,59 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.data.dataInterface = function() {
+        var historicFeed = fc.data.feed.coinbase();
+        var callbackGenerator = sc.util.callbackInvalidator();
+        var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
+        var currentData = [];
+        var dispatch = d3.dispatch('messageReceived', 'historicDataLoaded');
+
+        function dataInterface(period) {
+            callbackGenerator.invalidateCallback();
+            ohlcConverter.close();
+            historicFeed.granularity(period);
+            ohlcConverter.period(period);
+            updateHistoricFeedDateRangeToPresent(period);
+            historicFeed(callbackGenerator(function(err, data) {
+                if (!err) {
+                    currentData = data.reverse();
+                    ohlcConverter(liveCallback, currentData[currentData.length - 1]);
+                }
+                dispatch.historicDataLoaded(err, currentData);
+            }));
+        }
+
+        dataInterface.invalidate = function() {
+            ohlcConverter.close();
+            callbackGenerator.invalidateCallback();
+            return dataInterface;
+        };
+
+        d3.rebind(dataInterface, dispatch, 'on');
+
+        function updateHistoricFeedDateRangeToPresent(period) {
+            var currDate = new Date();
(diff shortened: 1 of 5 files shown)


RuthC-C:
Occasionally missing chunks of data when on the Bitcoin Data
image

RuthC-C:
When changing from Bitcoin to Data Generator it doesn't show anything:
image

Then when you pan/zoom it shows bitcoin data again:
image

RuthC-C:
Make the nav bar selection the same width on 1 min as it is on 5min/1hr:
image
image

src/assets/js/main.js
```diff
@@ -41,6 +42,23 @@
                 secondaryChart.on('viewChange', onViewChanged);
             }
             resize();
+        })
+        .on('dataTypeChange', function(type) {
+            if (type === 'bitcoin') {
+                dataInterface(dataModel.period);
```
jleft: Should we be updating the period here? By default won't we get daily data, even though the select will be set to 1hr?

src/assets/js/menu/main.js
```diff
@@ -22,9 +24,27 @@
                 dispatch.secondaryChartChange(chart);
             });
 
+        var dataTypeChangeOptions = function(selection) {
+            selection.on('change', function() {
+                var type = d3.select(this).property('value');
```
jleft: Can we just do `this.value`? Same for `periodChangeOptions`.

src/index.html
```diff
@@ -22,15 +22,19 @@
...
         <div class="col-md-12">
-          <select id="type-selection" selected="type" class="form-control">
-            <option value="generated">Data generator</option>
-            <option value="bitcoin">Bitcoin data</option>
-          </select>
           <div id="series-buttons" class="btn-group" data-toggle="buttons"></div>
           <div id="indicator-buttons" class="btn-group" data-toggle="buttons"></div>
           <div id="secondary-chart-buttons" class="btn-group" data-toggle="buttons"></div>
-          <div class="pull-right" id="reset-button">
-            <button class="btn btn-default" type="button">Reset</button>
```
jleft: Reset button removed?

src/index.html
```diff
@@ -22,15 +22,19 @@
...
+          <div class="col-md-2">
```
jleft: Any reason to add this column back in?

src/assets/js/main.js
```diff
@@ -49,78 +67,42 @@
...
-        return callbackGenerator(onHistoricDataLoaded);
-    }
+            }
+            if (socketEvent.type === 'message') {
+                dataModel.data = data;
+            }
+            render();
+        })
+        .on('historicDataLoaded', function(err, data) {
+            if (err) {
```
jleft: If we get an error, would we still want to render?

src/assets/js/main.js
```diff
@@ -49,78 +68,43 @@
     // Set Reset button event
     function resetToLive() {
```
jleft: Any reason of this change in functionality?

src/assets/js/main.js
```diff
@@ -49,78 +68,43 @@
...
-        if (socketEvent.type === 'message' && latestBasket) {
-            newBasketReceived(latestBasket);
-        } else if (socketEvent.type === 'error' ||
-            (socketEvent.type === 'close' && socketEvent.code !== 1000)) {
-            console.log('Error loading data from coinbase websocket: ' +
+    var dataInterface = sc.data.dataInterface()
+        .on('messageReceived', function(socketEvent, data) {
+            if (socketEvent.type === 'error' ||
+                (socketEvent.type === 'close' && socketEvent.code !== 1000)) {
+                console.log('Error loading data from coinbase websocket: ' +
```
jleft: We'll want better error handling in the future - I'll create an issue.

src/assets/js/main.js
```diff
@@ -49,78 +68,43 @@
...
-            updateDataAndResetChart(newData.reverse());
-            ohlcConverter(liveCallback, newData[newData.length - 1]);
-        } else { console.log('Error getting historic data: ' + err); }
-    }
-
-    function historicCallback() {
-        return callbackGenerator(onHistoricDataLoaded);
-    }
+            }
+            if (socketEvent.type === 'message') {
```
jleft: Would `else if` make sense here?

src/assets/js/main.js
```diff
@@ -49,78 +68,43 @@
...
 
-    function updateHistoricFeedDateRangeToPresent() {
-        var currDate = new Date();
-        var startDate = d3.time.minute.offset(currDate, -200);
-        historicFeed.start(startDate)
-            .end(currDate);
+    function setPeriodChangeVisibility(visible) {
+        var visibility = visible ? 'visible' : 'hidden';
+        d3.select('#period-selection')
+        .style('visibility', visibility);
```
jleft: Indent.

Ideally this should be outside of main. Would doing this in `dataTypeChangeOptions` make sense?

src/assets/js/menu/main.js
```diff
@@ -22,9 +24,27 @@
                 dispatch.secondaryChartChange(chart);
             });
 
+        var dataTypeChangeOptions = function(selection) {
+            selection.on('change', function() {
+                //var type = d3.select(this).property('value');
```
jleft: Remove

src/assets/js/menu/main.js
```diff
@@ -22,9 +24,27 @@
...
+                dispatch.dataTypeChange(this.value);
+            });
+        };
+
+        var periodChangeOptions = function(selection) {
+            selection.on('change', function() {
+                //var period = d3.select(this).property('value');
```
jleft: Remove

src/index.html
```diff
@@ -21,18 +21,25 @@
     <div class="container" id="app-container">
 
       <div class="row menu" id="head-row">
-        <div class="col-md-12">
```
jleft: Content should be placed in columns, so we still need this (http://getbootstrap.com/css/#grid-intro)

src/index.html
```diff
@@ -21,18 +21,25 @@
...
-          </select>
+
           <div id="series-buttons" class="btn-group" data-toggle="buttons"></div>
           <div id="indicator-buttons" class="btn-group" data-toggle="buttons"></div>
           <div id="secondary-chart-buttons" class="btn-group" data-toggle="buttons"></div>
           <div class="pull-right" id="reset-button">
             <button class="btn btn-default" type="button">Reset</button>
           </div>
-        </div>
+          <div class="col-md-2">
```
jleft: Do we need this column?

src/assets/js/main.js
CBarrett2: It's not strictly to do with the PR, but Ruth pointed out that the old way displays a variable sized nav bar brush on reset/init as it's based on the number of points, and sometimes points are missing in real data, so it will go further back in time than it should.

src/assets/js/data/dataInterface.js
```diff
@@ -0,0 +1,61 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.data.dataInterface = function() {
```
jleft: Just an idea... could this also manage the data generator source? Or would it be worth considering creating another component to manage this and the data generator - reducing the logic in main?

src/assets/js/main.js
CBarrett2: I'm not really sure! Perhaps we could have this done in a separate listener in another file, not sure which would be most appropriate though

src/index.html
CBarrett2: I think so, otherwise the period-selection dropdown goes onto the next line and fills the entire thing. There may be another way of doing things though

src/assets/js/data/dataInterface.js
CBarrett2: It could manage the data generator also, but I think the only thing it would add is to automatically call dataInterface.invalidate() when fake data is asked for, everything else would just be a direct wrapping/rebinding of the data generator. Only one line of code would be removed from main as far as I can see

src/index.html
jleft: If we're nesting a column within a column, the nested column, should be wrapped in a `row`.

How about using a solution similar to that used for `#type-selection` - https://github.com/ScottLogic/d3fc-showcase/blob/develop/src/assets/styles/menu.less#L3-L8 - these properties could be used in a class and applied to both `#type-selection` and `period-selection`?

src/assets/js/data/dataInterface.js
```diff
@@ -0,0 +1,67 @@
...
+                    currentData = data.reverse();
+                    ohlcConverter(liveCallback(currentData), currentData[currentData.length - 1]);
+                }
+                dispatch.dataLoaded(err, currentData);
+            }));
+        }
+
+        dataInterface.generateData = function() {
+            dataInterface.invalidate();
+            dispatch.dataLoaded(null, dataGenerator(200));
```
jleft: Final comment, then this is good to go :)

Can we pull `200` out into a descriptive variable? As it's used here and in `updateHistoricFeedDateRangeToPresent` below.

CBarrett2:
I should probably point out that this can still break if you make too many requests, but that's a problem from the API - possibly should be handled in another issue.

jleft:
LGTM, thanks @CBarrett2

jleft:
> I should probably point out that this can still break if you make too many requests, but that's a problem from the API - possibly should be handled in another issue.

#216

PR #211, opened 2015-09-02
SVG now respects padding

### LINKED ISSUE #210: Dimension calculations for chart SVGs don't respect padding
We should subtract the horizontal padding from the width to obtain the appropriate dimension for the container, to prevent the chart from overflowing.

dimensions-padding

djmiley:
Resolves #210


djmiley:
Leave this, don't merge!

PR #215, opened 2015-09-03
Changed bollinger styling to new color scheme

### LINKED ISSUE #184: Bollinger Band Styling
Red for Upper
Green for Lower
Grey for Fill

See google doc.

djmiley:
Resolves #184


src/assets/styles/bollinger.less
```diff
@@ -0,0 +1,19 @@
+.plot-area {
```
jleft: I think these would be fine in `plotArea.less` - we can always add comments, so we know which rules relate to which series/indicator.

src/assets/styles/bollinger.less
```diff
@@ -0,0 +1,19 @@
...
+
+  .average > path.line{
```
jleft: Space between `line` and `{` - same for below too.

src/assets/styles/variables.less
```diff
@@ -9,6 +9,11 @@
 @candlestick-down-fill: #c60;
 @candlestick-up-fill: #6c0;
 
+// Bollinger
```
jleft: Again, I'd probably group under the plot area heading. Or split up the plot area variables into annotation, moving average and candlestick.

src/assets/styles/variables.less
```diff
@@ -9,6 +9,11 @@
...
+@bollinger-upper-colour: #0f0;
```
jleft: Would it make sense to use the `#c60` and `#6c0` (from candlestick) as the upper and lower colours - then we'd have consistent shades of red/green? Might look rubbish though - I'm not sure!

djmiley:
Yes, happy to change the PR for those suggestions

src/assets/styles/variables.less
```diff
@@ -2,15 +2,18 @@
...
 
 // Plot Area
-@annotation-text-colour: #fff;
-@moving-average-line-colour: #f0f;
 @candlestick-down-stroke: #000;
 @candlestick-up-stroke: #000;
 @candlestick-down-fill: #c60;
 @candlestick-up-fill: #6c0;
+@moving-average-line-colour: #f0f;
+@bollinger-upper-colour: #6c0;
```
jleft: Maybe pull `#6c0` as another variable - `@primary-green`? And `#c60` - `@primary red`?

Should the colours be the other way around? #184:

> Red for Upper
Green for Lower

djmiley: Doh.

jleft:
:+1:

PR #217, opened 2015-09-03
Implemented toggle of up to two secondary charts

### LINKED ISSUE #187: Add upto 2 secondary charts

djmiley:
Resolves #187


src/assets/js/chart/primaryChart.js
```diff
@@ -181,7 +181,7 @@
         };
 
         primaryChart.toggleIndicator = function(indicator) {
-            if (indicator.show) {
+            if (indicatorMulti.series().indexOf(indicator.option) === -1) {
                 indicatorMulti.series()
```
djmiley: Look at this!

src/assets/js/chart/primaryChart.js
```diff
@@ -44,7 +44,11 @@
...
         var extent = fc.util.extent(visibleData, extentAccessor);
 
-        if (indicators.length) {
+        if (currentIndicators.length) {
+            var indicators = [];
+            for (var i = 0; i < currentIndicators.length; i++) {
+                indicators.push(currentIndicators[i].valueString);
+            }
             var movingAverageShown = (indicators.indexOf('movingAverage') !== -1);
             var bollingerBandsShown = (indicators.indexOf('bollinger') !== -1);
```
djmiley: Is there an easier way to test on a shared property of an array of objects?

src/assets/js/chart/primaryChart.js
```diff
@@ -44,7 +44,11 @@
         }
         var extent = fc.util.extent(visibleData, extentAccessor);
 
-        if (indicators.length) {
+        if (currentIndicators.length) {
+            var indicators = [];
```
jleft:
```js
 var indicators = currentIndicators.map(function(indicator) { return indicator.valueString; });
```

src/assets/js/chart/primaryChart.js
jleft: Not sure what you mean?

src/assets/js/chart/primaryChart.js
djmiley: Placed my line note in the wrong position. Thanks!

src/assets/js/chart/primaryChart.js
```diff
@@ -102,14 +107,12 @@
...
-        var indicatorStrings = [];
-
         function updateMultiSeries() {
-            if (indicatorMulti.series()) {
-                multi.series([gridlines, currentSeries.option, closeLine, indicatorMulti]);
-            } else {
-                multi.series([gridlines, currentSeries.option, closeLine]);
+            multi.series([gridlines, currentSeries.option, closeLine]);
+            if (currentIndicators.length) {
+                for (var i = 0; i < currentIndicators.length; i++) {
```
jleft: Could use array [`map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) again here?

src/assets/styles/primary.less
```diff
@@ -0,0 +1,3 @@
+.primary-row {
+  width: 100%;
```
jleft: Do we need this? Doesn't `.row` have a width of 100%?

src/assets/styles/secondary.less
```diff
@@ -1,3 +1,7 @@
-#secondary-row > .col-chart {
+.secondary-row {
+  width: 100%;
```
jleft: Do we need this? Doesn't `.row` have a width of 100%?

src/assets/js/menu/generator/toggleGroup.js
```diff
@@ -0,0 +1,38 @@
...
+                    value: function(d) { return d.valueString; }
+                });
+        }
+
+        function toggleGenerator(selection) {
+            selection.call(layoutButtons);
+
+            selection.selectAll('.btn')
+                .on('click', function() {
+                    var toggledOption = d3.select(this)
```
jleft: Could we get the `checked` property from the `input` and send that in the event, so we know if the option has been toggled on or off?

jleft:
:+1:

PR #218, opened 2015-09-03
Feature/multiple indicator selection

### LINKED ISSUE #182: Allow multiple indicators to be displayed at the same time.

RuthC-C:
Can select both/neither/one of the indicators to show at a time.
resolves #182


RuthC-C:
the file movingAverage.js was meant to be deleted but hasn't.

RuthC-C:
Just rebased it to the most recent develop - also deleted the extra <code>primaryChart.changeIndicator</code> function at the bottom of the primary chart file

src/assets/js/chart/primaryChart.js
```diff
@@ -65,9 +64,17 @@
...
             .series([gridlines, currentSeries, closeLine]);
 
+        var indMulti = fc.series.multi()
+                    .key(function(series, index) {
+                        if (series.isLine) {
+                            return index;
+                        }
+                        return series;
+                    });
+
```
djmiley: Can we just use an array?

src/assets/js/chart/primaryChart.js
```diff
@@ -142,8 +149,16 @@
             return primaryChart;
         };
 
-        primaryChart.changeIndicator = function(indicator) {
-            currentIndicator = indicator;
+        primaryChart.toggleIndicator = function(indicator) {
+            var ind = indMulti.series();
```
djmiley: Can we just operate on `indMulti.series()` directly?

src/assets/js/menu/main.js
```diff
@@ -3,6 +3,17 @@
...
+            .decorate(function(select) {
+                select.enter()
+                    .classed('movingAverage', true);
+            })
+            .yValue(function(d) { return d.movingAverage; });
+
+        var movingAverageIndicator = sc.menu.option('Moving Average', 'movingAverage', movingAverage);
+        var bollingerIndicator = sc.menu.option('Bollinger Bands', 'bollinger', fc.indicator.renderer.bollingerBands());
+        var indicators = [movingAverageIndicator, bollingerIndicator];
+
```
djmiley: Can this not just be in `primaryChart/indicators.js`?

src/assets/js/menu/primaryChart/indicators.js
```diff
@@ -4,31 +4,29 @@
...
         };
 
+        primaryChartSeriesMenu.indicators = function(x) {
+            if (!arguments.length) {
+                return indicators;
+            }
+            indicators = x;
+            return primaryChartSeriesMenu;
+        };
+
```
djmiley: Similar, is this file not where we set our available indicators?

src/assets/js/chart/primaryChart.js
```diff
@@ -65,9 +64,17 @@
             })
             .series([gridlines, currentSeries, closeLine]);
 
+        var indMulti = fc.series.multi()
+                    .key(function(series, index) {
+                        if (series.isLine) {
+                            return index;
+                        }
```
djmiley: Do we need `.key` here?

src/assets/js/menu/main.js
RuthC-C: Keeping it out of <code>indicators.js</code> keeps said file cleaner and more generic - However I can move it back there.

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                            .property('checked');
+                        selectedIndicator = element
+                            .datum();
+                        selectedIndicator.show = checked;
+                        dispatch.indicatorSelect(selectedIndicator);
+                    }, 0);
+
+                });
+        }
+
```
RuthC-C: Like a button - I am not sure what you're getting at?

djmiley: Ok, seen it now.

src/assets/js/chart/primaryChart.js
```diff
@@ -38,19 +38,10 @@
...
-            extent[1] = Math.max(movingAverageExtent[1], extent[1]);
-        } else if (indicatorString === 'bollinger') {
-            var bollingerBandsVisibleDataObject = visibleData.map(function(d) { return d.bollingerBands; });
-            var bollingerBandsExtent = fc.util.extent(bollingerBandsVisibleDataObject, ['lower', 'upper']);
-            extent[0] = Math.min(bollingerBandsExtent[0], extent[0]);
-            extent[1] = Math.max(bollingerBandsExtent[1], extent[1]);
-        } else if (indicatorString !== 'no-indicator') {
-            throw new Error('Unexpected indicator type');
-        }
+        var bollingerBandsVisibleDataObject = data.map(function(d) { return d.bollingerBands; });
```
jleft: This changes the behaviour of the primary chart as it will show the same amount padding on the y scale regardless of whether indicators are visible or not. Is it possible to adjust this (as it was previously) based on which indicators are visible?

src/assets/js/chart/primaryChart.js
```diff
@@ -93,9 +82,11 @@
             })
             .series([gridlines, currentSeries, closeLine]);
 
+        var indMulti = fc.series.multi();
```
jleft: `indMulti` -> `indicatorMulti`?

Is using a multi preferable to using an array?

src/assets/js/chart/primaryChart.js
jleft: `ind` -> `indicators`?

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+    sc.menu.indicatorChoice = function() {
+        var dispatch = d3.dispatch('indicatorSelect');
+
+        function indicatorButtons(sel) {
+            sel.selectAll('label')
+                .data(sel.datum())
+                .enter()
+                .append('label')
+                .classed('btn btn-default', true)
+                .classed('checked', false)
```
jleft: Does this do anything?

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                .text(function(d, i) {return d.displayString; })
```
jleft:
```js
.text(function(d) { return d.displayString; })
```

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                .insert('input')
```
jleft: Any reason to use `insert` over `append`?

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                .property('checked', false)
+                .attr({
+                    type: 'checkbox',
+                    name: 'indicatorType',
+                    id: function(d) {
```
jleft: Could put this all on one line for consistency

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                        return d.valueString;
+                    },
+                    value: function(d, i) { return d.valueString; }
```
jleft: Could remove `i`, as it's not used.

src/assets/js/menu/indicatorChoice.js
```diff
@@ -0,0 +1,51 @@
...
+                });
+        }
+
+        function indicatorChoice(selection) {
+            selection.call(indicatorButtons);
+
+            selection.selectAll('.btn')
+                .on('click', function() {
+                    var element = d3.select(this);
+                    setTimeout(function() {
```
jleft: Do we need this?

src/assets/js/chart/primaryChart.js
RuthC-C: multi is needed over array as we are storing values that are technically series (i.e. <code>fc.series.line()</code> for moving average)

src/assets/js/chart/primaryChart.js
RuthC-C: For now I have changed it so that no matter which indicator you have showing it will <code>extent</code> as if the bollinger bands were showing - this makes the transition between moving average and bollinger bands smoother.

RuthC-C: And the <code>extent</code> is normal when no indicator is showing.

src/assets/js/chart/primaryChart.js
```diff
@@ -25,7 +25,7 @@
         return annotatedTickValues;
     }
 
-    function findTotalYExtent(visibleData, currentSeries, currentIndicator) {
+    function findTotalYExtent(visibleData, currentSeries, indicator) {
```
jleft: `indicator` or `indicators`?

src/assets/js/chart/primaryChart.js
```diff
@@ -38,18 +38,11 @@
         }
         var extent = fc.util.extent(visibleData, extentAccessor);
 
-        var indicatorString = currentIndicator.valueString;
-        if (indicatorString === 'movingAverage') {
-            var movingAverageExtent = fc.util.extent(visibleData, 'movingAverage');
-            extent[0] = Math.min(movingAverageExtent[0], extent[0]);
-            extent[1] = Math.max(movingAverageExtent[1], extent[1]);
-        } else if (indicatorString === 'bollinger') {
+        if (indicator.length > 0) {
```
jleft: Is it possible to check which indicators are selected and adjust the extent accordingly? With this implementation there can be a lot of padding displayed if moving average is on and bollinger bands is off: 

ma-padding

src/assets/js/chart/primaryChart.js
```diff
@@ -170,8 +163,17 @@
...
 
-        primaryChart.changeIndicator = function(indicator) {
-            currentIndicator = indicator;
+        primaryChart.toggleIndicator = function(indicator) {
+            if (indicator.show) {
+                indicatorMulti.series()
+                    .push(indicator.option);
+            } else {
+                var index = indicatorMulti.series()
+                    .indexOf(indicator.option);
```
jleft: Could this ever return `-1`?

src/assets/js/chart/primaryChart.js
jleft: An array is passed to `series()` on multi and the array is used internally, so I think we could use an array in place here. Let's go with multi for now then :)

src/assets/js/chart/primaryChart.js
```diff
@@ -38,18 +38,19 @@
...
-            extent[1] = Math.max(movingAverageExtent[1], extent[1]);
-        } else if (indicatorString === 'bollinger') {
-            var bollingerBandsVisibleDataObject = visibleData.map(function(d) { return d.bollingerBands; });
-            var bollingerBandsExtent = fc.util.extent(bollingerBandsVisibleDataObject, ['lower', 'upper']);
-            extent[0] = Math.min(bollingerBandsExtent[0], extent[0]);
-            extent[1] = Math.max(bollingerBandsExtent[1], extent[1]);
-        } else if (indicatorString !== 'no-indicator') {
-            throw new Error('Unexpected indicator type');
+        if (indicators.length > 0) {
+            if (indicators.indexOf('bollinger') > -1) {
```
jleft: If bollinger bands and moving average are selected, this assumes that the moving average's extent will always be between that of bollinger bands. This isn't necessarily always the case:

ma-crop

If both moving average and bollinger bands are selected, can we calculate the extent based upon bollinger bands, moving average and the series?

PR #222, opened 2015-09-04
Pad charts with an extra period for some series

CBarrett2:
Automatically detects which charts need padding rather than having it set on the charts themselves


djmiley:
Can `navChart` implement same kind of zoom function? Pan and zoom functionality may need to be split.

djmiley:
Took this, leaving this here for now.

djmiley:
Continued in #277

PR #228, opened 2015-09-07
Track live data

### LINKED ISSUE #178: Pan chart automatically to track new streaming data

djmiley:
Resolves #178 .

Continues #189

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -41,8 +41,14 @@
         render();
     }
 
+    var trackingLiveData = true;
     function onViewChanged(domain) {
         dataModel.viewDomain = [domain[0], domain[1]];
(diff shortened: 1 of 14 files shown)


src/assets/js/main.js
```diff
@@ -68,6 +74,10 @@
                 socketEvent.type + ' ' + socketEvent.code);
             } else if (socketEvent.type === 'message') {
                 dataModel.data = data;
+                if (trackingLiveData()) {
```
tlsim: Tested with 1min live update - I don't see the domain shift when the incoming datum has a date greater than the 'latest viewed'.

djmiley:
Good to merge.

PR #229, opened 2015-09-07
Reduced horizontal padding on primary chart

djmiley:
Reduces horizontal padding from 10% either side to 4% either side, in reference to Rob's comments in the sprint demo.


jleft:
:+1:

PR #230, opened 2015-09-08
Zoom-in sticks to live data

### LINKED ISSUE #185: Zoom should not detach from latest price while tracking live data
#178

djmiley:
Resolves #185


djmiley:
See #228

djmiley:
`trackingLiveData` needs to be thought about.

djmiley:
I've thought of a way this might work if zoom is refactored

PR #235, opened 2015-09-10
Refactored zoom behaviour

djmiley:
- Attempted componentisation of zoom
- Wrapped SVGs in chart container and pass datamodel to that
- Added transparent foreground rectangle to primary chart


djmiley:
Superseded by #253

PR #242, opened 2015-09-11
Fixed less for data source and period length

djmiley:


src/assets/styles/menu.less
```diff
@@ -1,19 +1,17 @@
...
     min-width: 0;
     width: auto;
     display: inline;
     vertical-align: middle;
   }
+}
 
-  [id*="-buttons"] {
-    padding-top: @option-group-padding;
+.sidebar-menu [id*="-buttons"]{
```
jleft: Can we use `.btn-group` instead of `[id*="-buttons"]`? Also put this selector within `.sidebar-menu` and add a space before the opening `{`

src/assets/styles/menu.less
```diff
@@ -1,19 +1,19 @@
...
-  [id*="-buttons"] {
-    padding-top: @option-group-padding;
-
-    .active {
-      background-color: @toggled-background-colour;
-      color: @toggled-text-colour;
-    }
-  }
+.sidebar-menu {
+ .btn-group {
```
jleft: Two space indentation, please. Also below.

djmiley: Very sorry, that was quite silly of me.

jleft: No worries :)

jleft:
Thanks :+1:

PR #245, opened 2015-09-15
Moved y-axis to right of secondary charts

### LINKED ISSUE #200: Add y-axis to MACD panel

### LINKED ISSUE #241: Y axis on right side of RSI
Y axis needs to be on the right side of the RSI chart for blog post.
(requested by Rob)

djmiley:
Resolves #200 . Resolves #241 

Re-implemented secondary charts using `linearTimeSeries`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/scottlogic/d3fc-showcase#readme",
   "dependencies": {
     "bootstrap": "^3.3.5",
-    "d3fc": "^1.0.0"
+    "d3fc": "^1.2.0"
   },
   "devDependencies": {
     "cordova": "^5.1.1",
--- a/src/assets/styles/primary.less
+++ b/src/assets/styles/primary.less
@@ -1,5 +1,4 @@
-.plot-area {
-
+.primary {
   path.line {
     stroke: @primary-blue;
 
@@ -33,6 +32,4 @@
   .area > path.area {
     fill: @bollinger-fill-colour;
   }
-
 }
-
--- a/src/assets/styles/secondary.less
+++ b/src/assets/styles/secondary.less
@@ -1,3 +1,16 @@
-.secondary-row > .col-chart {
-  padding-right : calc(@bootstrap-padding-width + @y-axis-width);
+.secondary {
+  .x-axis {
+    path {
+      stroke: @offwhite;
+    }
+  }
+
+  &.secondary-rsi {
+    .right-handle {
+      text {
+        display: none;
+      }
+    }
+  }
 }
+
--- a/src/assets/styles/variables.less
+++ b/src/assets/styles/variables.less
@@ -2,6 +2,8 @@
 @primary-green: #6c0;
 @primary-red: #c60;
 @primary-white: #f0f8ff;
+@primary-black: #000;
+@offwhite: #fefefe;
 
 // Buttons
 @toggled-background-colour: #5cb85c;
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -30,7 +30,9 @@
             svgSecondary.datum(dataModel)
                 .filter(function(d, i) { return i < secondaryCharts.length; })
                 .each(function(d, i) {
-                    d3.select(this).call(secondaryCharts[i]);
+                    d3.select(this)
+                        .attr('class', 'chart secondary ' + secondaryCharts[i].valueString)
+                        .call(secondaryCharts[i].option);
                 });
 
             svgXAxis.datum(dataModel)
@@ -141,11 +143,11 @@
                     render();
                 })
                 .on('secondaryChartChange', function(toggledChart) {
-                    if (secondaryCharts.indexOf(toggledChart.option.option) !== -1 && !toggledChart.toggled) {
-                        secondaryCharts.splice(secondaryCharts.indexOf(toggledChart.option.option), 1);
+                    if (secondaryCharts.indexOf(toggledChart.option) !== -1 && !toggledChart.toggled) {
+                        secondaryCharts.splice(secondaryCharts.indexOf(toggledChart.option), 1);
                     } else if (toggledChart.toggled) {
                         toggledChart.option.option.on('viewChange', onViewChanged);
-                        secondaryCharts.push(toggledChart.option.option);
+                        secondaryCharts.push(toggledChart.option);
                     }
                     svgSecondary.selectAll('*').remove();
                     updateLayout();
--- a/src/assets/js/chart/macd.js
+++ b/src/assets/js/chart/macd.js
@@ -2,11 +2,35 @@
     'use strict';
 
     sc.chart.macd = function() {
+        var yAxisWidth = 45;
+
         var dispatch = d3.dispatch('viewChange');
 
-        var macdAlgorithm = fc.indicator.algorithm.macd();
+        var macdTimeSeries = fc.chart.linearTimeSeries()
+            .xAxisHeight(0)
+            .yAxisWidth(yAxisWidth)
+            .yOrient('right');
 
+        var zero = fc.annotation.line()
+            .value(0)
+            .label('');
         var macdRenderer = fc.indicator.renderer.macd();
+        var multi = fc.series.multi()
+            .series([zero, macdRenderer])
+            .mapping(function(series) {
+                if (series === zero) {
+                    return [0];
+                }
+                return this.data;
+            })
+            .decorate(function(g) {
+                g.enter()
+                    .attr('class', function(d, i) {
+                        return ['multi zero', 'multi'][i];
+                    });
+            });
+
+        var macdAlgorithm = fc.indicator.algorithm.macd();
 
         function macd(selection) {
             var data = selection.datum().data;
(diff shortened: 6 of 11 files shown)


src/assets/js/chart/macdChart.js
```diff
@@ -2,11 +2,29 @@
...
+            .yValue(function(d, i) { return d.macd.macd; });
+        var signal = fc.series.line()
+            .yValue(function(d, i) { return d.macd.signal; });
+        var divergence = fc.series.bar()
+            .yValue(function(d, i) { return d.macd.divergence; });
 
-        var macd = fc.indicator.renderer.macd();
+        var multi = fc.series.multi()
+            .series([zero, macd, signal, divergence]);
+
```
jleft: Can we use `fc.indicator.renderer.macd()`?

src/assets/js/chart/macdChart.js
```diff
@@ -14,33 +32,48 @@
...
-                .range([0, fc.util.innerDimensions(selection.node()).width]);
-            macd.yScale()
-                .domain([-maxYExtent, maxYExtent])
-                .range([fc.util.innerDimensions(selection.node()).height, 0]);
+            // Redraw
+            macdTimeSeries.plotArea(multi);
+            selection.call(macdTimeSeries);
 
+            // Color axis lines so y-axis is emphasised
+            selection.select('g.y-axis').selectAll('path').style('stroke', '#000');
```
jleft: Can this be done in Less?

src/assets/js/chart/rsiChart.js
```diff
@@ -2,42 +2,70 @@
...
+        var annotations = fc.annotation.line()
+            .label('');
+        var indicator = fc.series.line()
+            .yValue(function(d, i) { return d.rsi; });
 
-        var rsi = fc.indicator.renderer.relativeStrengthIndex()
-            .yScale(rsiScale);
+        var multi = fc.series.multi()
+            .series([annotations, indicator]);
+
```
jleft: Can we use `fc.indicator.renderer.relativeStrengthIndex()`?

src/assets/js/chart/rsiChart.js
```diff
@@ -2,42 +2,70 @@
...
+                            return 'multi ' + ['annotations', 'indicator'][i];
+                        });
+                });
+
+            // Redraw
+            rsiTimeSeries.plotArea(multi);
+            selection.call(rsiTimeSeries);
+
+            // Color y-axis so that it is not visible
+            selection.select('g.y-axis').selectAll('path').style('stroke', '#fefefe');
```
jleft: Can this be done in CSS?

src/assets/styles/variables.less
```diff
@@ -2,6 +2,7 @@
 @primary-green: #6c0;
 @primary-red: #c60;
 @primary-white: #f0f8ff;
+@d3fc-offwhite: #fefefe;
```
jleft: `@d3fc-offwhite` -> `@off-white`

src/assets/styles/primary.less
```diff
@@ -0,0 +1,41 @@
+.primary {
+  .plot-area {
```
jleft: As linear time series will introduce a plot area to the secondary charts, should these styles also apply to the secondary chart too to be consistent?

src/assets/js/chart/macdChart.js
djmiley: Need to look at this more

src/assets/js/chart/rsiChart.js
djmiley: Need to look at this more

djmiley: I think this would be uglier.

src/assets/js/chart/macdChart.js
djmiley: I think this would be uglier

src/assets/js/chart/rsiChart.js
jleft: If the issue is binding data appropriately, maybe it's multi (particularly the mapping functionality) to the rescue (again!)?

Something like this:

```js
var rsi = fc.indicator.renderer.relativeStrengthIndex();
var multi = fc.series.multi()
    .series([rsi])
    .mapping(function() {
            return data.data;
        }
    });
```

djmiley: Ok, yes, I was missing something. Thanks. :)

src/assets/js/chart/macdChart.js
```diff
@@ -2,11 +2,23 @@
...
+
         var dispatch = d3.dispatch('viewChange');
 
-        var macdAlgorithm = fc.indicator.algorithm.macd();
+        var macdTimeSeries = fc.chart.linearTimeSeries()
+            .xAxisHeight(0)
+            .yAxisWidth(yAxisWidth)
+            .yOrient('right');
 
+        var zero = fc.annotation.line()
```
jleft: No too fussed either way, but could just use a line?

src/assets/js/chart/macdChart.js
```diff
@@ -14,33 +26,45 @@
...
+            var paddedYExtent = sc.util.domain.padYDomain([-maxYExtent, maxYExtent], 0.04);
+            macdTimeSeries.yDomain(paddedYExtent);
 
-            macd.xScale()
-                .domain(viewDomain)
-                .range([0, fc.util.innerDimensions(selection.node()).width]);
-            macd.yScale()
-                .domain([-maxYExtent, maxYExtent])
-                .range([fc.util.innerDimensions(selection.node()).height, 0]);
+            multi.mapping(function(series) {
```
jleft: Can we set the mapping and decorate outside of this function, where multi is initialised?

src/assets/js/chart/rsiChart.js
```diff
@@ -2,42 +2,60 @@
...
-            rsi.xScale()
-                .domain(viewDomain)
-                .range([0, fc.util.innerDimensions(selection.node()).width]);
-            rsi.yScale().range([fc.util.innerDimensions(selection.node()).height, 0]);
+            rsiTimeSeries.xDomain(viewDomain)
+                .yDomain([0, 100]);
 
             rsiAlgorithm(data);
 
+            multi.mapping(function(series) {
```
jleft: Similar comment to above - can we set the mapping outside of this function, where multi is initialised?

src/assets/js/chart/rsiChart.js
```diff
@@ -2,42 +2,60 @@
...
+                    return data;
+                })
+                .decorate(function(g) {
+                    g.enter()
+                        .attr('class', function() {
+                            return 'multi';
```
jleft: Do we need to do this? I think this is the default behaviour?

src/assets/js/main.js
```diff
@@ -28,7 +28,12 @@
         svgSecondary.datum(dataModel)
             .filter(function(d, i) { return i < secondaryChart.length; })
             .each(function(d, i) {
-                d3.select(this).call(secondaryChart[i]);
+                var selection = d3.select(this);
```
jleft: How about this?

```js
d3.select(this)
    .attr('class', 'chart secondary ' + secondaryChart[i].valueString)
    .call(secondaryChart[i].option);
```

src/assets/styles/secondary.less
```diff
@@ -1,3 +1,17 @@
-.secondary-row > .col-chart {
-  padding-right : calc(@bootstrap-padding-width + @y-axis-width);
-}
\ No newline at end of file
+.secondary {
+  .x-axis {
+
```
jleft: Remove line

package.json
```diff
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/scottlogic/d3fc-showcase#readme",
   "dependencies": {
     "bootstrap": "^3.3.5",
-    "d3fc": "^1.0.0"
+    "d3fc": "1.0.0"
```
jleft: Can we use 1.0.1?

src/assets/js/chart/rsiChart.js
```diff
@@ -2,42 +2,53 @@
...
+        var yAxisWidth = 45;
+
         var dispatch = d3.dispatch('viewChange');
 
-        var rsiScale = d3.scale.linear()
-            .domain([0, 100]);
+        var rsi = fc.indicator.renderer.relativeStrengthIndex();
+        var multi = fc.series.multi()
+            .series([rsi])
+            .mapping(function(series) {
```
jleft: Don't need `series`.

src/assets/js/chart/macdChart.js
djmiley: Happy to remove it, but I think it makes the chart clearer.

jleft: I was thinking that maybe we could just use a `line` element or `fc.series.line()` (or maybe even an axis without any ticks), as I don't think we need any of the labelling which the annotation offers, but it's probably not worth changing... happy to stick with the annotation.

I agree, I do think it looks a lot nicer with the line :)

package.json
```diff
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/scottlogic/d3fc-showcase#readme",
   "dependencies": {
     "bootstrap": "^3.3.5",
-    "d3fc": "^1.0.0"
+    "d3fc": "1.0.1"
```
jleft: Sorry... 1.2.0 might be more relevant now?

jleft:
:+1: thanks @djmiley

PR #246, opened 2015-09-15
Replaced shift to live data with scalable function

djmiley:


src/assets/js/util/domain/scaleLiveDataDomain.js
```diff
@@ -0,0 +1,14 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.util.domain.scaleLiveDataDomain = function(domain, data, ratio) {
```
jleft: A couple of ideas:

+ `ratio` could be an optional parameter and default to `1` if  not provided (in this scenario, it's acting like it did previously, which I think is still useful).
+ I'm not sure about the name - the function seems to do two things: move the domain to the latest price, and (potentially) alter the length of the domain based upon the ratio - how about `sc.util.domain.moveToLatest`? (Assuming ratio is made an optional param.)

PR #247, opened 2015-09-15
Readded ticks to primary y-axis

djmiley:


src/assets/js/chart/primaryChart.js
```diff
@@ -134,13 +135,13 @@
...
-                            if (d === closePrice) {
-                                return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
-                            }
+                    s.filter(function(d) { return d === closePrice; })
+                        .each(function(d) {
+                            var selection = d3.select(this);
+                            selection.classed('closeLine', true);
+                            selection.select('path').attr('d', function(d) {
+                                    return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
+                                });
```
djmiley: Sounds silly, is this right indentation?

jleft: Looks fine to me :)

Do we need the `each` and can we chain the methods on the selection?

```js
s.filter(function(d) { return d === closePrice; })
    .classed('closeLine', true);
    .select('path')
    .attr('d', function(d) {
        return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
    });
```

djmiley: Yes, much nicer. Thanks.

src/assets/js/chart/primaryChart.js
```diff
@@ -134,14 +135,12 @@
...
-                        .select('path').attr('d', function(d) {
-                            if (d === closePrice) {
+                    s.filter(function(d) { return d === closePrice; })
+                        .classed('closeLine', true)
+                        .select('path')
+                        .attr('d', function(d) {
                                 return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
-                            }
-                        });
+                            });
```
jleft: Drop indentation a level for this and the line above

PR #248, opened 2015-09-15
Removed chart suffix from files

### LINKED ISSUE #199: Remove 'Chart' suffix from sc.chart modules
`sc.chart.primaryChart` -> `sc.chart.primary`
`sc.chart.rsiChart` -> `sc.chart.rsi`
etc.

djmiley:
Resolves #199


src/assets/js/app.js
```diff
@@ -18,30 +18,30 @@
             viewDomain: []
         };
 
-        var primaryChart = sc.chart.primaryChart();
-        var secondaryChart = [];
+        var primary = sc.chart.primary();
```
jleft: I'm not convinced about renaming these variables - I don't think it's as clear as it was previously. I wouldn't mind naming `navChart`, `navigator` though.

I'm also not sure about the event names - at this level in the application, is it as clear from the event name what's happening?

djmiley: I agree, I'll have a look.

djmiley: `navigator` is a reserved keyword

djmiley:
> I'm not convinced about renaming these variables - I don't think it's as clear as it was previously. I wouldn't mind naming navChart, navigator though.

> I'm also not sure about the event names - at this level in the application, is it as clear from the event name what's happening?

`navigator` is a reserved keyword

src/assets/js/app.js
jleft: Cool, didn't know that!

src/assets/js/menu/primary/indicators.js
```diff
@@ -20,7 +20,7 @@
                 dispatch.primaryChartIndicatorChange(indicator);
             });
 
-        var primaryChartSeriesMenu = function(selection) {
+        var primaryChartIndicatorMenu = function(selection) {
```
jleft: Good spot

jleft:
:+1:

PR #249, opened 2015-09-16
Refactored zoom to increase consistency

djmiley:
- Added transparent foreground rectangle for zooming on primary, rsi and macd charts
- Removed dataModel, data inconsistency
- Now pass selection into zoomControl, rather than selection and dataModel

### DIFF
--- a/src/assets/js/chart/macd.js
+++ b/src/assets/js/chart/macd.js
@@ -30,6 +30,11 @@
                     });
             });
 
+        var createForeground = fc.util.dataJoin()
+            .selector('rect.foreground')
+            .element('rect')
+            .attr('class', 'foreground');
+
         var macdAlgorithm = fc.indicator.algorithm.macd();
 
         function macd(selection) {
@@ -50,11 +55,8 @@
             macdTimeSeries.plotArea(multi);
             selection.call(macdTimeSeries);
 
-            selection.selectAll('rect.foreground')
-                .data([dataModel])
-                .enter()
-                .append('rect')
-                .attr('class', 'foreground')
+            var foreground = createForeground(selection, [dataModel])
+                .style('opacity', 0)
                 .layout({
                     position: 'absolute',
                     top: 0,
@@ -69,11 +71,11 @@
             var zoom = d3.behavior.zoom();
             zoom.x(macdTimeSeries.xScale())
                 .on('zoom', function() {
-                    sc.util.zoomControl(zoom, selection.select('rect.foreground'), macdTimeSeries.xScale());
+                    sc.util.zoomControl(zoom, foreground, macdTimeSeries.xScale());
                     dispatch.viewChange(macdTimeSeries.xDomain());
                 });
 
-            selection.select('rect.foreground').call(zoom);
+            foreground.call(zoom);
         }
 
         d3.rebind(macd, dispatch, 'on');
--- a/src/assets/js/chart/macd.js
+++ b/src/assets/js/chart/macd.js
@@ -33,15 +33,14 @@
         var macdAlgorithm = fc.indicator.algorithm.macd();
 
         function macd(selection) {
-            var data = selection.datum().data;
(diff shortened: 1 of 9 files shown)


src/assets/js/chart/macd.js
```diff
@@ -51,15 +50,30 @@
             macdTimeSeries.plotArea(multi);
             selection.call(macdTimeSeries);
 
+            selection.selectAll('rect.foreground')
```
jleft: Could be a good use-case for dataJoin?

```js
// Would pull this outside of the render loop
var createForeground = fc.util.dataJoin()
    .selector('rect.foreground')
    .element('rect')
    .attr('class', 'foreground');

createForeground(selection, [dataModel])
    .layout({
        position: 'absolute',
        top: 0,
        right: yAxisWidth,
        bottom: 0,
        left: 0
    });
```

src/assets/js/chart/macd.js
```diff
@@ -51,15 +50,30 @@
...
+                });
+
+            selection.layout();
+
             // Behaves oddly if not reinitialized every render
             var zoom = d3.behavior.zoom();
             zoom.x(macdTimeSeries.xScale())
                 .on('zoom', function() {
-                    sc.util.zoomControl(zoom, selection.select('.plot-area'), data, macdTimeSeries.xScale());
+                    sc.util.zoomControl(zoom, selection.select('rect.foreground'), macdTimeSeries.xScale());
```
jleft: Instead of selecting the foreground `rect`, could just assign it to a `var` when it's created (above), then use that (and below, where zoom is called)?

src/assets/styles/base.less
```diff
@@ -18,4 +18,7 @@ html, body {
 .chart {
   height: 100%;
   width: 100%;
+  rect.foreground {
+    opacity: 0;
```
jleft: I'd use `visibility: hidden;`

src/assets/js/util/zoomControl.js
```diff
@@ -1,23 +1,23 @@
 (function(d3, fc, sc) {
     'use strict';
 
-    sc.util.zoomControl = function(zoom, selection, data, scale) {
+    sc.util.zoomControl = function(zoom, selection, scale) {
         var tx = zoom.translate()[0];
-        var ty = zoom.translate()[1];
 
-        var xExtent = fc.util.extent(data, ['date']);
+        var xExtent = fc.util.extent(selection.datum().data, ['date']);
```
jleft: I like this, so data can be removed from the arguments.

Could just bind the data, rather than the entire data model, to the selection (i.e. the foreground `rect`) which this is being called on? Then zoom only has what it needs to know about.

djmiley: I'd like to stick with the dataModel, because my approach in #230 suggests that, for zooming to live data, it will want to know about the `dataModel.trackingLive` property?

src/assets/js/chart/macd.js
djmiley: Nice!

src/assets/js/util/zoomControl.js
jleft: Ok, sure, as we're using the same dataModel everywhere at the moment, we'll keep it like this as it's consistent.

Ultimately, I think we'll use different models for each component (as they'll need different things), but we can work on this as a bigger refactoring task later, after the core functionality is in.

src/assets/js/chart/primary.js
```diff
@@ -153,15 +156,27 @@
             timeSeries.plotArea(multi);
             selection.call(timeSeries);
 
+            var foreground = createForeground(selection, [dataModel])
+                .style('opacity', 0)
```
djmiley: > I'd use `visibility: hidden;`

This didn't work in allowing me to zoom, and using dataJoin set style to `opacity:1;` automatically.

PR #253, opened 2015-09-17
Componentised zoom

djmiley:
Continues #235


src/assets/js/behaviour/zoom.js
```diff
@@ -0,0 +1,50 @@
...
+            zoomBehaviour.x(zoomScale)
+                .on('zoom', function() {
+                    var scaleControlled = sc.util.zoom.controlScale(zoomBehaviour, selection, zoomScale);
+                    sc.util.zoom.controlPan(zoomBehaviour, selection, zoomScale);
+
+                    var panned = (zoomBehaviour.scale() === 1)  && !scaleControlled;
+                    var scaled = (zoomBehaviour.scale() !== 1) || scaleControlled;
+                    if ((panned && allowPan) || (scaled && allowScalable)) {
+                        dispatch.zoomed(zoomScale.domain());
+                    }
```
djmiley: Figured this out last night. If we replace lines 23 to 25 with:

                    var domain = zoomScale.domain();
                    if (selection.datum().trackingLive && (zoomBehaviour.scale() > 1)) {
                        domain = sc.util.domain.moveToLatest(zoomScale.domain(), selection.datum().data);
                    }
                    if ((panned && allowPan) || (scaled && allowScalable)) {
                        dispatch.zoomed(domain);
                    }

, this solves zoom sticking to live data without using the time tolerance in sc.util.domain.trackingLiveData

src/assets/js/behaviour/zoom.js
```diff
@@ -0,0 +1,50 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.behaviour.zoom = function(scale) {
```
jleft: As much as I love our British spelling, maybe change to `behavior` for consistency with D3?

Any reason to pass the scale on the factory method, rather than having a getter/setter for it?

src/assets/js/behaviour/zoom.js
```diff
@@ -0,0 +1,50 @@
...
+
+        var dispatch = d3.dispatch('zoomed');
```
jleft: `zoomed` -> `zoom`?

src/assets/js/behaviour/zoom.js
```diff
@@ -0,0 +1,50 @@
...
+
+        var zoomBehaviour = d3.behavior.zoom();
+        var zoomScale = scale;
+
+        var allowPan = true;
+        var allowScalable = true;
```
jleft: `allowZoom`?

src/assets/js/util/zoom/controlPan.js
```diff
@@ -0,0 +1,22 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.util.zoom.controlPan = function(zoom, selection, scale) {
```
jleft: I'd be tempted to move both this and controlScale into sc.behaviour.zoom as private functions, as I think they'll only ever be needed by the zoom component.

src/assets/js/behaviour/zoom.js
```diff
@@ -0,0 +1,50 @@
...
+
+        function zoom(selection) {
+
+            zoomBehaviour.x(zoomScale)
+                .on('zoom', function() {
+                    var scaleControlled = sc.util.zoom.controlScale(zoomBehaviour, selection, zoomScale);
```
jleft: `scaleControlled` felt a bit cryptic to me at first, I has to look at controlScale's implementation to work out what it was returning.

If controlScale handles the zoom functionality, then it could be named to reflect this. Inverting what it currently returns could lead to something like this, which I think is a bit easier to understand?

```js
// Or maybe 'zoomed' instead of 'hasZoomed'
var hasZoomed = zoom(zoomBehaviour, selection, zoomScale);
```

src/assets/js/behaviour/zoom.js
jleft: :+1:

PR #256, opened 2015-09-18
Fixed panning before indicators begin bug

### LINKED ISSUE #254: Panning before indicators breaks line render
capture

Rendering line breaks on `NaN` when panning to penultimate data before beginning of indicators on primary chart

djmiley:
Resolves #254

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -51,13 +51,13 @@
             if (bollingerBandsShown) {
                 var bollingerBandsVisibleDataObject = visibleData.map(function(d) { return d.bollingerBands; });
                 var bollingerBandsExtent = fc.util.extent(bollingerBandsVisibleDataObject, ['lower', 'upper']);
-                extent[0] = Math.min(bollingerBandsExtent[0], extent[0]);
-                extent[1] = Math.max(bollingerBandsExtent[1], extent[1]);
+                extent[0] = d3.min([bollingerBandsExtent[0], extent[0]]);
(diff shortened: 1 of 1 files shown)


jleft:
`d3.min()` and `d3.max()` both ignore `undefined` values, unlike `Math.min()` and `Math.max()`, which is what is what we want to do. Could we do something like this, which should solved the issue and avoid the loop before?

```js
if (bollingerBandsShown) {
    var bollingerBandsVisibleDataObject = visibleData.map(function(d) { return d.bollingerBands; });
    var bollingerBandsExtent = fc.util.extent(bollingerBandsVisibleDataObject, ['lower', 'upper']);
    extent[0] = d3.min([bollingerBandsExtent[0], extent[0]]);
    extent[1] = d3.max([bollingerBandsExtent[1], extent[1]]);
}
if (movingAverageShown) {
    var movingAverageExtent = fc.util.extent(visibleData, 'movingAverage');
    extent[0] = d3.min([movingAverageExtent[0], extent[0]]);
    extent[1] = d3.max([movingAverageExtent[1], extent[1]]);
}
```

src/assets/js/chart/primary.js
```diff
@@ -44,7 +44,18 @@
         }
         var extent = fc.util.extent(visibleData, extentAccessor);
 
-        if (currentIndicators.length) {
+        var visibleDataIndicatorsUndefined = (function() {
```
jleft: I think this assumes that if an indicator will has an `undefined` value on a datum, the value for all other indicators will also be `undefined`, but this may not necessarily be the case, for example if their window length is configured differently.

PR #257, opened 2015-09-21
Some refactoring of zoom

djmiley:
Definitely a work in progress!!!!


djmiley:
Resonance between render and zoom.translate causes the problems if not reinitialised

PR #264, opened 2015-09-21
Implement volume chart

### LINKED ISSUE #255: Volume chart
Possibly add a new secondary chart if an obvious candidate is available. Information on volume is available in both data generator and bitcoin data.

djmiley:
Resolves #255


djmiley:
#272

Issue #240, opened 2015-09-10
Improve styling of menu


jleft:

jleft:
No longer relevant (complete redesign).

PR #265, opened 2015-09-22
Introduce new style layout for d3fc-showcase

### LINKED ISSUE #165: Add Loading Screen for Bitcoin data loading

### LINKED ISSUE #205: Create loading screen/text and hide charts while loading

### LINKED ISSUE #231: Style theme used for d3fc-showcase
Styling similar to Techan.js. Implement a more professional UI.

djmiley:
Resolves #165. Resolves #205. Resolves #231.


djmiley:
#240

PR #251, opened 2015-09-17
Implemented dropdown menu generator

djmiley:
Part of #259

PR #269, opened 2015-09-23
Wrap model in object class and refactor data changes in head menu

djmiley:


tlsim:
#251

djmiley:
Can put some of the sc.util.domain function on here possibily or a similar construction

djmiley:
Close this for now, replaced in #284

PR #273, opened 2015-09-24
Fix for unlimited zooming

### LINKED ISSUE #168: Zoomed in extent is unlimited (Restrict zoom to 5 periods width)
It's possible to zoom in too far using the mousewheel causing a failed render:
`Invalid value for <path> attribute d="M-Infinity,-466.1657812499783L-Infinity` ... 

Zooming in on the primary or nav chart beyond the error causes the chart's zoom level to 'wrap around' and render using the full nav domain.

djmiley:
Resolves #168

### DIFF
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -10,6 +10,7 @@
 
         var allowPan = true;
         var allowZoom = true;
+        var minimumViewableTime = 0;
         var trackingLatest = true;
 
         function controlPan(zoomExtent) {
@@ -38,7 +39,12 @@
             var xExtent = fc.util.extent(selection.datum().data, ['date']);
             var width = selection.attr('width') || parseInt(selection.style('width'), 10);
 
+            var domain;
+
             zoomBehavior.x(scale)
+                .on('zoomstart', function() {
+                    domain = scale.domain();
+                })
                 .on('zoom', function() {
                     var min = scale(xExtent[0]);
                     var max = scale(xExtent[1]);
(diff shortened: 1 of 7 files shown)


djmiley:
#274

src/assets/js/behavior/zoom.js
```diff
@@ -10,6 +10,7 @@
 
         var allowPan = true;
         var allowZoom = true;
+        var minimumViewableTime = 0;
```
djmiley: Could use `minimumPeriods`, but this would require `sc.behavior.zoom` to make use of the `selection.datum().period` ?

src/assets/js/chart/nav.js
```diff
@@ -39,13 +39,31 @@
...
+                var minimumViewableTime = zoom.minimumViewableTime();
+                var brushTimeExtent = sc.util.timeExtent([brush.extent()[0][0], brush.extent()[1][0]]);
+
+                if (Math.abs(brushTimeExtent) >= minimumViewableTime) {
                     dispatch.viewChange([brush.extent()[0][0], brush.extent()[1][0]]);
+                } else if (Math.abs(brushTimeExtent) > 0) {
+                    var centeredDate = new Date((brush.extent()[1][0].getTime() + brush.extent()[0][0].getTime()) / 2);
+                    var centeredDomain = [d3.time.second.offset(centeredDate, -minimumViewableTime / 2),
+                        d3.time.second.offset(centeredDate, +minimumViewableTime / 2)];
+                    dispatch.viewChange(sc.util.domain.centerOnDate(centeredDomain, model.data, centeredDate));
```
djmiley: This is the bit that makes the brush move smoothly when selecting.

tlsim: Not sure what we want here. When selecting a new extent outside the existing extent by clicking and dragging, it feels weird that the new extent immediately goes to the minimum size, and continuing to drag moves the extent from the initial click position. Is this something that might be better handled on `brushend`?

djmiley:
Expected behavior : no crossing of brush lines

Issue #260, opened 2015-09-21
Technical Design


PaulScho:

djmiley:
(d3, fc, sc)
util.domain -> domain
dimensions.layout -> layout
callbackinvalidator -> data.callbackinvalidator
datamodel -> model

djmiley:
Create an object that deals with domain and data referencing in #269

jleft:
Calculating Y extent from indicators: 

> What if we put an extentAccessor property on the option when we initialise in sc.menu.side, i.e; yLowValue, ['lower', 'upper']? And then ask for the extentAccessor on multi.series.

https://github.com/ScottLogic/d3fc-showcase/pull/281/files#r40999610

jleft:
Closing as split into smaller tasks.

PR #276, opened 2015-09-29
Introduces sc.menu.group component

djmiley:


jleft:
#260

src/assets/js/menu/group.js
```diff
@@ -0,0 +1,47 @@
...
+
+    sc.menu.group = function() {
+
+        var dispatch = d3.dispatch('optionChange');
+
+        var option;
+        var generator;
+
+        function group(selection) {
+            var optionGenerator = generator
```
jleft: `.on('optionChange', function(option) {` on the same line as `generator`

I think this can also be moved outside of the `group` function?

jleft:
Looks good :+1: 

@tlsim - any thoughts?

tlsim:
:+1: Looks good to me.

jleft:
Thanks @djmiley and @tlsim :)

PR #277, opened 2015-09-30
Pads extent for some series

### LINKED ISSUE #89: Add padding to extent of data domain

djmiley:
Resolves #89


djmiley:
Reimplemented in #353

PR #281, opened 2015-10-01
Implements exponential moving average

### LINKED ISSUE #151: Different types of specific indicators
Simple vs. exponential vs. other types of moving average/bollinger bands ?

djmiley:
Resolves #151


src/assets/js/chart/primary.js
```diff
@@ -59,7 +60,12 @@
                 extent[0] = d3.min([movingAverageExtent[0], extent[0]]);
                 extent[1] = d3.max([movingAverageExtent[1], extent[1]]);
             }
-            if (!(movingAverageShown || bollingerBandsShown)) {
+            if (exponentialMovingAverageShown) {
+                var exponentialMovingAverageExtent = fc.util.extent(visibleData, 'exponentialMovingAverage');
+                extent[0] = Math.min(exponentialMovingAverageExtent[0], extent[0]);
```
jleft: Won't this blow up when there are only `undefined` values visible (at the start)?

Could use `d3.min()` and `d3.max()`, like above, to solve this.

I'd also consider calculating the extent just once, with all of the indicators' extents, rather than every time we check if an indicator is shown - might look a bit neater.

djmiley: What if we put an `extentAccessor` property on the option when we initialise in `sc.menu.side`, i.e; `yLowValue`, `['lower', 'upper']`?

jleft: Nice idea, definitely worth considering as part of #260

PR #282, opened 2015-10-01
Display none when secondary chart not shown

### LINKED ISSUE #280: There should be no gap between the charts and the x-axis
Would likely be fixed by setting `display: none;` on the hidden secondary charts, rather than setting their height to 0.

axis-gap

djmiley:
Resolves #280

### DIFF
--- a/src/assets/js/util/layout.js
+++ b/src/assets/js/util/layout.js
@@ -25,9 +25,11 @@
             .style('height', primaryHeightRatio * useableScreenHeight / totalHeightRatio + 'px');
         container.selectAll('.secondary-row')
             .filter(function(d, i) { return i < secondaryChartsShown; })
+            .style('display', 'block')
             .style('height', secondaryHeightRatio * useableScreenHeight / totalHeightRatio + 'px');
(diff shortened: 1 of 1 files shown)


src/assets/js/util/layout.js
```diff
@@ -25,9 +25,11 @@
             .style('height', primaryHeightRatio * useableScreenHeight / totalHeightRatio + 'px');
         container.selectAll('.secondary-row')
             .filter(function(d, i) { return i < secondaryChartsShown; })
+            .style('display', 'block')
             .style('height', secondaryHeightRatio * useableScreenHeight / totalHeightRatio + 'px');
         container.selectAll('.secondary-row')
             .filter(function(d, i) { return i >= secondaryChartsShown; })
+            .style('display', 'none')
             .style('height', '0px');
```
jleft: Do we still need to set the height to 0 here?

djmiley: It looks more consistent when you inspect the element, although it is not required. Happy to remove it.

jleft:
:+1:

PR #283, opened 2015-10-02
Adds test for all sc.util.domain functions

djmiley:


src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,7 +3,12 @@
 
     sc.util.domain.centerOnDate = function(domain, data, centerDate) {
         var dataExtent = fc.util.extent(data, 'date');
-        var domainTimeExtent = (domain[1].getTime() - domain[0].getTime()) / 1000;
+        var domainTimes = domain.map(function(d) { return d.getTime(); });
+        var domainTimeExtent = (d3.max(domainTimes) - d3.min(domainTimes)) / 1000;
```
jleft: Not sure about the naming, to me extent is an array with the min and max value from an array - `[d3.min(array), d3.max(array)]` - maybe `domainTimeDifference`?

src/assets/js/util/domain/padYDomain.js
```diff
@@ -1,10 +1,11 @@
 (function(d3, fc, sc) {
     'use strict';
-    sc.util.domain.padYDomain = function(domain, paddingPercentage) {
-        var paddedDomain = [];
-        var variance = domain[1] - domain[0];
-        paddedDomain[0] = domain[0] - variance * paddingPercentage;
-        paddedDomain[1] = domain[1] + variance * paddingPercentage;
-        return paddedDomain;
+    sc.util.domain.padYDomain = function(yExtent, paddingPercentage) {
+        var paddingArray = paddingPercentage.length ?
```
jleft: Maybe `Array.isArray(paddingPercentage)`?

src/assets/js/util/domain/padYDomain.js
```diff
@@ -1,10 +1,11 @@
...
+            [paddingPercentage[0], paddingPercentage[1]] : [paddingPercentage, paddingPercentage];
```
jleft: When `paddingPercentage` is an array is `[paddingPercentage[0], paddingPercentage[1]] === paddingPercentage`?

src/assets/js/util/domain/padYDomain.js
```diff
@@ -1,10 +1,11 @@
...
+        var variance = yExtent[1] - yExtent[0];
```
jleft: Does this assume `yExtent[1] >= yExtent[0]`? Should we use the absolute value from this?

src/assets/js/util/domain/trackingLatestData.js
```diff
@@ -3,7 +3,7 @@
 
     sc.util.domain.trackingLatestData = function(domain, data) {
         var dataExtent = fc.util.extent(data, 'date');
-        var latestViewedTime = domain[1].getTime();
+        var latestViewedTime = d3.max(domain.map(function(d) { return d.getTime(); }));
```
jleft: I wouldn't use `map()` here:

```js
d3.max(domain, function(d) { return d.getTime(); });
```

test/util/domain/centerOnDateSpec.js
```diff
@@ -0,0 +1,63 @@
...
+            expect(sc.util.domain.centerOnDate(domain, data, latestDate)).toEqual([wednesday, friday]);
+        });
+
+        it('should not center on a different date that is not contained within the data domain range', function() {
+            var centerDate = d3.time.day.offset(friday, 1);
+            expect(sc.util.domain.centerOnDate(domain, data, centerDate)).toEqual([tuesday, thursday]);
+        });
+
+        it('should return expected values when called with unordered data and domain', function() {
+            var centerDates = [thursday, wednesday, monday, d3.time.day.offset(friday, 1)];
```
jleft: I'd consider splitting this into more tests, as it seems to be doing exactly the same thing as the above tests, just with a different domain and different data.

Additional `describe` blocks could help organise the tests and common code can be factored out into functions to reduce repetition.

test/util/domain/moveToLatestSpec.js
```diff
@@ -32,5 +39,13 @@
             expect(newExtentSize).toBe(1000);
```
jleft: Should we also be checking the domain itself, and not just its size?

test/util/domain/moveToLatestSpec.js
```diff
@@ -32,5 +39,13 @@
...
         });
 
+        it('should return the data extent if the domain extent is too large', function() {
+            var extent = [new Date(1000), new Date(10000)];
+            var data = [obj(1000), obj(3000)];
+            var newExtent = sc.util.domain.moveToLatest(extent, data, 0.5);
+            var newExtentSize = newExtent[1].getTime() - newExtent[0].getTime();
+            expect(newExtentSize).toBe(2000);
```
jleft: Should we also be checking the domain itself, and not just its size?

test/util/domain/moveToLatestSpec.js
```diff
@@ -17,13 +17,20 @@
             expect(newExtentSize).toBe(5000);
         });
 
-        it('should move the extent to have its end at the last point of data', function() {
+        it('should move the extent to end at the last data point if extent ends before the data', function() {
             var extent = [new Date(1000), new Date(6000)];
             var data = [obj(1000), obj(10000)];
             var newExtent = sc.util.domain.moveToLatest(extent, data);
             expect(newExtent[1].getTime()).toBe(data[1].date.getTime());
```
jleft: Worth also checking the domain's size?

test/util/domain/moveToLatestSpec.js
```diff
@@ -17,13 +17,20 @@
...
         });
 
+        it('should move the extent to end at the last data point if extent ends after the data', function() {
+            var extent = [new Date(11000), new Date(16000)];
+            var data = [obj(1000), obj(10000)];
+            var newExtent = sc.util.domain.moveToLatest(extent, data);
+            expect(newExtent[1].getTime()).toBe(data[1].date.getTime());
```
jleft: Worth also checking the domain's size?

jleft:
Looking really good, great to have the comprehensive tests for the domain utils :+1:

src/assets/js/util/domain/filterDataInDateRange.js
```diff
@@ -2,12 +2,18 @@
     'use strict';
 
     sc.util.domain.filterDataInDateRange = function(domain, data) {
-        // Calculate visible data, given [startDate, endDate]
+        var startDate = d3.min(domain, function(d) { return d.getTime(); });
+        var endDate = d3.max(domain, function(d) { return d.getTime(); });
+
+        var dataSortedByDate = data.sort(function(a, b) {
+            return parseInt(a.date.getTime()) - parseInt(b.date.getTime());
```
jleft: Would just `a.date - b.date` work?

Be careful when using `parseInt` - without the radix parameter it may behave unpredictably:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

> radix
An integer between 2 and 36 that represents the radix (the base in mathematical numeral systems) of the above mentioned string. Specify 10 for the decimal numeral system commonly used by humans. **Always specify this parameter** to eliminate reader confusion and to guarantee predictable behavior. Different implementations produce different results when a radix is not specified.

djmiley: Yes, works.

PR #286, opened 2015-10-07
Added crosshair + added product and period entities

Afredon:
The primary chart now shows a crosshair and a legend.


Gruntfile.js
```diff
@@ -14,6 +14,9 @@ module.exports = function(grunt) {
             ],
             srcJsFiles: [
                 'src/assets/js/sc.js',
+                'src/assets/js/model/**/*Factory.js',
+                'src/assets/js/model/products.js',
+                'src/assets/js/model/**/*.js',
```
djmiley: Just `src/assets/js/model/**/*.js` ?

src/assets/js/chart/nav.js
```diff
@@ -75,4 +75,4 @@
 
         return nav;
     };
-})(d3, fc, sc);
\ No newline at end of file
+})(d3, fc, sc);
```
djmiley: Remove line

Gruntfile.js
Afredon: Unfortunately the alphabetic order is not correct for dependencies, hence the 3 lines. Factories need to be loaded before what they produce, and products must be loaded before the product_period association.
To simplify these lines we need proper dependencies management.

src/assets/js/chart/primary.js
```diff
@@ -216,4 +261,4 @@
 
         return primary;
     };
-})(d3, fc, sc);
\ No newline at end of file
+})(d3, fc, sc);
```
djmiley: Remove line

src/assets/js/menu/group.js
```diff
@@ -43,4 +52,4 @@
 
         return group;
     };
-})(d3, fc, sc);
\ No newline at end of file
+})(d3, fc, sc);
```
djmiley: Remove line

src/assets/js/menu/head.js
```diff
@@ -62,4 +53,4 @@
 
         return d3.rebind(head, dispatch, 'on');
     };
-})(d3, fc, sc);
\ No newline at end of file
+})(d3, fc, sc);
```
djmiley: Remove line

src/assets/js/menu/head.js
```diff
@@ -16,27 +16,18 @@
 
         setPeriodChangeVisibility(false);
 
-        var generated = sc.menu.option('Data Generator', 'generated', 'generated');
-        var bitcoin = sc.menu.option('Bitcoin Data', 'bitcoin', 'bitcoin');
+        function onOptionChange(product) {
+            setPeriodChangeVisibility(sc.model.getPeriodsAllowedForProduct(product.option).length > 1);
+            dispatch.dataProductChange(product);
+        }
```
djmiley: I don't think we need to pull this out as a function? If so, should rename it to `onDataProductChange`

src/assets/js/model/periodFactory.js
```diff
@@ -0,0 +1,33 @@
...
+            _timeFormat = config.timeFormat;
+        }
+
+        return {
+            display: _display,
+            seconds: _seconds,
+            timeFormat: d3.time.format(_timeFormat)
+        };
+    };
+})(d3, sc);
```
djmiley: Change to a d3 style of chaining calls, similar to this (unmerged) file. 
https://github.com/djmiley/d3fc-showcase/blob/8342fc43a23b982a58e1104da1adccce1cef019a/src/assets/js/model.js

src/assets/js/model/productFactory.js
```diff
@@ -0,0 +1,32 @@
...
+            _volumeFormat = config.volumeFormat;
+        }
+
+        return {
+            display: _display,
+            priceFormat: d3.format(_priceFormat),
+            volumeFormat: d3.format(_volumeFormat)
+        };
+    };
+})(d3, sc);
```
djmiley: Change to a d3 style of chaining calls, similar to this (unmerged) file. 
https://github.com/djmiley/d3fc-showcase/blob/8342fc43a23b982a58e1104da1adccce1cef019a/src/assets/js/model.js

src/assets/js/series/candlestick.js
```diff
@@ -59,7 +60,13 @@
...
             return candlestick;
         };
-
+        candlestick.xValue = function(x) {
+            if (!arguments.length) {
+                return xValue;
+            }
+            xValue = x;
+            return candlestick;
+        };
```
djmiley: Why do we add this? Is the crosshair implemented for all the series?

src/assets/js/series/candlestick.js
```diff
@@ -75,7 +82,14 @@
             yHighValue = x;
             return candlestick;
         };
+        candlestick.yCloseValue = function(x) {
+            if (!arguments.length) {
+                return yCloseValue;
+            }
+            yCloseValue = x;
+            return candlestick;
+        };
```
djmiley: Same as above

src/assets/styles/crosshair.less
```diff
@@ -0,0 +1,9 @@
+.crosshair line {
+  stroke: gray;
+  stroke-dasharray: 3, 10;
+}
+
+.crosshair circle {
+    r: 4;
+    fill: rgba(200, 200, 255, 0.8);
+}
```
djmiley: Bring line and circle into `.crosshair` selection, similar to `th` and `td` in `primary-legend.less`

src/assets/styles/primary-legend.less
```diff
@@ -0,0 +1,17 @@
...
+    line-height: 1em;
+    font-family: monospace;
+    th {
+        padding-right: 5px;
+        text-align: center;
+    }
+    td {
+        text-align: right;
+    }
+}
```
djmiley: Make `background-color` a variable in `variables.less`

src/assets/styles/crosshair.less
```diff
@@ -0,0 +1,9 @@
+.crosshair line {
+  stroke: gray;
+  stroke-dasharray: 3, 10;
+}
+
+.crosshair circle {
+    r: 4;
+    fill: rgba(200, 200, 255, 0.8);
```
djmiley: I'd probably make all of these variables, and can we change gray to # notation?

src/assets/js/model/periodFactory.js
Afredon: I did it d3 style at first, but it made everything a function, which was uncomfortable to use as it added a lot of parentheses to the code. Also it felt like violating the principle of less surprise when you have to write selectedProduct.display()
I would expect such code to actually display something.

djmiley:
Talking to @jleft the reason for (d3, fc, sc) is to ensure grunt uglify works correctly, so we need to implement it in all files.

src/assets/js/series/candlestick.js
Afredon: crosshair expect any component to respect an implicit interface which includes xValue. The optimised version of candlestick was missing this property, which made crosshair crash.

djmiley: Right yes, no problem.

src/assets/styles/primary-legend.less
djmiley: Also change file name to `primaryLegend.less` for consistency

src/assets/styles/crosshair.less
Afredon: I considered the variables (same than following comment), but there was no reuse of the value, it would introduce a level of indirection without a need for it. I can add it of course, it just seemed it wasn't needed for now.
If we have a rule saying the stroke color should be the same as the circle color, then it's definitely a variable, but I had no such rule at this point.

Afredon: and we can change gray to anything, it was just a suggestion

djmiley: I think it's more for making the less read quite simply, we currently decalre all color variables outside their less files.

src/assets/js/menu/group.js
```diff
@@ -31,6 +31,15 @@
...
 
+        group.options = function(array, adapter) {
+            var adapted = [];
+            var iElement;
+            for (iElement = 0; iElement < array.length; iElement++) {
+                adapted.push(adapter(array[iElement]));
+            }
+            return group.option.apply(group.option, adapted);
+        } ;
+
```
djmiley: Maybe change into `group.adaptor` and chain calls. @jleft  thoughts?

src/assets/js/model/periods.js
```diff
@@ -0,0 +1,26 @@
...
+        seconds: 3600,
+        timeFormat: '%b-%d %Hh'});
+    sc.model.period.minute5 = sc.model.periodFactory({
+        display: '5 Minutes',
+        seconds: 300,
+        timeFormat: '%H:%M'});
+    sc.model.period.minute1 = sc.model.periodFactory({
+        display: '1 Minute',
+        seconds: 60,
+        timeFormat: '%H:%M'});
```
djmiley: Can we just work with the inidivdually declared `sc.model.period.`, rather than the `sc.model.periods`? And simlilar for `sc.model.products`, we only use them in `group.options`.

src/assets/js/menu/head.js
Afredon: I didn't notice that Atom was adding these new lines to every file when I was saving, so I look around to understand why. It is important for file concatenation as it avoids some issues (like if you file ends by a comment, the first line of the next file concatenated will be ignored). I found the setting in Atom and can remove the new lines, but the arguments in favor of actually adding it everywhere were convincing, so let me know which you prefer.

src/assets/js/menu/head.js
Afredon: Just the habit of separating semantic levels as much as possible, but no problem to inline it back into the code.

jleft: Happy for this change to stay.

src/assets/js/chart/primary.js
```diff
@@ -116,16 +138,29 @@
...
+         * As the bands don't have to process any events, we make them ignoring all events so that
+         * they go to the next layer by setting the css style to pointer-events:none
+         *
+         * browser compatibility issue: pointer-events not implemented on Opera Mini 8
+         *
+         * @param {Array} indicators - Can contain or not bollingerBands renderer
+         */
+        function makeIndicatorsTransparentToMouseEvents(indicators) {
+            var decorator = function(g) { g.enter().style('pointer-events', 'none'); };
+            currentIndicators.map(function(indicator) { indicator.option.decorate(decorator); });
```
jleft: `map` -> `forEach`

src/assets/js/menu/group.js
Afredon: passing the collection and the adapter should not be separated as neither would work without the other.

jleft:
Incorporated in #294.

PR #288, opened 2015-10-09
Axis ticks

Afredon:
Data generated is now at midnight which allows to align with the daily ticks.
Zooming in now produces ticks which do not go under the granularity chosen, but reverts to d3 default behaviour when zooming out.


Afredon:
I worked out of the merge between add-crosshair and develop, I hope it is not a problem, as it seems to reinclude all the changes not yet validated from add-crosshair! (Git noob inside)

src/assets/js/chart/primary.js
```diff
@@ -116,16 +138,29 @@
...
+         * Bollinger bands are drawn on top of the main chart and capture mouse events.
+         * The crosshair needs these events to be working when the mouse is over the band.
+         * As the bands don't have to process any events, we make them ignoring all events so that
+         * they go to the next layer by setting the css style to pointer-events:none
+         *
+         * browser compatibility issue: pointer-events not implemented on Opera Mini 8
+         *
+         * @param {Array} indicators - Can contain or not bollingerBands renderer
+         */
+        function makeIndicatorsTransparentToMouseEvents(indicators) {
```
jleft: To avoid this, can we just draw the indicators under the crosshair? I expect all we'd need to do is change the order of the renderers in `updateMultiSeries()`, so the crosshairs are always the last element in the `baseChart` array.

src/assets/js/chart/primary.js
```diff
@@ -144,6 +179,16 @@
             }
         }
 
+        function setCrosshairSnap(series, data) {
+            crosshair.snap(fc.util.seriesPointSnapXOnly(series, data))
+            .on('trackingmove', function(crosshairData) {
```
jleft: Indent lines 184 to 189 an extra 4 spaces, please.

src/assets/js/chart/primary.js
```diff
@@ -67,14 +67,36 @@
...
+        }
         var yAxisWidth = 45;
-
         var dispatch = d3.dispatch('viewChange');
 
         var currentSeries = sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick());
         var currentYValueAccessor = function(d) { return d.close; };
         var currentIndicators = [];
 
+        var formatTime = function(x) {return sc.model.selectedPeriod.timeFormat(x); };
```
jleft: Add a space between `{` and `return` (same for two lines below too).

src/assets/js/data/dataInterface.js
```diff
@@ -48,10 +48,16 @@
...
 
-        dataInterface.generateData = function() {
+        dataInterface.generateDailyData = function() {
             dataInterface.invalidate();
-            dispatch.dataLoaded(null, dataGenerator(candlesOfData));
-            return dataInterface;
+
+            var now = new Date().getTime();
+            var millisecondsPerDay = 24 * 60 * 60 * 1000;
+            var roundedNow = millisecondsPerDay * Math.trunc(now / millisecondsPerDay);
```
jleft: When rounding the date, I'd probably use `now.setHours(0, 0, 0, 0)`, as I think it's a bit neater.

Otherwise I'd use `Math.floor` over `Math.trunc`, as support for the latter isn't great: https://kangax.github.io/compat-table/es6/#test-Math_methods_Math.trunc (No IE10 or IE11 support!)

src/assets/js/menu/generator/dropdownGroup.js
```diff
@@ -9,13 +9,19 @@
             .element('option');
 
         function layoutDropdown(sel) {
+            var _selectedIndex;
```
jleft: I'd remove the underscore for consistency (I don't think we do this anywhere else?).

src/assets/js/data/dataInterface.js
Afredon: I started with .setHours(0, 0, 0, 0) but I wanted something that would work the same with any period length and to show the rounding instead of overriding the behaviouir of the generator entirely, but as the daily generator is the only one needing rounding, I can switch back to .setHours

src/assets/js/model/periods.js
```diff
@@ -0,0 +1,47 @@
+(function(d3, fc, sc) {
+    'use strict';
+    /**
```
jleft: I'm happy for these comments to be removed, otherwise, can we add the `config.d3TimeInterval` to them?

src/assets/js/model/products.js
```diff
@@ -0,0 +1,49 @@
+(function(d3, fc, sc) {
+    'use strict';
+    /**
+    * @param {Object} config - object providing parameters.
```
jleft: I'm happy for these comments to be removed, otherwise, can we add the `config.volumeFormat` to them?

src/assets/js/model/products.js
```diff
@@ -0,0 +1,49 @@
...
+    */
+    var productFactory = function(config) {
+        return {
+            display: config.display || 'Unspecified Product',
+            priceFormat: d3.format(config.priceFormat || '.2f'),
+            volumeFormat: d3.format(config.volumeFormat || 's'),
+            getPeriods: function() {
+                var self = this;
+                var result = productPeriodAssociation
+                .filter(function(x) {return x.product === self;})
```
jleft: Indent (4 spaces) and add spaces within the braces (same for line below too):

```js
{ return x.product === self; }
```

src/assets/js/chart/primary.js
Afredon: Do you mean drawing the indicators first? In which case the bands would be under everything and that's a change of behavior, so I am not sure if that was expected.

src/assets/js/model/products.js
Afredon: Geez I miss my tools telling me when I forget to update comments ! :)

src/assets/js/data/dataInterface.js
Afredon: Also, it didn't even cross my mind that the basic maths methods would not have been implemented in all browsers since version 1. Thanks for pointing that out !

src/assets/js/chart/primary.js
jleft: I mean draw the indicators in the same order as they are currently drawn and draw the crosshair on top of everything, rather than under the indicators, so we'd set the series of the chart in `updateMultiSeries` to this order:

`multi.series(gridlines, currentSeries.option, closeLine, indicators, crosshair]);`

Afredon: Why not, it works too. I would then replace makeIndicatorsTransparentToMouseEvents by a new function called ensureCrosshairIsDrawnOnTop with a single line: 
multi.series(multi.series().concat(crosshair));

It's a matter of choice of which constraint you want to keep: inactive indicators, or crosshair always on top. I don't have any idea which one will be the easiest to keep in the future, depending on new functionalities added.

src/assets/js/menu/generator/dropdownGroup.js
Afredon: Can't remember why I put that underscore here, but yeah it has nothing to do in that code.

src/assets/js/chart/primary.js
Afredon: done

src/assets/js/chart/primary.js
Afredon: done

src/assets/js/menu/group.js
```diff
@@ -31,6 +36,15 @@
             return group;
         };
 
+        group.options = function(array, adaptor) {
```
jleft: Thinking about this, it might be confusing that there's both an `option` method and an `options` method on `group`...
How about consolidating this into one method, where the `adaptor` is optional?

```js
group.options = function(options, adaptor) {
    if (typeof adaptor === 'undefined') {
        adaptor = fc.util.fn.identity;
    }
    option = [];
    options.forEach(function(optionElement) {
        option.push(adaptor(optionElement));
    });
    return group;
};
```

@djmiley - any thoughts?

src/assets/styles/primary-legend.less
```diff
@@ -0,0 +1,17 @@
+#primary-legend {
```
jleft: Rename file to `primaryLegend.less` for consistency (and update in `style.less`, please.

src/assets/js/menu/group.js
Afredon: The fact the nature of the first parameter differs when we add the second parameter or not seems too dangerous to fuse them in a single method. However, I could remove options to optionCollection for example ?

djmiley: I agree, and I like that solution.

djmiley: I guess an alternative approach is setting an adaptor on the object before you call `.option`

src/assets/js/chart/primary.js
jleft: Yeah, both should work.

My instinct says draw the crosshairs on top, it seems more intuitive to me.

src/assets/js/data/dataInterface.js
jleft: > Also, it didn't even cross my mind that the basic maths methods would not have been implemented in all browsers since version 1. Thanks for pointing that out !

Haha, no worries! Yeah, browser inconsistencies ~~can be~~ are a pain.

src/assets/js/menu/group.js
Afredon: There's another problem too. It cannot be consolidated the way you did, as when we use .option we currently pass a group of parameters relevant to one option, while .options get an array of things to transform into options, each of them having a group of parameters.
So this second type of parameter would require 2 forEach to be processed in one go. And it starts to be confusing

Afredon: I really think we should just go for a better name and add meta comments to the method. What about .collection or .underlyingCollection, or something along these lines ?

Afredon: @djmiley It would technically work, but then it requires to use 2 commands to do the action. And it requires the user to know the internal plumbings of the "class" to use it properly, which is a violation of encapsulation.
Ideally we want a behaviour like "Form the group from a collection"

src/assets/styles/primary-legend.less
Afredon: done. Sorry @djmiley , you already told me too and I missed that one in my last push.

djmiley: No problem :)

src/assets/js/model/periods.js
Afredon: Added the comment. I had too often came accross code more than 10 years old where no one had any idea of what the variables/parameters were intended for to leave code without metacomments :)

src/assets/js/chart/primary.js
Afredon: modified to put the crosshair on top.

src/assets/js/data/dataInterface.js
Afredon: .setHours used instead of rounding intended for granularities which ended up not needing it.

src/assets/js/menu/group.js
jleft: > There's another problem too. It cannot be consolidated the way you did, as when we use .option we currently pass a group of parameters relevant to one option, while .options get an array of things to transform into options, each of them having a group of parameters.

I'm not sure I understand the first point - `.option()` expects a variable number of options (`sc.menu.option`), rather than a group of parameters relevant to one option, right?

For example: https://github.com/ScottLogic/d3fc-showcase/blob/0fc0c89d4b646ff8fa26d1f1714e90efdeed4293/src/assets/js/menu/side.js#L31

Afredon: /**
        * Builds a the group option list from a single collection of entities.
        *
        * @param {Object} collectionOfEntity - collection of the entity which will become the exhaustive list of group options.
        * @param {Function} entityAdaptor - a function taking an entity as a parameter and returning an option.
        * NB: Temporarily, the method can leverage Javascript weak typing by passing an array of Options and an Identity adaptor.
        * This should be removed as soon as every source of options is defined as entities.
        */
        group.formOptionListFromCollection = function(arrayOfEntity, entityAdaptor) {
            var adapted = [];
            var iElement;
            for (iElement = 0; iElement < arrayOfEntity.length; iElement++) {
                adapted.push(entityAdaptor(arrayOfEntity[iElement]));
            }
            return addSingleOption.apply(group.option, adapted);
        } ;

That's what we settled on with Drew. .option becomes private to leave only the behavior of feeding a cohesive group of options to group, and the places where the entities do not yet exist will pass things as an array of options with an identity adaptor.
Im, not sure if that is what you meant initially but that should fit the bill

src/assets/js/model/products.js
Afredon: done

src/assets/js/menu/group.js
jleft: > .option becomes private (renamed to addSingleOption)

But `.option` can add many options, rather than just a single option (unless I'm missing something)? If it's to be renamed to `addSingleOption` I'd expect it's implementation to be changed too?

Even is this is the case, I don't see the need for a private `addSingleOption` method: `addSingleOption` doesn't use the value of `this`, so `addSingleOption.apply(group.option, adapted) === addSingleOption.apply(null, adapted)`. `apply` converts `adapted`, an array, to a list, which is then iterated through in `addSingleOption` and each item in the list is added to the `option` array. Why not just assign the `adapted` array directly to `option`?

I think the `option`/`addSingleOption` method could be removed and a method like this should work:

```js
group.options = function(options, adaptor) {
    option = options.map(function(currentOption) { return adaptor(currentOption); });
    return group;
};
```

Afredon: Ahhh, that's where I was confused, I thought option was adding only a single option !

jleft: Yeah, sorry it wasn't very clear, a better name for the method would have been `options`!

Afredon: The code was right under my nose, but I made a confusion with menu.option and stopped seeing the code. Add the 2 hours sleep night, and I was completely off track here :(

src/assets/styles/primaryLegend.less
```diff
@@ -0,0 +1,17 @@
+#primary-legend {
+    position: absolute;
+    top: 0;
+    left: 0;
+    margin: 10px;
+    background-color: rgba(200, 200, 255, 0.8);
+    font-size: 0.8em;
```
jleft: Prefer `@font-size-small` from `node_modules\bootstrap\less\variables.less`

src/assets/styles/primaryLegend.less
```diff
@@ -0,0 +1,17 @@
...
+    line-height: 1em;
```
jleft: `line-height: 1;`

src/assets/styles/primaryLegend.less
```diff
@@ -0,0 +1,17 @@
...
+    font-family: monospace;
```
jleft: Prefer ` @font-family-monospace` from `node_modules\bootstrap\less\variables.less`

src/assets/styles/variables.less
```diff
@@ -41,3 +41,7 @@
 
 // Menu
 @option-group-padding: 15px;
+
+// Crosshair
+@crosshair-line-color: #808080;
+@crosshair-circle-color: rgba(130, 0, 200, 0.8);
```
jleft: I'd omit the opacity from this, as I think the stroke should be opaque; `fill-opacity` can be added to the `circle`

src/assets/styles/primaryLegend.less
Afredon: I really meant 1em and not x1 though

src/assets/styles/primaryLegend.less
Afredon: A lot of things to put to the variables. On it

src/assets/styles/variables.less
Afredon: Ah nice attribute !

src/assets/styles/primaryLegend.less
djmiley: I think we have access to the underlying bootstrap variables, so we don't need to declare these in `variables.less`

jleft: > I think we have access to the underlying bootstrap variables, so we don't need to declare these in `variables.less`

Yeah, exactly :+1:

Afredon: They're slightly different, but they will do. Thanks for pointing that out too

src/assets/js/app.js
```diff
@@ -120,8 +123,7 @@
                         !container.selectAll('.row-offcanvas-right').classed('active'));
                 });
 
-            container.select('.head-menu')
-                .call(headMenu);
+            renderMenu();
```
djmiley: Call in `app.run()`

src/assets/js/app.js
```diff
@@ -98,21 +103,19 @@
...
-                        model.period = 60 * 60 * 24;
+            headMenu.on('dataProductChange', function(product) {
+                    sc.model.selectedProduct = product.option;
+                    sc.model.selectedPeriod = product.option.getPeriods()[0];
+                    if (product.option === sc.model.product.bitcoin) {
+                        dataInterface(sc.model.selectedPeriod.seconds);
+                    } else if (product.option === sc.model.product.generated) {
+                        dataInterface.generateDailyData(sc.model.selectedPeriod.seconds);
                     }
+                    renderMenu();
```
djmiley: Do we need this here? If so, why not in 'dataPeriodChange`?

src/assets/js/chart/xAxis.js
```diff
@@ -7,14 +7,23 @@
...
-            .orient('bottom')
-            .ticks(6);
+            .orient('bottom');
 
         var dataJoin = fc.util.dataJoin()
             .selector('g.x-axis')
             .element('g')
             .attr('class', 'x-axis');
 
+        function preventTooSmallTimeIntervals(period) {
```
djmiley: Could this function name make it clearer that we are altering the `xAxis.ticks` property?

src/assets/js/chart/xAxis.js
```diff
@@ -7,14 +7,23 @@
...
+            var tickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
+            var minSecondsPerInterval = period.seconds;
+            if (tickSeconds < minSecondsPerInterval) {
```
djmiley: Maybe a ternary?

src/assets/js/app.js
Afredon: Each time we change product, we need to recalculate the period combo box (to fill it with the values for this new product and decide to show or hide it). That's why it is in dataProductChange.
However, when we change the period, there is no impact on the list of available products, so there is no need to rerender.

src/assets/js/app.js
Afredon: I don't get what you mean here

src/assets/js/chart/xAxis.js
Afredon: I was hoping TimeIntervals would help, but any other name is fine by me. 
what about preventTicksMoreFrequentThanPeriod ?

src/assets/js/menu/generator/dropdownGroup.js
```diff
@@ -9,13 +9,19 @@
             .element('option');
 
         function layoutDropdown(sel) {
+            var selectedIndex;
+            if (!!sel.datum() && !!sel.datum().propagated) {
```
djmiley: Would `var selectedIndex = sel.datum().propagated.selectedIndex || 0;` not work?

src/assets/js/menu/head.js
```diff
@@ -14,38 +14,33 @@
...
-        var oneMinutePeriod = sc.menu.option('1 min', '60', 60);
-
         var dataPeriodDropdown = sc.menu.group()
-            .option(hourPeriod, fiveMinutePeriod, oneMinutePeriod)
             .generator(sc.menu.generator.dropdownGroup())
             .on('optionChange', function(period) {
                 dispatch.dataPeriodChange(period);
             });
 
+        function setDropdownContent() {
```
djmiley: More of a `configureMenu`, given the setPeriodChangeVisibility.

src/assets/js/chart/xAxis.js
Afredon: If you want, but I never found ternary was improving readability for long lines. But I don't mind either

src/assets/js/menu/generator/dropdownGroup.js
Afredon: I'll try again, but I think sel.datum().propagated throws an error if datum() is undefined

src/assets/js/menu/head.js
Afredon: I renamed to ```configureDropdown```

src/assets/js/chart/xAxis.js
Afredon: If you meant ```  (tickSeconds < period.seconds) ?
    xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value) :
    xAxis.ticks(6);```, none of the linters I tried accept it.
Anyway the ternary is better used inside an expression, like ```x = condition ? a : b```.  Here it improves readability.

src/assets/js/menu/generator/dropdownGroup.js
Afredon: I confirm:
```caught TypeError: Cannot read property 'selectedIndex' of undefined```

File: src/assets/js/chart/xAxis.js (hunk 18 above)
Comment (djmiley) [inline src/assets/js/chart/xAxis.js]: Ok, yes no, problem.

File: src/assets/js/app.js [hunk 21]
```diff
@@ -165,7 +167,7 @@
 
             updateLayout();
 
-            dataInterface.generateData();
+            dataInterface.generateDailyData(sc.model.selectedPeriod.seconds);
```
djmiley: Don't pass in argument here

src/assets/js/menu/generator/dropdownGroup.js
djmiley: Can we not be calling `sel.datum().selectedIndex` ? The check for the the two levels `!!sel.datum() && !!sel.datum().propagated` doesn't feel right

src/assets/js/app.js
Afredon: Ah yes, I had started to unify the interface of the calls to dataInterface before rolling back, but forgot this argument. Removing !

src/assets/js/menu/group.js
```diff
@@ -9,25 +9,37 @@
...
         function group(selection) {
+            var propagatedDatum = this.datum();
             var optionGenerator = generator.on('optionChange', function(option) {
                     dispatch.optionChange(option);
                 });
 
             selection.each(function() {
+                var compoundDatum = option;
+                if (propagatedDatum !== undefined) {
+                    compoundDatum.propagated = propagatedDatum;
```
djmiley: Can we have this at the same level as option?

src/assets/js/app.js
djmiley: Return `headMenu` from this function and call externally.

src/assets/js/menu/generator/dropdownGroup.js
Afredon: I started initially with just !!sel.datum().propagated but got an error at some point and needed to add a check to the first level.
Otherwise the extra level is because I attach to datum() a propagated object when we're bubbling down from main event handler, and then I have to look into it for the field I want to address in the leafmost component which was the target.

src/assets/js/app.js
Afredon: Ah. It's called in app.run() already through ```initialiseHeadMenu```

src/assets/js/menu/group.js
Afredon: Yes it's possible by wrapping option and propagated together in an object.

src/assets/js/chart/primary.js
```diff
@@ -116,17 +138,26 @@
...
+         * Adds the crosshair component as the last element of the multi.series to ensure it is drawn on top
+         * This is to avoid problems of layers intercepting mouse events
+         * and thus preventing the crosshair to update the legend or snap properly.
+         *
+         * @param {Object} multi - the multi series to complete with a crosshair
+         */
+        function addCrosshairOnTop(multi) {
+            multi.series(multi.series().concat(crosshair));
+        }
+
```
djmiley: Move into `updateMultiSeries()`, it's fairly self-evident what we are doing so we shouldn't need a seperate function.

src/assets/js/chart/xAxis.js
djmiley: It may be nicer to return `tickSeconds < minSecondsPerInterval` in a function such as 'tickValueDifferenceSmallerThanPeriod' and set the `xAxis.ticks()` explicitly

src/assets/js/chart/primary.js
Afredon: It's to keep track explicitely of the rule we decided. A concatenation of a series to an array of series takes more time to understant that addCrosshairOnTop, and when the code grows and gets old, it stays self-evident. While 20 undocument changes to updateMultiSeries later, it might not be.
Also it expresses intent instead of how it is done, which is also easier to read.

src/assets/js/chart/xAxis.js
Afredon: no problem with that. By set ticks explicitly, do you mean the way it is or 2 functions setTicksFixedInterval and setTicksDefault ?

djmiley: I would probably set them explicitly, but it's open for discussion.

src/assets/js/chart/primary.js
djmiley: I think, for now, it's something we don't need so much, given we don't do the same for our `baseChart` variable. Possibly comment within `updateMultiSeries()`

src/assets/js/menu/generator/toggleGroup.js
```diff
@@ -10,7 +10,7 @@
             .attr('class', 'btn btn-default');
 
         function layoutButtons(sel) {
-            dataJoin(sel, sel.datum())
+            dataJoin(sel, sel.datum().option)
```
djmiley: Possible bug.

src/assets/js/chart/primary.js
```diff
@@ -67,14 +67,36 @@
     }
 
     sc.chart.primary = function() {
+        function renderLegend(dataPoint) {
+            d3.select('#primary-legend')
+                .datum(dataPoint)
+                .call(legend);
+        }
```
djmiley: Move out to `sc.chart.legend()`

src/assets/js/menu/generator/toggleGroup.js
Afredon: resolved while reworking the use of datum

src/assets/js/chart/primary.js
Afredon: Inlined

src/assets/js/menu/group.js
Afredon: Done. All datum while descending the chain group/menu chain now use an object with optionList and in the case of dropdownGroup selectedIndex as fields.

src/assets/js/menu/generator/dropdownGroup.js
Afredon: Simplified while reworking datum

src/assets/js/chart/primary.js
Afredon: done

Gruntfile.js
```diff
@@ -14,12 +14,14 @@ module.exports = function(grunt) {
...
             srcJsFiles: [
                 'src/assets/js/sc.js',
+                'src/assets/js/model/**/*.js',
                 'src/assets/js/chart/**/*.js',
                 'src/assets/js/menu/**/*.js',
                 'src/assets/js/util/**/*.js',
                 'src/assets/js/data/**/*.js',
                 'src/assets/js/behavior/**/*.js',
                 'src/assets/js/series/**/*.js',
+                'src/assets/js/pre-flux/**/*.js',
```
djmiley: Remove this namespace. I don't think there are any files here anymore.

src/assets/js/app.js
```diff
@@ -10,10 +10,10 @@
         var svgSecondary = container.selectAll('svg.secondary');
         var svgXAxis = container.select('svg.x-axis');
         var svgNav = container.select('svg.nav');
+        var divLegend = container.select('#primary-legend');
```
djmiley: Rename div to '#legend'

src/assets/js/app.js
```diff
@@ -22,11 +22,15 @@
         var secondaryCharts = [];
         var xAxis = sc.chart.xAxis();
         var nav = sc.chart.nav();
+        var headMenu = sc.menu.head();
+        var legendRenderer = sc.chart.legendRenderer();
```
djmiley: `var legend = sc.chart.legend();`

src/assets/js/app.js
```diff
@@ -42,6 +46,11 @@
                 .call(nav);
         }
 
+        function renderMenu() {
```
jleft: `renderMenu` -> `renderHeadMenu`

src/assets/styles/primaryLegend.less
```diff
@@ -0,0 +1,18 @@
+#primary-legend {
```
djmiley: Similar `legend.less`

src/assets/js/app.js
```diff
@@ -59,7 +68,19 @@
             render();
         }
 
+        function renderLegend() {
+            divLegend
```
jleft: We'd usually go for one method call a line - `.datum()` on same line as `divLegend`:

```js
divLegend.datum(sc.model.legendData)
    .call(legendRenderer);
```

src/assets/js/chart/legend.js
```diff
@@ -0,0 +1,30 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.chart.legendRenderer = function() {
+        var legend = fc.chart.legend();
+
+        function render(selection) {
```
djmiley: this function should be called `function legend(selection)`

src/assets/js/chart/legend.js
```diff
@@ -0,0 +1,30 @@
...
+                    ['open', function(d) { return formatPrice(d.open); }],
+                    ['high', function(d) { return formatPrice(d.high); }],
+                    ['low', function(d) { return formatPrice(d.low); }],
+                    ['close', function(d) { return formatPrice(d.close); }],
+                    ['volume', function(d) { return formatVolume(d.volume); }]
+                ]);
+            return render;
+        };
+
+        return render.configureForOHLCV();
```
djmiley: should be returning `legend`

src/assets/js/chart/legend.js
```diff
@@ -0,0 +1,30 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.chart.legendRenderer = function() {
```
jleft: To be consistent with our other renderers: `sc.chart.legendRenderer` -> `sc.chart.legend`

src/assets/js/app.js
djmiley: Bring data binding up a line

Gruntfile.js
Afredon: Yep, the files were removed in the last push. Removing this line.

src/assets/js/chart/primary.js
```diff
@@ -66,15 +66,23 @@
...
+        var yAxisWidth = 45;
+        var dispatch = d3.dispatch('viewChange', 'crosshairChange');
 
         var currentSeries = sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick());
         var currentYValueAccessor = function(d) { return d.close; };
         var currentIndicators = [];
 
+
+        var crosshairData = [];
+        var crosshair = fc.tool.crosshair()
```
jleft: Can we just do this?

```js
var crosshair = fc.tool.crosshair()
    .xLabel('')
    .yLabel('');
```

src/assets/js/chart/legend.js
djmiley: See other charts in `sc.chart` namespace

src/assets/js/model/userSelections.js
```diff
@@ -0,0 +1,6 @@
+(function(d3, fc, sc) {
+    'use strict';
+    // TODO: move to Instrument store initialisation
```
djmiley: Remove comment?

src/assets/js/menu/group.js
```diff
@@ -5,29 +5,38 @@
 
         var dispatch = d3.dispatch('optionChange');
 
-        var option;
+        var optionList;
         var generator;
 
         function group(selection) {
+            var propagatedDatum = this.datum() || {};
```
jleft: `this.datum()` -> `selection.datum()` 

They're equivalent in this context, but we tend to operate on the selection, rather than `this`.

src/assets/js/app.js
Afredon: Renamed all occurrences.

src/assets/js/app.js
Afredon: I'll rename it if you prefer to have everything called the same way, but that's one of the things that made it harder to read the code for me, calling renderers as the thing they output.

src/assets/js/app.js
Afredon: renamed

src/assets/js/app.js
Afredon: updated

src/assets/js/chart/legend.js
Afredon: The legend is configurable to extend to other types of data to display. Currently it has only one configuration, OHLCV. This is called on the initial return to set up a default configuration.

src/assets/js/chart/legend.js
Afredon: Done, but as mentioned above, the current names are a big source of confusion to me, as they don't represent what  it stored in the variable.

src/assets/js/menu/group.js
jleft: Actually ignore that comment, this can be moved into the `selection.each` below.

```js
selection.each(function() {
    var selection = d3.select(this);
    var options = selection.datum() || {};
    options.optionList = optionList;
    // etc.
```

src/assets/js/chart/primary.js
Afredon: Much better! done

src/assets/js/chart/legend.js
djmiley: Should the default configuration not be internally set. If there was another type of data to display, we would have to change this line, rather than simply calling setters.

Afredon: I don't get what you mean. 
Isn't ```return render.configureForOHLCV();``` private ?

src/assets/js/menu/group.js
Afredon: I changed the this to selection. And I actually moved the assignment outside of the loops as it is invariant.

src/assets/js/chart/legend.js
Afredon: Ah you mean if you need another configuration. Then you would just call like that:
```var legend = fc.chart.legend().configureXXX(); ```

File: src/assets/js/model/userSelections.js (hunk 36 above)
Comment (Afredon) [inline src/assets/js/model/userSelections.js]: done

File: src/assets/js/chart/legend.js (hunk 33 above)
Comment (jleft) [inline src/assets/js/chart/legend.js]: @djmiley - are you thinking of something like this?

```js
sc.chart.legend = function() {

    var formatPrice = function(x) { return sc.model.selectedProduct.priceFormat(x); };
    var formatVolume = function(x) { return sc.model.selectedProduct.volumeFormat(x); };
    var formatTime = function(x) { return sc.model.selectedPeriod.timeFormat(x); };

    var legendChart = fc.chart.legend()
        .items([
            ['date', function(d) { return formatTime(d.date); }],
            ['open', function(d) { return formatPrice(d.open); }],
            ['high', function(d) { return formatPrice(d.high); }],
            ['low', function(d) { return formatPrice(d.low); }],
            ['close', function(d) { return formatPrice(d.close); }],
            ['volume', function(d) { return formatVolume(d.volume); }]
        ]);

    function legend(selection) {
        selection.call(legendChart);
    }

    return legend;
};
```

Then if we want to add the configuration we could expose the `items()` method on `fc.chart.legend`? `d3.rebind(legend, legendChart, 'items');`

jleft:
Incorporated in #294.

PR #295, opened 2015-10-19
First iteration of adding models for renderers/views; pull out events

jleft:


src/assets/js/chart/primary.js
```diff
@@ -154,15 +154,26 @@
...
+
+        function changeYValueAccessor(yValueAccessor) {
+            currentYValueAccessor = yValueAccessor.option;
+        }
+
         function primary(selection) {
             var model = selection.datum();
+            changeSeries(model.series);
+            changeYValueAccessor(model.yValueAccessor);
+            currentIndicators = model.indicators;
```
djmiley: Why extract `changeSeries` as a function but not this?

src/assets/js/app.js
```diff
@@ -62,68 +60,77 @@
...
-            model.viewDomain = [domain[0], domain[1]];
-            model.trackingLatest = sc.util.domain.trackingLatestData(model.viewDomain, model.data);
+        function onViewChange(domain) {
+            primaryChartModel.viewDomain = [domain[0], domain[1]];
+            secondaryChartModel.viewDomain = [domain[0], domain[1]];
+            xAxisModel.viewDomain = [domain[0], domain[1]];
+            navModel.viewDomain = [domain[0], domain[1]];
+            primaryChartModel.trackingLatest = sc.util.domain.trackingLatestData(
+                primaryChartModel.viewDomain,
+                primaryChartModel.data);
```
djmiley: I think all the other charts (except `sc.chart.xAxis`) need to know about `trackingLatest` in order to stick to the latest data when zooming in.

src/assets/js/app.js
```diff
@@ -62,68 +60,77 @@
...
                     if (socketEvent.type === 'error' ||
                         (socketEvent.type === 'close' && socketEvent.code !== 1000)) {
                         console.log('Error loading data from coinbase websocket: ' +
                         socketEvent.type + ' ' + socketEvent.code);
                     } else if (socketEvent.type === 'message') {
                         updateModelData(data);
-                        if (model.trackingLatest) {
-                            var newDomain = sc.util.domain.moveToLatest(model.viewDomain, model.data);
-                            onViewChanged(newDomain);
+                        if (primaryChartModel.trackingLatest) {
```
djmiley: It's probably right that `primaryChartModel` would be our default frame of reference out of all the charts. Is this the intention here?

src/assets/js/app.js
```diff
@@ -62,68 +60,77 @@
...
-        function initialiseChartEventHandlers() {
-            primaryChart.on('crosshairChange', onCrosshairChanged);
-            primaryChart.on('viewChange', onViewChanged);
-            nav.on('viewChange', onViewChanged);
-        }
-
-        function resetToLive() {
-            var data = model.data;
+        function resetToLatest() {
+            var data = primaryChartModel.data;
```
djmiley: See below.

djmiley:
Like the refactoring, looks really good. :) Strikes me that we may want a basic chart model (with `data` and `viewDomain` on, maybe an  basic interactive chart model with `trackingLatest` on in addition) that extends to more specialised implementations.

src/assets/js/app.js
```diff
@@ -131,40 +138,38 @@
...
-                .on('primaryChartYValueAccessorChange', function(yValueAccessor) {
-                    primaryChart.changeYValueAccessor(yValueAccessor);
+                .on(sc.event.primaryChartYValueAccessorChange, function(yValueAccessor) {
+                    primaryChartModel.yValueAccessor = yValueAccessor;
                     render();
                 })
-                .on('primaryChartIndicatorChange', function(toggledIndicator) {
-                    primaryChart.toggleIndicator(toggledIndicator);
+                .on(sc.event.primaryChartIndicatorChange, function(toggledIndicator) {
+                    primaryChartModel.toggleIndicator(toggledIndicator);
```
djmiley: Could pull the `toggleIndicator` code out of `primaryChartModel` (near-exact duplication of lines 168-174), then set the new array of `indicators` on the model.

djmiley: Doesn't feel `toggleIndicator` belongs on the `primaryChartModel`

tlsim: @jleft do you see these more as view models in the MVVM sense?

src/assets/js/chart/primary.js
jleft: Not particular reason, I'll update to be consistent.

src/assets/js/app.js
jleft: Good catch :+1:

src/assets/js/app.js
jleft: > @jleft do you see these more as view models in the MVVM sense?

Yeah, I'm leaning that way at the moment.

djmiley: As discussed in kitchen, I have reconsidered my opinion on this, by considering implementation for indicators consisting of overlays and secondary charts.

jleft:
> Like the refactoring, looks really good. :) Strikes me that we may want a basic chart model (with data and viewDomain on, maybe an basic interactive chart model with trackingLatest on in addition) that extends to more specialised implementations.

Thanks @djmiley :). Yeah, I've had similar thoughts - I can create an issue to track the idea?

jleft:
Incorporated in #297

PR #297, opened 2015-10-19
Centralise events, refactor models for components, create secondary chart component

jleft:
We have two types of renderers:

+ Components (follow the standard component pattern); data is bound to the selection passed to them and configuration is set using methods on the component
+ Views (are composed of components); both data and configuration properties are bound to the selection passed to them, there are no methods for configuration

https://github.com/ScottLogic/d3fc-showcase/commit/b8d21c337a58a8476b2425c7a1c624539471d980 separates the views into their own namespace - is this a more logical/intuitive grouping?

### DIFF
--- a/src/assets/js/chart/volume.js
+++ b/src/assets/js/chart/volume.js
@@ -1,68 +1,31 @@
 (function(d3, fc, sc) {
     'use strict';
 
-    function formatVolume(x) { return sc.model.selectedProduct.volumeFormat(x); }
-
     sc.chart.volume = function() {
-        var yAxisWidth = 45;
-
-        var dispatch = d3.dispatch('viewChange');
-
-        var xScale = fc.scale.dateTime();
-
-        var volumeChart = fc.chart.cartesianChart(xScale, d3.scale.linear())
-            .xTicks(0)
-            .yOrient('right')
-            .yTickFormat(formatVolume)
-            .margin({
-                top: 0,
-                left: 0,
-                bottom: 0,
-                right: yAxisWidth
-            });
-
+        var dispatch = d3.dispatch(sc.event.viewChange);
+        var formatVolume = function(volume) { return sc.model.selectedProduct.volumeFormat(volume); };
         var volumeBar = fc.series.bar()
             .yValue(function(d) { return d.volume; });
-        var multi = fc.series.multi()
-            .series([volumeBar])
-            .mapping(function(series) {
-                return this.data;
-            });
 
-        var createForeground = sc.chart.foreground()
-            .rightMargin(yAxisWidth);
+        var chart = sc.chart.secondary()
+            .series([volumeBar])
+            .yTickFormat(formatVolume)
+            .on(sc.event.viewChange, function(domain) {
+                dispatch[sc.event.viewChange](domain);
+            });
 
         function volume(selection) {
             var model = selection.datum();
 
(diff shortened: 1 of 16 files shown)


djmiley:
:+1:

tlsim:
> We have two types of renderers

Maybe that's the issue. If they only differ in the way state is passed in, I feel we should use components for the charts too.

jleft:
> Maybe that's the issue. If they only differ in the way state is passed in, I feel we should use components for the charts too.

Following the chat earlier, are we happy with the approach of separate 'views' and 'components' (as defined above)? Naming probably isn't the best, so happy to rename if there are better ideas...

I'll also strip any methods from the models (also happy to rename 'models' if there's a better name).

+ [ ] Remove methods from models

tlsim:
With 'views' rename - :+1:

jleft:
+ [ ] Remove methods from models
+ [ ] Rename `view` to `chart`

src/assets/js/chart/volume.js
```diff
@@ -1,68 +1,31 @@
...
-            .yTickFormat(formatVolume)
-            .margin({
-                top: 0,
-                left: 0,
-                bottom: 0,
-                right: yAxisWidth
-            });
-
+        var dispatch = d3.dispatch(sc.event.viewChange);
+        var formatVolume = function(volume) { return sc.model.selectedProduct.volumeFormat(volume); };
```
tlsim: Should we only use `sc.model` at the top level, and retrieve sub-component models from the selection's data?

jleft: Definitely. The use of global models, as shown here, should be removed.

I didn't make that change in this pull request, as I wasn't focusing on the existing models - the periods, products and user selections.

Can create another issue for this (to refactor global models)? Happy to add TODOs (identifying any global usage of models, which should be removed) to this PR, if that would be useful?

tlsim: Another issue makes sense. This PR adds stuff to src/assets/js/model - is the plan to remove these and create the models at the top level? No need to add TODOs I think.

jleft: > This PR adds stuff to src/assets/js/model - is the plan to remove these and create the models at the top level?

As the 'models' in src/assets/js/model as just factories for the models, rather than the actual instances of the models, I don't think there is a problem with this approach? I thought it could be nice to separate them, as it would reduce the clutter in app.js and hopefully make it a but easier to find the structure of a model for a specific component.

In practice, the difference in implementation in app.js could look like this:

```js
var navModel = sc.model.nav();
```
vs.
```js
var navModel = {
    data: [],
    viewDomain: [],
    trackingLatest: true
};
```

There's not much difference between either way, so I'm happy to go with either approach, as long as we're consistent.

jleft: #304

tlsim: This makes sense. The global models had me confused! We'll need to be consistent in the usage and naming for factories. Currently period and product take a config object, and have 'Factory' in the name. I quite like the simpler type introduced here.

jleft: > This makes sense. The global models had me confused!

Yup, me too!

> We'll need to be consistent in the usage and naming for factories. Currently period and product take a config object, and have 'Factory' in the name. I quite like the simpler type introduced here.

Agreed.

PR #299, opened 2015-10-20
Fixes potential bug in close price line calculation

djmiley:
Removes use of `data[data.length - 1]` to calculate latest data in close price line & annotation and for default crosshair display

### DIFF
--- a/test/util/latestDataSpec.js
+++ b/test/util/latestDataSpec.js
@@ -23,43 +23,18 @@
             shuffledData = [obj(thursday), obj(wednesday), obj(monday), obj(friday), obj(tuesday)];
         });
 
-        it('should return a single latest data point in an array if it is unique', function() {
+        it('should return a single latest data point in an array', function() {
             var latestDatum = sc.util.latestData(data);
 
-            expect(latestDatum.length).toBe(1);
-            expect(latestDatum[0].date).toEqual(friday);
+            expect(latestDatum.date).toEqual(friday);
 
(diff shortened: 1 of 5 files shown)


test/util/latestDataSpec.js
```diff
@@ -0,0 +1,66 @@
...
+            expect(latestDatum.length).toBe(1);
+            expect(latestDatum[0].date).toEqual(friday);
+
+            var shuffledLatestDatum = sc.util.latestData(shuffledData);
+
+            expect(shuffledLatestDatum.length).toBe(1);
+            expect(shuffledLatestDatum[0].date).toEqual(friday);
+        });
+
+        it('should return the multiple latest data points in an array if there are more than one', function() {
```
jleft: I'm not sure about this... I think the chart would break if we were to have > 1 datum with the same date/time in the dataset.

Are we ever expecting more than one datum on the same date? If this were to happen then I feel it should be the responsibility of the dataInterface/data components to consolidate these into a single datum.

If we were ever to find > 1 datum with the same date/time anywhere else in the application (e.g. in app, primary chart etc.), then I expect an error to be thrown.

djmiley: Happy to change it to assuming that?

Afredon: What happens during the change to Daylight Saving Time on 24/24 markets ? Do we have a duplicate at 2:35 for example for bitcoin rates ?

djmiley:
Probably not needed so closing this.

PR #302, opened 2015-10-22
WIP - refactor menu

tlsim:
@jleft I think I'm getting the idea after yesterday's chat. Am I on the right track (with the structure and the TODO items) ? 

Also removed sc.menu.group - it's a wrapper around array.map and selection.datum. Calling the generator components directly seems simpler.


jleft:
:+1: Looks really good to me

src/assets/js/menu/side.js
```diff
@@ -44,34 +44,41 @@
...
-            .generator(sc.menu.generator.toggleGroup())
+        var secondaryChartToggle = sc.menu.generator.toggleGroup()
             .on('optionChange', function(chart) {
                 dispatch.secondaryChartChange(chart);
             });
 
         var side = function(selection) {
-            selection.each(function() {
+            // TODO: var model = selection.datum();
+            selection.each(function() { // TODO: Do we really need to support multiple selection here ?
```
tlsim: Low-barrel example *does* use selection.each(), with model available as first argument to the function.

jleft: I'm not too fussed either way, but for consistency we could use `selection.each()` with all of our components?

I also think the latter of these examples (below) is a bit nicer:

```js
// ...
```

```js
selection.each(function(model) {
    // ...
});
```

tlsim: I agree. Let's use the second example and be consistent in using `selection.each()`.

jleft: #303

PR #306, opened 2015-10-23
Some optimisations (rAF, event filtering, flexbox layout)

Afredon:
Ready for inspection. Next step is to include the head in the flexbox, which will allow to move the whole thing to flex and to remove the height calculation still left in layout.js
Performances on simple chart are now ok, but still not enough once we add secondary charts and indicators.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -25,22 +25,13 @@
         var legend = sc.chart.legend();
 
         function renderInternal() {
-<<<<<<< HEAD
-            svgPrimary.datum(primaryChartModel)
-=======
-
-            primaryChartContainer.datum(model)
->>>>>>> Post code review except LESS issues.
+            primaryChartContainer.datum(primaryChartModel)
                 .call(primaryChart);
 
             legendContainer.datum(sc.model.legendData)
                 .call(legend);
 
-<<<<<<< HEAD
-            svgSecondary.datum(secondaryChartModel)
-=======
-            secondaryChartsContainer.datum(model)
->>>>>>> Post code review except LESS issues.
+            secondaryChartsContainer.datum(secondaryChartModel)
                 .filter(function(d, i) { return i < secondaryCharts.length; })
                 .each(function(d, i) {
                     d3.select(this)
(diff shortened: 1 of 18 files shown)


djmiley:
Apologies for not using commenting correctly on the PR.

src/assets/js/util/layout.js
```diff
@@ -2,34 +2,27 @@
...
+
+        var headRowHeight = parseInt(container.select('.head-row').style('height'), 10) +
+            parseInt(container.select('.head-row').style('padding-top'), 10) +
+            parseInt(container.select('.head-row').style('padding-bottom'), 10);
+
+        var useableScreenHeight = window.innerHeight - headRowHeight;
+
+        var charts = container.select('#charts-container');
+        charts.style('height', useableScreenHeight);
+        charts.layout();
```
djmiley: Could chain all 3?

Afredon:
I didn't know there was an incorrect way to comment :)

src/assets/styles/xAxis.less
```diff
@@ -1,9 +0,0 @@
-.x-axis-row {
-
-  height: @x-axis-height;
-
-  .chart {
-    padding-right: @y-axis-width;
-  }
-
```
djmiley: Is it possible to not use `yAxisWidth` and use this instead?

src/assets/js/util/layout.js
Afredon: Ah yes I could, but anyway I inted to get rid of the 2 first lines as soon as possible.

djmiley: Ah ok

src/assets/js/app.js
```diff
@@ -25,27 +25,39 @@
...
+            xAxisContainer.datum(model)
                 .call(xAxis);
 
-            svgNav.datum(model)
+            navbarContainer.datum(model)
                 .call(nav);
         }
 
+        var requestId = null;
+        function render() {
```
jleft: I think this is in d3fc now? `fc.util.render()` - can we use that?

djmiley:
The comments won't collapse when they are resolved because I commented on the individual commits. I can delete the comments that are definitely resolved, if you like?

src/assets/js/app.js
Afredon: Done, the code comes from d3fc before it was pushed in the newest version, so it's simpler indeed. But it is not documented yet though.

src/assets/styles/xAxis.less
Afredon: It could be possible if we add the value read from the css to the model and then pass it to the 5 charts that are using this value. I added a use to xAxis, but the same constant is also in Primary and all Secondary charts, which creates duplication and make it hard to change our mind about the width of this axis.

djmiley: Yes, I agree. Probably best to have a layout object. `{
                top: 0,
                left: 0,
                bottom: 0,
                right: yAxisWidth
            }`

src/assets/js/util/layout.js
Afredon: Fixed anyway, what I intend to do and what I will have time to actually do being quite different.

src/assets/styles/xAxis.less
Afredon: I suggest to add that to the issues for now and rework it once we have the final layout. That might solves some issues by removing the need to set the layout in code.

jleft:
Thanks @Afredon :+1:

PR #309, opened 2015-10-26
Refactor side menu

### LINKED ISSUE #301: Button group components are not idempotent
buttonsbug

Screenshot - calling app.run() twice.
This is probably due to calling `insert` on the update selections returned by `fc.util.dataJoin` #271 

Rest of app seems fine!

tlsim:
* Fix #301
* Add side menu model
* optionList -> options
* In buttonGroup, use Bootstrap for styling only - remove data-toggle
* Manage toggled items in model, allowing for use of buttonGroup for toggles. Remove toggleGroup.
* Add side menu to `render`


src/assets/js/app.js
```diff
@@ -32,6 +34,7 @@
                 .call(legend);
 
             svgSecondary.datum(secondaryChartModel)
+                // TODO: Add component: group of secondary charts.
```
jleft: :+1:

src/assets/js/app.js
```diff
@@ -157,45 +164,46 @@
...
+                .on(sc.event.secondaryChartChange, function(chart) {
+                    chart.isSelected = !chart.isSelected;
+                    secondaryCharts = sideMenuModel.secondaryChartOptions.filter(function(option) {
+                        return option.isSelected;
+                    });
+                    // TODO: This doesn't seem to be a concern of menu.
+                    secondaryCharts.forEach(function(chartOption) {
+                        chartOption.option.on(sc.event.viewChange, onViewChange);
+                    });
+                    // TODO: Remove .remove! (could a secondary chart group component manage this?).
```
jleft: :+1:

src/assets/js/menu/option.js
```diff
@@ -1,10 +1,13 @@
 (function(d3, fc, sc) {
     'use strict';
+
+    // TODO: Move to /model
     sc.menu.option = function(displayString, valueString, option) {
         return {
-            displayString: displayString,
-            valueString: valueString,
-            option: option
+            displayString: displayString, // TODO: is 'displayName' better?
```
jleft: I prefer `displayName` or maybe `name`?

src/assets/js/menu/option.js
```diff
@@ -1,10 +1,13 @@
...
+            valueString: valueString, // TODO: is this an id?
```
jleft: I think this is an id - I've been thinking for a while that a GUID component would be useful when we need to assign a unique id something. For example, when keying on multi.

src/assets/js/model/menu/seriesOptions.js
```diff
@@ -0,0 +1,22 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.model.menu.seriesOptions = function() {
+
+        // TODO: Could 'isLine' go on primary chart series model instead?
```
jleft: This is my view on `isLine` (TLDR: we should add an ID property to the series, and remove `isLine`): https://github.com/ScottLogic/d3fc-showcase/pull/154#issuecomment-137704112

This implementation is good for now though :-)

src/assets/js/model/menu/side.js
```diff
@@ -0,0 +1,15 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.model.menu.side = function() {
+
+        // TODO: Instantiate series/indicator components outside of menu model?
```
jleft: We may want to in the future (for example, we may want to pass the menu and primary chart the same instance of the initial series), but if there's no need for it now, I'm happy with this approach.

jleft:
Thanks, @tlsim. This looks really good to me, I'd be happy to see this merged :+1:

tlsim:
Thanks for taking a look @jleft! I mainly added the TODOs as ideas for discussion. Looks we can make tasks from a couple in planning tomorrow.

src/assets/js/menu/option.js
Afredon: It's the string that is presented to the user, so display would be enough if you want to shorten it, but name is misleading.

src/assets/js/menu/option.js
Afredon: It's the string that is presented to the user, so display would be enough if you want to shorten it, but name is misleading. Label, userDisplay, text are possible alternatives

src/assets/js/menu/option.js
Afredon: Ultimately it should disappear entirely, as the option contains the id once the entities are defined properly.

tlsim: Sure, 'text' or 'label' is fine. My objection is with the 'string' part.

src/assets/js/menu/option.js
Afredon: Also, in C# I would strongly support removing "string" as you don't put types in variable names, but in a weak typed language like Javascript, I tend to prefer the hungarian notation.

Afredon: Weirdly enough, this kept in my head for one hour and I get why I would keep displayString here: string is not an indication of the storage type but of what it is, a string on the screen. The same way that we say headModel to talk about a model for the head menu. It could be stored as an array of char, its role is still to end up as a string. Now my english is limited, so a synonym might be better to avoid confusion with type, but display alone is confusing (as it could be a verb), name too as it is not what it is, displayName doesn't really make sense to me, so label or text are definitely ma favourite short variable names.

jleft:
LGTM :+1:

PR #310, opened 2015-10-26
Pull models out of global scope and upgrade to d3fc 3.0.0

### LINKED ISSUE #304: Remove usage of global models
Model objects should not be created in global scope (such as periods, products and user selections), but should be created at the top level - app.js - then passed into the components below, which can then retrieve sub-component models from the model passed into the components (bound to the selection's data).

### LINKED ISSUE #307: Update d3fc dependency to 3.x.x

jleft:
~~Definitely still a WIP, but I thought this was worth showing as (I think) it works and follows the desired pattern (models not in global scope, but declared and mutated in app.js and passed down to components by binding the model to the selection).~~

Any thoughts welcome :-)

Resolves #304. Resolves #307.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -12,11 +12,45 @@
         var navbarContainer = container.select('#navbar-container');
         var legendContainer = container.select('#legend');
 
-        var primaryChartModel = sc.model.primaryChart();
-        var secondaryChartModel = sc.model.secondaryChart();
+        var day1 = sc.model.period({
+            display: '1 Day',
+            seconds: 86400,
+            d3TimeInterval: {unit: d3.time.day, value: 1},
+            timeFormat: '%b-%d'});
+        var hour1 = sc.model.period({
+            display: '1 Hour',
+            seconds: 3600,
+            d3TimeInterval: {unit: d3.time.hour, value: 1},
+            timeFormat: '%b-%d %Hh'});
+        var minute5 = sc.model.period({
+            display: '5 Minutes',
+            seconds: 300,
+            d3TimeInterval: {unit: d3.time.minute, value: 5},
+            timeFormat: '%H:%M'});
+        var minute1 = sc.model.period({
+            display: '1 Minute',
+            seconds: 60,
+            d3TimeInterval: {unit: d3.time.minute, value: 1},
+            timeFormat: '%H:%M'});
+
+        var generated = sc.model.product({
+            display: 'Data Generator',
+            volumeFormat: '.3s',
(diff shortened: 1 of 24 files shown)


src/assets/js/app.js
```diff
@@ -137,18 +189,18 @@
...
-                        dataInterface(sc.model.selectedPeriod.seconds);
-                    } else if (product.option === sc.model.product.generated) {
+                    updateModelSelectedProduct(product.option);
+                    updateModelSelectedPeriod(product.option.periods[0]);
+                    if (product.option === bitcoin) {
+                        dataInterface(product.option.periods[0].seconds);
+                    } else if (product.option === generated) {
                         dataInterface.generateDailyData();
                     }
                     render();
```
djmiley: Can we remove this call to `render`?

src/assets/js/chart/legend.js
```diff
@@ -16,11 +15,18 @@
                 ['volume', function(d) { return formatVolume(d.volume); }]
             ]);
 
+
```
djmiley: Line

src/assets/js/chart/legend.js
```diff
@@ -16,11 +15,18 @@
...
         function legend(selection) {
-            if (!selection.datum()) {
-                selection.datum(sc.model.latestDataPoint);
-            }
-            selection.call(legendComponent);
+            selection.each(function(model) {
+                var container = d3.select(this);
```
djmiley: What is the reason you're not just calling it on the 'selection'? Or why is it not implemented in the other charts?

djmiley: Sorry,forget this.

djmiley:
Yes, looking good at the moment.

src/assets/js/app.js
jleft: We could do, as render would be called via a change in dataInterface, but I think explicitly calling render is clearer.

tlsim: We're calling `renderInternal` in requestAnimationFrame via `fc.util.render`. Doesn't that make any extra `renderInternal` calls redundant?

jleft: @tlsim - I agree, but I'm not fully sure I understand the question?

tlsim: No problem, I should have looked at https://github.com/ScottLogic/d3fc/blob/master/src/util/render.js first.

tlsim:
LGTM :+1:

jleft:
Thanks @djmiley and @tlsim for the review :)

Issue #308, opened 2015-10-26
Period selector displays '1 Hour' for every selection


tlsim:
Regression at c56caaa
Changing period using period selector successfully sets the chart's period, but the selector itself always displays '1 Hour'. Expected the selector to show the updated period when '5 minutes' or '1 minute' is selected.

Afredon:
I can take a look once I'm done with my rebase if you want

djmiley:
It's because it never worked in the first place, I think, and `selectedIndex` would always be -1. Before `sc.event.dataPeriodChange` does not cause the head menu to rerender, whereas now it does. I solved it by keying on the option's valueString.

djmiley:
See https://github.com/djmiley/d3fc-showcase/blob/3a88ef66e0b3d48e01c12c8f03982078f7b235b3/src/assets/js/menu/generator/dropdownGroup.js for a fix

Afredon:
I modified this part in my latest accepted push, and changed the way selectedIndex is passed to the final comboBox through an extension of datum, is that not the version you started from ?

djmiley:
My suspicion is that `selectedIndex: sc.model.products.indexOf(sc.model.selectedProduct)` set `selectedIndex` to -1

tlsim:
> changed the way selectedIndex is passed to the final comboBox through an extension of datum

This makes sense - selectedIndex will need to be a model property so that subsequent `render` calls render the dropdown with the correct item displayed.

Afredon:
selectedProduct was the model property, and the code @djmiley highlighted is where the selected index was deduced from the model property, There shouldn't be a need to store the index itself.

tlsim:
Ok great - I hadn't seen `selectedProduct`.

PR #340, opened 2015-10-27
Fix issue #308 (Period Selector)

Afredon:
Just something that went wrong during a refactoring

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -150,6 +150,7 @@
                 .on(sc.event.dataPeriodChange, function(period) {
                     sc.model.selectedPeriod = period.option;
                     dataInterface(sc.model.selectedPeriod.seconds);
+                    render();
                 })
                 .on(sc.event.resetToLatest, resetToLatest)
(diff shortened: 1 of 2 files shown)


src/assets/js/app.js
```diff
@@ -150,6 +150,7 @@
                 .on(sc.event.dataPeriodChange, function(period) {
                     sc.model.selectedPeriod = period.option;
                     dataInterface(sc.model.selectedPeriod.seconds);
+                    render();
```
jleft: Do we need to render here?

jleft:
Looks good to me, thanks @Afredon :+1:

src/assets/js/app.js
Afredon: as explained above, we don't as dataInterface will trigger rendering through the event chain (twice I think), but it feels like side effect at the moment as we call render() on the other event (dataProductChange).

Afredon: I expect once refactorings are done we will be in position to decide exactly when render() is called.

Issue #342, opened 2015-10-28
Layout not working in Firefox and IE


jleft:
![ff-layout](https://cloud.githubusercontent.com/assets/2017615/10785042/5b0ab91e-7d59-11e5-85f8-74637e8ae34a.PNG)

jleft:
Fixed in #348

PR #345, opened 2015-10-28
Fixes main area height issue on several browsers.

Afredon:
#342 Layout not working on Firefox and IE
-----------------------------------------------
Firefox requires a unit to be provided.
I did a global search on .style('height') and .style('width') and found no other occurences. However, most of them could be in .attr('height') which I didn't look for until we spot an issue.


jleft:
:+1:

The branch is out of sync with `develop`, but once rebased this is good to merge.

jleft:
Rebased and merged in #348.

PR #350, opened 2015-10-29
Legend displayed inline

### LINKED ISSUE #319: Legend

jdhodges431:
Closes #319.

### DIFF
--- a/src/assets/js/chart/legend.js
+++ b/src/assets/js/chart/legend.js
@@ -7,6 +7,21 @@
         var formatTime;
         var lastDataPointDisplayed;
 
+        var legendItems =  [
+            'T',
+            function(d) { return formatTime(d.date); },
+            'O',
+            function(d) { return formatPrice(d.open); },
+            'H',
+            function(d) { return formatPrice(d.high); },
+            'L',
+            function(d) { return formatPrice(d.low); },
+            'C',
+            function(d) { return formatPrice(d.close); },
+            'V',
+            function(d) { return formatVolume(d.volume); }
+        ];
+
         function legend(selection) {
             selection.each(function(model) {
                 var container = d3.select(this);
@@ -17,47 +32,17 @@
 
                 if (model.data !== lastDataPointDisplayed) {
                     lastDataPointDisplayed = model.data;
-                    container.datum(model.data);
+
+                    var span = container.select('p')
+                        .selectAll('span')
+                        .data(legendItems);
+
+                    span.enter()
+                        .append('span')
+                        .attr('class', function(d, i) { return i % 2 === 0 ? 'legendLabel' : 'legendValue'; });
+
+                    span.text(function(d, i) { return i % 2 === 0 ? d : d(model.data); });
                 }
-
-                var d = container.datum();
-
-                var legendComponents =  [
-                    'T',
-                    formatTime(d.date),
-                    'O',
-                    formatPrice(d.open),
-                    'H',
-                    formatPrice(d.high),
-                    'L',
-                    formatPrice(d.low),
-                    'C',
-                    formatPrice(d.close),
-                    'V',
-                    formatVolume(d.volume)
-                ];
-
-                var span = container.select('p')
-                    .selectAll('span')
-                    .data(legendComponents);
-
-                span.text(function(d) {
-                    return d;
-                });
-
-                span.enter()
-                    .append('span')
-                    .text(function(d) {
-                        return d;
-                    });
-
-                span.each(function(d, i) {
-                    if ((i + 1) % 2 === 0) {
-                        this.setAttribute('class', 'legendValue');
-                    } else {
-                        this.setAttribute('class', 'legendLabel');
-                    }
-                });
             });
         }
 
--- a/src/assets/js/chart/legend.js
+++ b/src/assets/js/chart/legend.js
@@ -33,9 +33,13 @@
                 if (model.data !== lastDataPointDisplayed) {
                     lastDataPointDisplayed = model.data;
 
-                    var span = container.select('p')
-                        .selectAll('span')
-                        .data(legendItems);
+                    var p = container.selectAll('p')
+                        .data([legendItems]);
+
+                    p.enter().append('p');
+
+                    var span = p.selectAll('span')
+                        .data(function(d) { return d; });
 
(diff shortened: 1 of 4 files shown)


src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
 
                 if (model.data !== lastDataPointDisplayed) {
```
jleft: If this evaluates to `true`, then we should render, if `false`, then we shouldn't.

This is a performance optimisation to ensure we don't render if the data point passed to this component is the same as the previous time it was called (as nothing would change, the render is unnecessary), therefore all the rendering code below should only be called when this evaluates to `true`.

src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
...
                     lastDataPointDisplayed = model.data;
-                    container.datum(model.data)
-                        .call(legendComponent);
+                    container.datum(model.data);
```
jleft: I don't think we need to do this?

src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
...
                 }
+
+                var d = container.datum();
```
jleft: Here, `model === container.datum()` - we can just use`model`, no need to use `container.datum()`

jleft: Here, `model === container.datum()`, or, if `model.data !== lastDataPointDisplayed`, then `model.data === container.datum()` - is this the intention?

Generally we can just use`model`, no need to use `container.datum()`.

src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
...
+
+                var legendComponents =  [
```
jleft: I'd move this outside of the render function (where `legendComponent` was previously declared), as we don't need to redeclare on every render.

Then an array of data could be created by calling the functions in `legendComponents` with the data on the model.

_Components_ is a loaded term when talking about D3/d3fc. In their simplest form, a component is a function which performs some transformation to the DOM. I'd suggest renaming this - maybe `items` or `legendItems` would work?

src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
...
+                    .selectAll('span')
+                    .data(legendComponents);
+
+                span.text(function(d) {
+                    return d;
+                });
+
+                span.enter()
+                    .append('span')
+                    .text(function(d) {
```
jleft: The `text` on the `spans` could be set twice, if we enter, before updating, then we can ensure we only update the text for each `span` once per render.

```js
span.enter()
    .append('span');

span.text(function(d) { return d; }); // span.text(String); might work too
```

src/assets/js/chart/legend.js
```diff
@@ -26,9 +17,47 @@
...
+                        return d;
+                    });
+
+                span.each(function(d, i) {
```
jleft: Generally, we have a preference for chaining methods and avoid using `.each()` when setting properties on a selection, unless it necessary. I'd do something like this:

```js
.attr('class', function(d, i) { return i % 2 === 0 ? 'legendLabel' : 'legendValue'; });
```

src/assets/js/chart/legend.js
Afredon: I don't get what you mean here. It feels the opposite as span.enter() is restricted to new values, while span after that can contain enter, update and exit, can't it ?

src/assets/js/chart/legend.js
Afredon: values will change much more often than labels, so it feels weird to set the classes at the same place.

jleft: Would probably make more sense on `enter()`.

src/assets/js/chart/legend.js
jleft: > I don't get what you mean here. It feels the opposite as span.enter() is restricted to new values, while span after that can contain enter, update and exit, can't it ?

You're right - my mistake, it's been a long day...

I'd still put enter before update though, one fewer call to `text()` looks neater to me.

Afredon: Ah yes, I did not see the update before, so yes your code suggestion is good. Just the explanation that was weird :)

http://bl.ocks.org/mbostock/3808218

src/assets/styles/legend.less
```diff
@@ -1,19 +1,27 @@
...
   line-height: 1em;
   font-family: @font-family;
   font-size: @legend-font-size;
-  th {
-    padding-right: 5px;
-    text-align: center;
-    font-weight: normal;
+
+  .legendLabel, .legendValue {
+    float: left;
```
jleft: Could we use `display: inline-block` instead?

src/assets/js/chart/legend.js
```diff
@@ -26,8 +32,16 @@
...
+
+                    var span = container.select('p')
+                        .selectAll('span')
+                        .data(legendItems);
+
+                    span.enter()
+                        .append('span')
+                        .attr('class', function(d, i) { return i % 2 === 0 ? 'legendLabel' : 'legendValue'; });
+
+                    span.text(function(d, i) { return i % 2 === 0 ? d : d(model.data); });
```
jleft: This looks quite neat to me:

```js
span.text(function(d) { return d3.functor(d)(model.data); });
```

jleft:
Thanks @jdhodges431 - just a few minor comments, then this looks good to me :+1:

src/assets/js/chart/legend.js
```diff
@@ -26,8 +32,16 @@
 
                 if (model.data !== lastDataPointDisplayed) {
                     lastDataPointDisplayed = model.data;
-                    container.datum(model.data)
-                        .call(legendComponent);
+
+                    var span = container.select('p')
```
tlsim: Might be better to select by class name here instead instead of 'p' - it wasn't immediately obvious to me what the empty paragraph in index.html was for. You could create this element with d3 by data binding `[legendItems]` - appending to the enter selection would then create a single element.

tlsim:
Looks good @jdhodges431 :+1:

jleft:
Rebased and merged in #352.

PR #353, opened 2015-10-30
Implements padding of domain for all series

### LINKED ISSUE #89: Add padding to extent of data domain

djmiley:
Resolves #89

### DIFF
--- a/src/assets/js/model/nav.js
+++ b/src/assets/js/model/nav.js
@@ -5,6 +5,7 @@
         return {
             data: [],
             viewDomain: [],
+            padding: 24 * 60 * 60 / 2,
             trackingLatest: true
         };
     };
--- a/src/assets/js/model/nav.js
+++ b/src/assets/js/model/nav.js
@@ -1,11 +1,11 @@
 (function(d3, fc, sc) {
     'use strict';
 
-    sc.model.nav = function() {
+    sc.model.nav = function(initialPeriod) {
         return {
             data: [],
             viewDomain: [],
-            padding: 24 * 60 * 60 / 2,
+            padding: initialPeriod.seconds / 2,
             trackingLatest: true
         };
     };
--- a/src/assets/js/model/primaryChart.js
+++ b/src/assets/js/model/primaryChart.js
@@ -1,13 +1,13 @@
 (function(d3, fc, sc) {
     'use strict';
 
-    sc.model.primaryChart = function(initialProduct) {
+    sc.model.primaryChart = function(initialProduct, initialPeriod) {
 
         var model = {
             data: [],
             trackingLatest: true,
             viewDomain: [],
-            padding: 24 * 60 * 60 / 2,
+            padding: initialPeriod.seconds / 2,
             selectorsChanged: true
         };
 
--- a/src/assets/js/model/primaryChart.js
+++ b/src/assets/js/model/primaryChart.js
@@ -20,10 +20,7 @@
             }
         });
 
-        var candlestick = sc.series.candlestick();
-        candlestick.isPadded = true;
-
-        var _series = sc.menu.option('Candlestick', 'candlestick', candlestick);
+        var _series = sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick());
         Object.defineProperty(model, 'series', {
             get: function() { return _series; },
             set: function(newValue) {
--- a/src/assets/js/model/primaryChart.js
+++ b/src/assets/js/model/primaryChart.js
@@ -2,10 +2,12 @@
     'use strict';
 
     sc.model.primaryChart = function(initialProduct) {
+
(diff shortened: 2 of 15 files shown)


djmiley:
Fairly rough copy but it's working. Have a look and tell me what you think.

src/assets/js/chart/xAxis.js
```diff
@@ -38,8 +38,9 @@
 
                 container.layout();
 
+                var paddedDomain = model.viewDomain;
                 xScale.range([0, xAxisContainer.layout('width')])
-                    .domain(model.viewDomain);
+                    .domain(paddedDomain);
```
djmiley: Remove changes

src/assets/js/model/primaryChart.js
```diff
@@ -2,11 +2,15 @@
...
     sc.model.primaryChart = function(initialProduct) {
+        var candlestick = sc.series.candlestick();
+        candlestick.isPadded = true;
+
         return {
             data: [],
             trackingLatest: true,
             viewDomain: [],
-            series: sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick()),
+            padding: 24 * 60 * 60 / 2,
```
djmiley: Get initial padding from series setup

src/assets/js/model/nav.js
```diff
@@ -5,6 +5,7 @@
         return {
             data: [],
             viewDomain: [],
+            padding: 24 * 60 * 60 / 2,
```
djmiley: Get initial padding from series setup

src/assets/js/model/secondaryChart.js
```diff
@@ -5,6 +5,7 @@
         return {
             data: [],
             viewDomain: [],
+            padding: 24 * 60 * 60 / 2,
```
djmiley: Get initial padding from series setup

src/assets/js/model/xAxis.js
```diff
@@ -4,6 +4,7 @@
     sc.model.xAxis = function(initialPeriod) {
         return {
             viewDomain: [],
+            padding: 24 * 60 * 60 / 2,
```
djmiley: Get initial padding from series setup

src/assets/js/util/domain/padTimeExtent.js
```diff
@@ -0,0 +1,29 @@
...
+        if (domainBeforeStartOfData(domain, data)) {
+            paddedDomain[0] = d3.time.second.offset(new Date(domain[0]), -padding);
+        }
+        if (domainAfterEndOfData(domain, data)) {
+            paddedDomain[1] = d3.time.second.offset(new Date(domain[1]), +padding);
+        }
+
+        return paddedDomain;
+    };
+})(d3, fc, sc);
```
djmiley: Add tests

djmiley: Look into `fc.util.extent()` rather than `sc.util.domain.padTimeExtent()`

djmiley:
Good points, the padding does need to alter the available domain, rather than just pad the neds in my opinion, as this allows us to fully interact with it through smooth panning and zooming.

djmiley:
We need to talk about how this should work conceptually.

Possible approaches include: this, using element width and converting to time (or worse), padding all series or snapping to amounts of data points.

djmiley:
I've also forgotten about the Volume and MACD charts, which possibly suggest padding all case?

djmiley:
There's a bug. I know what is causing it but it'll take a bit of work. I'm going to go with padding for everything.

djmiley:
Still a bit of work to do, but I think it is on the right track.

src/assets/js/util/domain/padTimeExtent.js
djmiley: Can replace this with the proportional part of fc.util.extent();

src/assets/js/model/nav.js
djmiley: Need to look at how we pass padded and unpadded domains, is there a more elegant way?

src/assets/js/util/domain/padTimeExtent.js
djmiley: Probably not.

djmiley: Going to hopefully do some work on absolute padding on `fc.util.extent()` in d3fc, which should remove the need for this function. https://github.com/ScottLogic/d3fc/issues/794

djmiley:
If I can submit this https://github.com/ScottLogic/d3fc/pull/803, then this PR will be simplified.

djmiley:
Superseded by #484

PR #359, opened 2015-11-03
Product dropdown styling complete

### LINKED ISSUE #358: Product selector

rjmcneill:
Closes #358

### DIFF
--- a/src/assets/styles/dropdown.less
+++ b/src/assets/styles/dropdown.less
@@ -2,8 +2,8 @@
   .dropdown-toggle {
     border: none;
     background-image: none;
-    background-color: @background-color !important;
-    outline: none !important;
+    background-color: @background-color;
+    outline: none;
     box-shadow: none;
     color: @offwhite;
     font-size: @product-font-size;
@@ -18,23 +18,27 @@
     background-color: @dropdown-color;
     text-align: center;
     cursor: pointer;
-    
-    a {
-      color: @offwhite !important;
+
+    > li > a {
+      color: @offwhite;
       font-size: @product-dropdown-font-size;
-    }
-    a:hover {
-      background: none;
+
+      &:hover,
+      &:focus {
+        background: none;
+      }
     }
 
-    li {
+    > li {
       padding-left: 20px;
       padding-right: 20px;
       padding-top: 8px;
       padding-bottom: 8px;
-    }
-    li:hover {
-      background-color: @dropdown-hover !important;
+
+      &:hover,
+      &:focus {
+        background-color: @dropdown-hover;
+      }
     }
   }
 
--- a/src/assets/styles/dropdown.less
+++ b/src/assets/styles/dropdown.less
@@ -1,0 +1,47 @@
+.dropdown {
+  .dropdown-toggle {
+    border: none;
+    background-image: none;
+    background-color: @background-color !important;
+    outline: none !important;
+    box-shadow: none;
+    color: @offwhite;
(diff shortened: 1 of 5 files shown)


src/assets/styles/variables.less
```diff
@@ -63,11 +70,10 @@
 @crosshair-circle-color: @primary-blue;
 @crosshair-circle-opacity: 0.5;
 
-// Legend
+// legend-color
 @legend-background-color: @background-color;
 @legend-font-color: @offwhite;
-@legend-font-size: 10px;
```
rjmcneill: This shouldn't have been removed, accidentally lost it in the merge.
Will fix along with any other comments

src/assets/styles/dropdown.less
```diff
@@ -0,0 +1,47 @@
+.dropdown {
+  .dropdown-toggle {
+    border: none;
+    background-image: none;
+    background-color: @background-color !important;
```
jleft: Generally, I'd suggest only using `!important` if there's no other (reasonable) way to achieve the desired effect - it can lead to maintenance difficulties and makes debugging more difficult.

I'd guess these styles are overriding the styles set by `btn` and `btn-primary`? Can we just remove these classes from `#product-dropdown-button`?

https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#The_!important_exception

---

As these styles are largely just overriding the styles set in `node_modules/bootstrap/less/dropdowns.less`, it can be worth looking there to see which rules need to be changed.

src/assets/styles/dropdown.less
```diff
@@ -0,0 +1,47 @@
...
+    
+    a {
+      color: @offwhite !important;
+      font-size: @product-dropdown-font-size;
+    }
+    a:hover {
+      background: none;
+    }
+
+    li {
```
jleft: Using `>` we could target the `li` child(ren) of `.dropdown-menu` here, rather than all descendants:

```less
> li {
  padding-left: 20px;
  padding-right: 20px;
  padding-top: 8px;
  padding-bottom: 8px;

  &:hover,
  &:focus {
    background-color: @dropdown-hover;
  }
}
```

Then we shouldn't need to use `!important`.

src/assets/styles/dropdown.less
```diff
@@ -0,0 +1,47 @@
...
+
+  .dropdown-menu {
+    -webkit-border-radius: 2px 2px 2px 2px;
+    -moz-border-radius: 2px 2px 2px 2px;
+    border-radius: 2px 2px 2px 2px;
+    background-color: @dropdown-color;
+    text-align: center;
+    cursor: pointer;
+    
+    a {
```
jleft:
```less
> li > a {
  color: @offwhite;
  font-size: @product-dropdown-font-size;

  &:hover,
  &:focus {
    background: none;
  }
}
```

src/index.html
```diff
@@ -21,7 +21,9 @@
 <div class="container" id="app-container">
     <div class="row row-offcanvas-right head-menu head-row">
         <div class="col-xs-12 col-sm-12 col-md-9 head-sub-row">
-            <select id="product-dropdown" selected="product" class="form-control"></select>
+            <div id="product-dropdown" class="dropdown">
+                <button id="product-dropdown-button" class="btn btn-primary dropdown-toggle" type="button" data-toggle="dropdown"></button>
```
jleft: Do we need the `btn` and `btn-primary` classes?

PR #363, opened 2015-11-04
remove foreground, flex, layoutSuspended, zoom

Afredon:


src/assets/styles/flex.less
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    #navbar-row {
+        flex: 0 0.5 @navigator-height;
+        display: flex;
+        flex-direction: row;
+
+        #navbar-container {
+            flex: 1;
+            margin-right: 15px; // TODO: understand why 15px is needed to align right sides instead of @y-axis-width
```
djmiley: Bootstrap, I believe?

jleft: The default right margin for [`fc.chart.cartesian()` is 30px](https://github.com/ScottLogic/d3fc/blob/5d397e1274cfe1a6342e8c5a35fda66e9948264f/src/chart/cartesian.js#L19), which we're not overriding.

`30px` (from `fc.chart.cartesian()`) + `15px` margin-right for `#navbar-container` === `45px` === `@y-axis-width`

Afredon: ah thanks, so it should be easy to fix by reseting the cartesian chart margin to 0 for the navbat, hopefully.

Afredon: I updated the TODO message for now. Things start to be a bit tricky to merge, so I'd rather fix that in a later group of changes.

src/assets/js/app.js
```diff
@@ -5,12 +5,25 @@
...
+        var containers = {
+            app: appContainer,
+            charts: chartsContainer,
+            primary: chartsContainer.select('#primary-container'),
+            secondaries: chartsContainer.selectAll('.secondary-container'),
+            xAxis: chartsContainer.select('#x-axis-container'),
+            navbar: chartsContainer.select('#navbar-container'),
+            legend: appContainer.select('#legend'),
+            suspendLayout: function(value) {
+                        var self = this;
```
jleft: This block should have its indent decreased by 8 spaces.

src/assets/js/app.js
```diff
@@ -5,12 +5,25 @@
...
+                        Object.keys(self).forEach(function(key) {
+                            if (typeof self[key] !== 'function') {
```
jleft: Could potentially do something like this:

```js
if (self[key] instanceof d3.selection) { ...
```

but IE9 doesn't support it  :(, so can keep as-is...

src/assets/js/app.js
```diff
@@ -52,55 +65,79 @@
...
+            if (layoutRedrawnInNextRender) {
+                containers.suspendLayout(true);
+                layoutRedrawnInNextRender = false;
+            }
         }
 
-        var render = fc.util.render(renderInternal);
+        var fcRender = fc.util.render(renderInternal);
+
+        function render(redrawLayout) {
```
jleft: How about something like this?

```js

function render(redrawLayout) {
    if (redrawLayout) {
        containers.suspendLayout(false);
    }

    fcRender();

    containers.suspendLayout(true);
}
```

I think the `layoutRedrawnInNextRender` and associated `if` statements could then be removed, which would look a bit tidier to me.

src/assets/js/behavior/zoom.js
```diff
@@ -1,12 +1,13 @@
...
     'use strict';
 
-    sc.behavior.zoom = function() {
+    sc.behavior.zoom = function(clientWidth) {
 
         var dispatch = d3.dispatch('zoom');
 
         var zoomBehavior = d3.behavior.zoom();
         var scale;
+        var width = clientWidth;
```
jleft: Could just name the argument `width`, then remove this local var?

src/assets/js/chart/nav.js
```diff
@@ -62,11 +62,17 @@
...
 
-            selection.call(zoom);
+            selection.select('.plot-area').call(zoom);
         }
 
         d3.rebind(nav, dispatch, 'on');
 
+        nav.dimensionChanged = function(selection) {
+            layoutWidth = parseInt(selection.style('width'));
+            viewScale.range([0, layoutWidth]);
```
jleft: Do we need this? Won't Cartesian chart do this for us?

src/assets/js/chart/primary.js
```diff
@@ -195,18 +195,22 @@
...
-            var zoom = sc.behavior.zoom()
+            var zoom = sc.behavior.zoom(zoomWidth)
                 .scale(xScale)
                 .trackingLatest(model.trackingLatest)
                 .on('zoom', function(domain) {
                     dispatch[sc.event.viewChange](domain);
                 });
-            selection.select('.plot-area')
-                .call(zoom);
+            selection.select('.plot-area').call(zoom);
```
jleft: `.call(zoom);` on new line

src/assets/js/util/layout.js
```diff
@@ -1,29 +1,32 @@
...
 
         var useableScreenHeight = window.innerHeight - headRowHeight;
 
-        container.select('#charts-container')
-            .style('height', useableScreenHeight + 'px')
-            .layout();
+        containers.charts
+            .style('height', useableScreenHeight + 'px');
+
+        charts.xAxis.dimensionChanged(containers.xAxis);
```
jleft: I'm not keen on these methods for `dimensionChanged`.

To be consistent with our pattern, I'd remove these methods and add a `width` property to the model passed to the components which require it.

src/assets/styles/flex.less
```diff
@@ -0,0 +1,35 @@
+#charts-container {
```
jleft: 2 space indentation should be used in the Less files. (I know there are a few inconsistencies across the styles at the moment though!)

src/assets/styles/flex.less
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    #navbar-row {
+        flex: 0 0.5 @navigator-height;
+        display: flex;
+        flex-direction: row;
+
+        #navbar-container {
+            flex: 1;
+            margin-right: 15px; // TODO: replace by @y-axis-width once navbar default margin from cartesian chart has been removed.
```
jleft: Any reason not to do this now? Seems like it would be a small change?

src/index.html.tpl
```diff
@@ -27,26 +27,18 @@
...
-                    <svg class="secondary-container" layout-style="flex: 1"></svg>
-                </g>
-                <g id="x-axis-row" layout-style="height: 20; flexDirection: row" >
-                    <svg id="x-axis-container" layout-style="flex: 1"></svg>
-                </g>
-                <g id="navbar-row" layout-style="minHeight: 100; flexDirection: row">
-                    <svg id="navbar-container" layout-style="flex: 1"></svg>
-                </g>
-            </svg>
+            <div id="charts-container">
```
jleft: Much neater :)

jleft:
Thanks @Afredon - some really interesting ideas :)

src/assets/styles/flex.less
Afredon: Only because of the issue of header changing its height at different times. However, if I can leave the bottom margin in the navbar, then yes I can do it right now with no chain of consequence

jleft: Could the right margin for the cartesian chart in nav be set to the yAxisWidth of 60px? This would then be consistent with the primary and the secondary charts and we could remove this style.

src/assets/js/app.js
Afredon: done

jleft:
Just tried this in Firefox and IE and the layout doesn't look right - probably worth investigating further (I don't think we need to focus too much on FF and IE, but it should still look reasonable).

src/assets/js/app.js
Afredon: I don't mind breaking IE9 if you prefer your nicer suggestion :)

Afredon:
Especially as I have fixed an issue for FF and IE last week !

src/assets/styles/flex.less
Afredon: Ideally I would remove the ones from primary and secondary and move everything to the style sheets. Right now we have some in js and some in css and it's a violation of SPOT. And for chosing the single place where to put anything about margins, I'd chose css over the javascript code.

src/assets/js/app.js
Afredon: Good point. I needed to do it this way when I was moving the code, but you're right, now that all render call have a boolean associated, I don't need any global variable anymore

Afredon: Done, but I would keep a similar test on redrawLayout at the end of render to avoid applying suspendLayout(true) on each pass as it is much costier than the boolean check.

src/assets/js/behavior/zoom.js
Afredon: Good point, done

src/assets/js/chart/primary.js
Afredon: done

src/assets/js/util/layout.js
Afredon: The idea was to extract from the renderers 2 kinds of main events : dimension changes and metadata changes. It happens to be width that is currently the only change, but the concept was just dimension change in general.

Afredon: But no problem to replace with a width property for now.

src/assets/js/chart/nav.js
Afredon: we need to update viewScale for the zoom. I'm not sure what you mean the cartesian chart could do for us here ?

src/assets/styles/flex.less
Afredon: done, and it's not nice to trick me :)

src/assets/js/chart/primary.js
```diff
@@ -207,6 +207,11 @@
 
         d3.rebind(primary, dispatch, 'on');
 
+        primary.dimensionChanged = function(selection) {
+            zoomWidth = parseInt(selection.style('width')) - yAxisWidth;
```
djmiley: This isn't on secondary charts, which confuses me, given that we treat them in similar ways? (I may be wrong about this)

src/assets/js/util/layout.js
djmiley: Why don't we call this on `charts.secondaries`? It seems inconsistent.

src/assets/js/behavior/zoom.js
```diff
@@ -1,7 +1,7 @@
 (function(d3, fc, sc) {
     'use strict';
 
-    sc.behavior.zoom = function() {
+    sc.behavior.zoom = function(width) {
```
djmiley: Should we be using the d3 style of method chaining?

src/assets/js/chart/primary.js
```diff
@@ -195,7 +195,7 @@
             primaryChart.plotArea(multi);
             selection.call(primaryChart);
 
-            var zoom = sc.behavior.zoom()
+            var zoom = sc.behavior.zoom(zoomWidth)
```
djmiley: Why can't we use `container.select('.plot-area-container')`, and implement in the the same way as in `src/assets/js/chart/secondary.js`?

src/assets/js/chart/primary.js
Afredon: yes I have stashed my first attempt of applying the same to secondary charts, it's going to be a bit more complicated.

src/assets/js/behavior/zoom.js
Afredon: The reason I put it this way is because it's necessary and not optional, so normally you put that in the constructor parameters. However, I since noticed a same problem with d3fc (in the data generator), so if it seems normal for javascript devs, I'll change it for consistency

djmiley: Yes, more of an aesthetic thing in this case.

src/assets/js/chart/primary.js
Afredon: Looking at it again, I can't remember what was the issue as it seems it would be easy to apply to secondary charts too indeed.

src/assets/js/behavior/zoom.js
jleft: FWIW, there are cases where the factory method has arguments:

https://github.com/ScottLogic/d3fc/blob/master/src/chart/cartesian.js#L12
https://github.com/mbostock/d3/wiki/Requests#d3_xhr

src/assets/js/chart/nav.js
jleft: Ah, sorry, didn't realise it was used with zoom - ignore that comment!

src/assets/js/behavior/zoom.js
Afredon: ok then I'm keen on keeping the required parameters in the constructor, and not for aesthetics, but for the "hard to misuse" principle of APIs.

src/assets/js/chart/nav.js
Afredon: Actually the whole dimensionChanged methods for now are just there to be able to reconfigure the zoom only after a redimensioning instead of each frame. Which goes with the explanation I gave earlier as to why it was not directly in the width property.

src/assets/js/chart/primary.js
Afredon: The idea is to access the DOM only when needed instead of every frame, and similarly to update the zoom method only in the situations where something actually changed.
I'll try and modify secondaries to behave similarly to the changes I made to primary, navbar and xaxis

src/assets/js/app.js
Afredon: Ok, now I remember why I did it this way :)
renderInternal is called only once per frame, but render can be called several times. So if you put the control to render and have render(true) and render(false) in the same frame, you will never recalculate the layout.

src/assets/js/util/layout.js
Afredon: @jleft I removed the return nav/primary/etc to make it clearer that it is not the same kind of action than the ones covered by the pattern.
```  nav.dimensionChanged = function(selection) {
            layoutWidth = parseInt(selection.style('width'));
            viewScale.range([0, layoutWidth]);
        };```

Comment (Afredon) [inline src/assets/js/util/layout.js]: @djmiley Done

File: src/assets/js/model/primaryChart.js [hunk 15]
```diff
@@ -2,15 +2,51 @@
...
-            product: initialProduct
+            metaChanged: false
         };
+
+        var _product = initialProduct;
+        Object.defineProperty(model, 'product', {
+            get: function() { return _product; },
+            set: function(newValue) {
+                _product = newValue;
+                model.metaChanged = true;
```
tlsim: We'll probably want to revisit model change detection in #328, but this looks ok for now.

Afredon: That'll be more code to delete, always a pleasure to do so :)

tlsim:
I'm seeing different behaviour compared to the develop branch in Chrome when resizing - for some window widths, the navbar width gets too small and the menu buttons overlap the main chart.

Afredon:
For the header, the problem comes from a fixed height as a temporary measure until the header is added to flex. I can start working on that to remove the issue.

src/assets/js/model/primaryChart.js
```diff
@@ -2,15 +2,51 @@
...
             trackingLatest: true,
             viewDomain: [],
-            series: sc.menu.option('Candlestick', 'candlestick', sc.series.candlestick()),
-            yValueAccessor: {option: function(d) { return d.close; }},
-            indicators: [],
-            product: initialProduct
+            metaChanged: false
         };
+
+        var _product = initialProduct;
```
jleft: If `_product` can be set by passing `initialProduct` into the factory function, then wouldn't the price format of the primary chart not be updated (on the first render, for example) as `metaChanged` would still be `false`?

src/assets/js/model/primaryChart.js
jleft: Yeah, definitely need to consider how to approach this - if we started doing things like this:

```js
if (_product !== newvalue) {
    _product = newValue;
    model.metaChanged = true;
}
```

And/or had properties which were dependent on others (so needed to start dependency tracking), started tracking changes for each property etc., then I think it could get quite complicated. I don't think we want to implement our own binding framework! I'm ok with this for now too :)

jleft:
Really good to see the performance improvements, thanks @Afredon! Happy to see this merged after the comment on the primary chart model about the initialProduct is addressed.

src/assets/js/model/primaryChart.js
Afredon: You're right and it stands for all properties, so I think the best thing to do is set the initial value of metaChanged to true, which makes sense as creating is a change from the non existing state.

src/assets/js/model/primaryChart.js
Afredon: It could be simplified even without using immutables, by separating into dataModel and contextModel. Sending a contextModel to primary would trigger the metadataChanged method without having to inspect it at all.

PR #369, opened 2015-11-05
Legend resizes depending on time format length, also legend shows no …

jdhodges431:
Legend resizes depending on time format length, also legend shows no values when crosshair is not active.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -139,7 +139,7 @@
         }
 
         function initialiseLegendData(data) {
-            legendModel.data = data[data.length - 1];
+            legendModel.data = null;
         }
 
         function updateModelSelectedProduct(product) {
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -16,12 +16,12 @@
             display: 'Daily',
             seconds: 86400,
             d3TimeInterval: {unit: d3.time.day, value: 1},
-            timeFormat: '%b-%d'});
+            timeFormat: '%b %d'});
         var hour1 = sc.model.period({
             display: '1 Hr',
             seconds: 3600,
             d3TimeInterval: {unit: d3.time.hour, value: 1},
-            timeFormat: '%b-%d %Hh'});
+            timeFormat: '%b %d %Hh'});
         var minute5 = sc.model.period({
             display: '5 Min',
             seconds: 300,
@@ -121,7 +121,7 @@
         }
 
         function onCrosshairChange(dataPoint) {
-            legendModel.data = dataPoint ? dataPoint : primaryChartModel.data[primaryChartModel.data.length - 1];
+            legendModel.data = dataPoint;
             render();
         }
 
@@ -136,7 +136,9 @@
             primaryChartModel.data = data;
             secondaryChartModel.data = data;
             navModel.data = data;
-            // TODO: probably only want to update this if the user isn't already mousing over with the crosshair
+        }
+
+        function initialiseLegendData(data) {
             legendModel.data = data[data.length - 1];
         }
 
@@ -186,6 +188,7 @@
                         console.log('Error getting historic data: ' + err);
                     } else {
                         updateModelData(data);
+                        initialiseLegendData(data);
                         resetToLatest();
                     }
                 });
--- a/src/assets/js/chart/legend.js
+++ b/src/assets/js/chart/legend.js
@@ -6,7 +6,6 @@
         var formatVolume;
         var formatTime;
         var lastDataPointDisplayed;
-        var lastTimeFormatLength;
 
         var legendItems =  [
             'T',
@@ -31,7 +30,7 @@
                 formatVolume = model.product.volumeFormat;
                 formatTime = model.period.timeFormat;
 
-                if (model.data !== lastDataPointDisplayed) {
+                if (!model.data || model.data !== lastDataPointDisplayed) {
                     lastDataPointDisplayed = model.data;
 
                     var span = container.selectAll('span')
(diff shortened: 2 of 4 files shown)


src/assets/js/app.js
```diff
@@ -136,7 +136,9 @@
             primaryChartModel.data = data;
             secondaryChartModel.data = data;
             navModel.data = data;
-            // TODO: probably only want to update this if the user isn't already mousing over with the crosshair
+        }
+
+        function initialiseLegendData(data) {
```
jleft: Should we be doing this? I thought the only time the legend should be active is when the crosshair is being used?

src/assets/js/chart/legend.js
```diff
@@ -33,19 +34,22 @@
...
-                        .data(function(d) { return d; });
+                    var span = container.selectAll('span')
+                        .data(legendItems);
 
                     span.enter()
                         .append('span')
                         .attr('class', function(d, i) { return i % 2 === 0 ? 'legendLabel' : 'legendValue'; });
 
-                    span.text(function(d) { return d3.functor(d)(model.data); });
+                    span.text(function(d, i) { return i % 2 === 0 ? d : model.data ? d(model.data) : ''; });
```
jleft: Has this been changed to handle when `model.data === undefined`?

src/assets/styles/legend.less
```diff
@@ -2,10 +2,9 @@
   position: absolute;
   top: 0;
   right: 65px;
```
jleft: Probably a good place to use `@y-axis-width`?

src/assets/styles/legend.less
```diff
@@ -2,10 +2,9 @@
...
-  margin-top: 15px;
-  padding: 3px 0 3px 10px;
+  margin-top: 10px;
+  padding: 10px;
```
jleft: The designs suggest to me that their should be a `10px` gap between the top of the legend and the top of the plot area, and a `10px` gap between the right side of the legend and the right side of the plot area.

It looks like there's a `20px` gap on the top and a `30px` gap to the right? Should this be adjusted?

src/assets/styles/legend.less
```diff
@@ -15,13 +14,13 @@
   }
 
   .legendLabel {
-    min-width: 15px;
-    color: #4f696c;
+    margin-right: 4px;
+    color: @legend-label-font-colour;
     font-weight: 900;
```
jleft: I don't think we should be using a different font-weight.

src/assets/styles/legend.less
```diff
@@ -15,13 +14,13 @@
...
   }
 
   .legendValue {
```
jleft: The design specifies a spacing to the right of `10px`; would `min-width: 35px;` and `margin-right: 10px;` work?

src/assets/js/chart/legend.js
```diff
@@ -33,19 +34,22 @@
...
+
+                    if (model.data && formatTime(model.data.date).length !== lastTimeFormatLength) {
+                        lastTimeFormatLength = formatTime(model.data.date).length;
+
+                        span.style('min-width', function(d, i) {
```
jleft: Could we get away with setting a larger `min-width` in the CSS to avoid this? Or would that look odd?

jleft:
d3fc makes use of a new SVG legend component in 4.1.0 https://github.com/ScottLogic/d3fc/pull/724 - might be a useful alternative? (Don't have to use it now, but worth considering...)

src/assets/js/chart/legend.js
```diff
@@ -33,19 +34,22 @@
...
 
-                    var p = container.selectAll('p')
-                        .data([legendItems]);
-
-                    p.enter().append('p');
-
-                    var span = p.selectAll('span')
-                        .data(function(d) { return d; });
+                    var span = container.selectAll('span')
+                        .data(legendItems);
```
jleft: Nice - this is much cleaner than before.

src/assets/js/app.js
jdhodges431: So should the legend be empty when the page first loads? If so, is it ok to initialise legendModel.data to null?

src/assets/js/chart/legend.js
jdhodges431: Yeah that's right, is this ok?

src/assets/js/chart/legend.js
jdhodges431: I suppose we could, there's a large gap for smaller time formats. Is this ok? 
image

src/assets/js/app.js
jleft: > So should the legend be empty when the page first loads?

Based on the designs, I think so.

> If so, is it ok to initialise legendModel.data to null?

Yes.

src/assets/js/chart/legend.js
jleft: I was just wondering if could do something clever with `undefined` values, like this: https://github.com/mbostock/d3/wiki/SVG-Shapes#line_defined

This is good for now :) We can always revisit if/when we look at the d3fc legend component

src/assets/js/chart/legend.js
jleft: I think that would my preference for now - if we have feedback that it should be different, then I think it would be easy to add this back in? I could be convinced otherwise though... if you feel strongly about it?

jdhodges431: I have no strong feelings about it, best thing to do is bring it up at the sprint meeting.

src/assets/styles/legend.less
jdhodges431: This seems to cause the legend to resize for values with more digits. A min-width of 45px seems to cover the worst case scenario without any resizing issues.

Ideally we'd adjust the size based on the maximum possible value for the current data being viewed, but this seems like it'd be unnecessary complexity at this point.

jleft:
Thanks @jdhodges431 :+1:

Rebased and merged in #381

PR #372, opened 2015-11-05
Series colours updated

### LINKED ISSUE #370: Series colour
All series should be the same colour: `@secondary-orange: #FF7F00;`

rjmcneill:
Closes #370 

Updated line, point, and area to `@secondary-orange`.
Also added some slight fade onto the area since it was a bit bright.

RSI, MACD, and Volume have not been updated, they should match the style correctly

### DIFF
--- a/src/assets/styles/variables.less
+++ b/src/assets/styles/variables.less
@@ -44,7 +44,7 @@
 
 // Plot Area
 @stroke-width: 1px;
-@area-fill: fade(@tertiary-blue, 75%);
+@area-fill: fade(@series-colour, 75%);
 @series-colour: @secondary-orange;
 
 // Plot Area - Candlestick
--- a/src/assets/styles/variables.less
+++ b/src/assets/styles/variables.less
@@ -45,7 +45,7 @@
 // Plot Area
 @stroke-width: 1px;
 @area-fill: fade(@tertiary-blue, 75%);
-@point-colour: @tertiary-blue;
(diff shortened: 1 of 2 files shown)


src/assets/styles/variables.less
```diff
@@ -45,7 +45,7 @@
 // Plot Area
 @stroke-width: 3px;
 @area-fill: fade(@tertiary-blue, 75%);
```
jleft: Could we update this to: `@area-fill: fade(@series-colour, 75%);`, then reference this variable for `.area` in primary.less?

jleft:
Looks good - please could we also update the colours for the OHLC series to orange? Both the up and down variants (arguably we should use different colours for each, but this way we're being consistent and it's still very easy to change later).

rjmcneill:
Sure, no bother.
Should the moving average and bollingerbands be left the colour they are though?

jleft:
> Should the moving average and bollingerbands be left the colour they are though?

No changes need to be made to ma or boll - they're correct :smile:

rjmcneill:
@jleft the build failed on CI.
I think you said before that this happens sometimes and it just needs rerun?

jleft:
> @jleft the build failed on CI.
> I think you said before that this happens sometimes and it just needs rerun?

Sorry about that. Yeah, it occasionally fails when it shouldn't. All looks good now - thanks!

PR #374, opened 2015-11-05
Font sizing updated to match UX design

### LINKED ISSUE #314: Fonts
Font is [Roboto](https://www.google.com/fonts/specimen/Roboto). Line spacing is generally 1.5 and the weight for all text is light (300).

Numbers are right aligned. Semantic, variable length text is left aligned.

Create sensible variables; something like this:

```less
@font-size-base: 16px;

@font-size-large: 18px;
@font-size-huge: 20px;
@font-size-title: 24px;

@font-size-small: 14px;
@font-size-tiny: 12px;

@line-height-base: 1.5;

@font-weight-base: 300;
```

rjmcneill:
Closes #314

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -68,7 +68,7 @@
 
     sc.chart.primary = function() {
 
-        var yAxisWidth = 45;
+        var yAxisWidth = 60;
         var dispatch = d3.dispatch(sc.event.viewChange, sc.event.crosshairChange);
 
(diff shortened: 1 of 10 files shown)


jleft:
`@y-axis-width` and `@x-axis-height` in `variables.less` should also be updated, as should the `yAxisWidth` in `sc.chart.secondary`

rjmcneill:
Missed those sorry!
Update won't be long :)

jleft:
No worries - thanks for the update :+1:

PR #376, opened 2015-11-06
Crosshair updated to match UX designs

### LINKED ISSUE #321: Crosshair

rjmcneill:
Closes #321 

Updated the crosshair to match the UX designs for the candlestick view.
Currently left the crosshair the way it was previously done for all other views.

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -171,6 +171,60 @@
             movingAverage(model.data);
             bollingerAlgorithm(model.data);
 
+            // Update the crosshair depending on the series displayed
+            if (currentSeries.valueString === 'candlestick' || currentSeries.valueString === 'ohlc') {
+                //If the candlestick or ohlc selected, updated the crosshair to add classes to reflect UX designs
+                selection.each(function(data) {
+                    var barWidth = fc.util.fractionalBarWidth(0.75);
+                    var xValue = function(d) { return d.date; };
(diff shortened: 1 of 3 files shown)


src/assets/js/series/candlestick.js
```diff
@@ -30,6 +30,34 @@
             .element('path')
             .attr('class', 'down');
 
+        var crosshair = function(selection) {
```
jleft: This is a neat idea, but I don' think this is the responsibility of the candlestick series.

The candlestick component should be responsible for rendering the candlestick series, not the candlestick series and the crosshair. I also don't see the candlestick series as _having a_ crosshair; the crosshair belongs to the primary chart.

Could the crosshair in primary chart be conditionally [decorated](https://github.com/ScottLogic/d3fc/blob/ba456f3df4f96a8ab5e1aa9ed3fcd52854ef5a10/src/tool/crosshair.js#L89) (based on the series shown) to achieve the desired result?

src/assets/js/chart/primary.js
```diff
@@ -169,6 +168,38 @@
             movingAverage(model.data);
             bollingerAlgorithm(model.data);
 
+            // Update crosshair if candlestick or ohlc series selected
```
jleft: Could this be done using [`decorate()`](https://github.com/ScottLogic/d3fc/blob/master/src/tool/crosshair.js#L142-L148) on the crosshair?

http://d3fc.io/components/introduction/2-decorate-pattern.html

PR #386, opened 2015-11-11
Reset button styling and position updated to new design

### LINKED ISSUE #315: Buttons

### LINKED ISSUE #323: Reset button

rjmcneill:
Closes #323 
Closes #315 as well since this is the last button that needed styling


hello.patch
```diff
@@ -0,0 +1,51 @@
+diff --git a/src/assets/js/chart/nav.js b/src/assets/js/chart/nav.js
```
jleft: Remove this file

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
         var layoutWidth;
 
         function nav(selection) {
-            var model = selection.datum();
+            var navbarContainer = selection.select('#navbar-container');
+            var navbarReset = selection.select('#navbar-reset');
+            var model = navbarContainer.datum();
+
+            navbarContainer.style('margin-right', '0');
```
jleft: This is overwriting a style in the Less - can we just change the Less?

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
...
 
             viewScale.domain(model.viewDomain);
+            navbarReset
+                .style('margin-right', '21')
+                .style('margin-left', '15')
+                .style('margin-top', '10')
+                .select('g')
```
jleft: Do we need the `g` element?

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
...
+                .on('click', function() { dispatch[sc.event.resetToLatest](); })
+                .attr('xmlns', 'http://www.w3.org/2000/svg')
```
jleft: Don't think we need this?

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
...
+                .classed('reset-button-inactive', model.trackingLatest)
+                .classed('reset-button-active', !model.trackingLatest)
+                .html(function() {
```
jleft: Stylistically, prefer:

```js
.append('path')
.attr('d', 'M1.5 1.5h13.438L23 20.218 14.937 38H1.5l9.406-17.782L1.5 1.5z');
```

However, will this not be updating the `path` element on every render? Can we make use of an enter selection to make sure this doesn't happen? (Could be an argument for using `decorate` on cartesian chart)

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
...
+                    return '<path d="M1.5 1.5h13.438L23 20.218 14.937 38H1.5l9.406-17.782L1.5 1.5z"></path>';
+                });
+
+            viewScale.domain(model.viewDomain)
```
jleft: The domain is being set twice (here and on line 41) - only need to set it once?

src/assets/js/chart/nav.js
```diff
@@ -30,9 +32,28 @@
...
+                .range([0, navbarContainer.node().getAttribute('layout-width')]);
```
jleft: Don't need this - see `nav.dimensionChanged`

src/assets/js/chart/nav.js
```diff
@@ -67,7 +88,8 @@
                     dispatch[sc.event.viewChange](domain);
                 });
 
-            selection.select('.plot-area').call(zoom);
+            navbarContainer.select('.plot-area').call(zoom);
+            navbarContainer.call(zoom);
```
jleft: Only need to call zoom on the plot area?

src/assets/styles/navigator.less
```diff
@@ -1,33 +1,52 @@
...
+      fill: @font-colour;
+    }
   }
 
-  text {
-    font-size: @font-size-base;
-    fill: @font-colour;
+  #navbar-reset {
+    stroke: @primary-white;
+    width: 24px;
```
jleft: If we're flex for layout, would this be more appropriate?

```less
flex: 0 0 24px;
```

src/assets/styles/navigator.less
```diff
@@ -1,33 +1,52 @@
...
+    height: 40px;
+    opacity: .199;
```
jleft: Could just use `0.2`?

src/assets/styles/navigator.less
```diff
@@ -1,33 +1,52 @@
...
+    fill-rule: evenodd;
```
jleft: Do we need this?

src/index.html.tpl
```diff
@@ -37,6 +34,7 @@
                 </div>
                 <div id="navbar-row">
                     <svg id="navbar-container"></svg>
+                    <svg id="navbar-reset"><g></g></svg>
```
jleft: Don't think we need the `g` (again, possibly an argument for using `decorate` on cartesian chart)

src/assets/js/chart/nav.js
Afredon: There is also a mix-up that probably comes from rebasing on my last changes with the .call(zoom) on same line than the select and now a duplication of it and an unneeded repetition of navbarContainer either.

src/assets/js/chart/nav.js
rjmcneill: Sorry, these came from merging in Arnaud's stuff incorrectly

src/assets/styles/navigator.less
rjmcneill: This is what was given to us in the icons @jleft 
Should we just make it 0.2 though?
Same with the oddeven fill rule

jleft: The design also specifies:

> Colour when user scrolls away from most up-to-date data = #FFFFFF fill - with 20% opacity

I'd use `0.2` in this case - I think that was the intention, based on the above comment in the designs (it's a nicer number too!).

> Same with the oddeven fill rule

Happy to use this, if it's needed. If it's not, then I think it'd be best to remove it.

Issue #263, opened 2015-09-21
Change OHLC selection to be in configuration of individual primary chart indicators


PaulScho:

djmiley:
Move OHLC selection into configuration for primary chart indicators only, see #73

jleft:
Closing: The OHLC config for all series/indicators has been removed and the configuration of individual indicators is covered by #73

PR #387, opened 2015-11-12
Added dropdowns for both series and indicator selection.

### LINKED ISSUE #316: Drop-downs/components

jdhodges431:
Closes #316.


src/assets/js/app.js
```diff
@@ -60,6 +60,7 @@
         var primaryChartModel = sc.model.primaryChart(generated);
         var secondaryChartModel = sc.model.secondaryChart(generated);
         var sideMenuModel = sc.model.menu.side();
```
jleft: Can we remove the side menu now?

Much of the functionality and code is duplicated between the side menu and components menu. It's fine if we lose the OHLC functionality - this is going to be changed to a per-indicator basis anyway #263.

src/assets/js/app.js
```diff
@@ -106,6 +108,10 @@
             containers.app.select('.head-menu')
                 .datum(headMenuModel)
                 .call(headMenu);
+            //
```
jleft: Remove `//`

src/assets/js/menu/components.js
```diff
@@ -0,0 +1,67 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.menu.components = function() {
```
jleft: I'm not sure about the naming - components is fairly generic. I'm struggling to think of a better name though... maybe `sc.menu.plotArea` as it affects what is on the plot area for the charts? Does anyone else have any suggestions?

src/assets/js/menu/components.js
```diff
@@ -0,0 +1,67 @@
...
+        var components = function(selection) {
+            selection.each(function(model) {
+                var selection = d3.select(this);
+
+                var selectedSeriesIndex = model.seriesOptions.map(function(option) {
+                    return option.isSelected;
+                }).indexOf(true);
+
+                if (selectedSeriesIndex !== lastSeries) {
+                    lastSeries = selectedSeriesIndex;
```
jleft: > Components have no internal state. They can be called on multiple selections without needing to clear down any internal state.
http://d3fc.io/components/introduction/4-component-design.html

If a component stores internal state, issues can arise if a component is called upon a selection with multiple elements with different data bound to each element.

Although we may get a performance increase by checking the last series and last indicator values here, I'd suggest that we remove this to avoid the problem as described above; it will simplify the code here and be consistent with the pattern. If we want to add this optimisation, I'd suggest we do it later. If we were to do it, I don't think it would be renderers responsibility.

Interested to hear opinions on this :)

src/assets/js/menu/components.js
```diff
@@ -0,0 +1,67 @@
...
+
+                    selection.select('#series-dropdown')
+                        .datum({options: model.seriesOptions,
```
jleft:
```js
.datum({
    options: model.seriesOptions,
    selectedIndex: selectedSeriesIndex
})
```

Same for indicators too please

src/assets/js/menu/generator/dropdownGroup.js
```diff
@@ -8,28 +8,63 @@
             .selector('ul')
             .element('ul');
 
+        var config;
+        var lastSelectedIndex;
```
jleft: Same comment as above about storing state

src/assets/js/menu/components.js
Afredon: I'm sure the naming is not satisfying as it threw me off when I first tried to get what was happening. Does that cover series and indicators ? I call that metadata or context for now, but it would be food to have a universal vocabulary about these. Selectors could do it too.

jleft: > Selectors could do it too

Selectors came to my mind too - I think that could work :+1:

jdhodges431: Components is used in the design spec but selectors sounds much better.

Afredon: I'll change my metadataChanged method to selectorsChanged method for clarity then if we're all ok with that name

src/assets/js/menu/generator/dropdownGroup.js
```diff
@@ -8,28 +8,63 @@
...
+                var model = selection.datum();
+                var ul = dataJoin(selection, [model.options]);
+
+                if (!config) {
+                    config = {};
+                    selection.each(function() {
+                        config.title = this.getAttribute('component-title');
+                        config.careted = this.hasAttribute('careted');
+                        config.listIcons = this.hasAttribute('list-icons');
+                        config.icon = this.hasAttribute('icon');
```
jleft: This is an interesting solution - adding attributes to the elements - but quite different to our current approach.

I'm no expert on adding custom attributes to elements, but if we go down this route I'd suggest that we use [`data-*` attributes](http://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes) to avoid any compatibility problems in the future.

Our current method of associating some data with an element is by adding the properties to a 'model' (just a JS object) and binding the model to the element (`selection.datum(model)`) - this would work here. An alternative would be to add these properties as [configuration properties to the component](http://bost.ocks.org/mike/chart/#reconfiguration).

I think this is worth further discussion.

src/assets/js/menu/generator/dropdownGroup.js
```diff
@@ -8,28 +8,63 @@
...
-                .data(fc.util.fn.identity);
+                var li = ul.selectAll('li')
+                    .data(fc.util.fn.identity);
 
-            li.enter()
-                .append('li')
-                .on('click', dispatch.optionChange)
-                .append('a')
-                .attr('href', '#');
+                li.enter()
```
jleft: As a general point, when we're manipulating the DOM, we'd normally use D3, e.g.:

Prefer:

```js
selection.text(function(d) { return d.displayString });
```

instead of:

```js
this.textContent = d.displayString;
```

jleft:
Rebased and updated in #435

PR #388, opened 2015-11-13
Navigator handles styled and functionality implemented

### LINKED ISSUE #324: Navigator handles

rjmcneill:
Closes #324

### DIFF
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -100,7 +100,7 @@
                     brushHide = true;
                 }
 
-                nav.setHide(selection, brushHide);
+                setHide(selection, brushHide);
 
                 if (brush.extent()[0][0] - brush.extent()[1][0] !== 0) {
                     dispatch[sc.event.viewChange]([brush.extent()[0][0], brush.extent()[1][0]]);
@@ -108,7 +108,7 @@
             })
             .on('brushend', function() {
                 brushHide = false;
-                nav.setHide(selection, brushHide);
+                setHide(selection, brushHide);
 
                 if (brush.extent()[0][0] - brush.extent()[1][0] === 0) {
                     dispatch[sc.event.viewChange](sc.util.domain.centerOnDate(viewScale.domain(),
@@ -133,14 +133,14 @@
 
         d3.rebind(nav, dispatch, 'on');
 
-        nav.setHide = function(selection, brushHide) {
+        function setHide(selection, brushHide) {
             selection.select('.plot-area')
                 .selectAll('.e')
                 .classed('hidden', brushHide);
             selection.select('.plot-area')
                 .selectAll('.w')
                 .classed('hidden', brushHide);
-        };
+        }
 
         nav.dimensionChanged = function(container) {
             layoutWidth = parseInt(container.style('width'));
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -9,7 +9,7 @@
         var navChart = fc.chart.cartesian(fc.scale.dateTime(), d3.scale.linear())
             .yTicks(0)
             .margin({
-                bottom: 40
+                bottom: 40      // Variable also in navigator.less - should be used once ported to flex
             });
 
         var viewScale = fc.scale.dateTime();
@@ -25,25 +25,25 @@
                 selection.enter()
                     .select('.e')
                     .append('circle')
-                    .attr('cy', 34)
+                    .attr('cy', 29)
                     .attr('r', 7)
                     .attr('class', 'outer-handle');
                 selection.enter()
                     .select('.e')
                     .append('circle')
-                    .attr('cy', 34)
+                    .attr('cy', 29)
                     .attr('r', 4)
                     .attr('class', 'inner-handle');
                 selection.enter()
                     .select('.w')
                     .append('circle')
-                    .attr('cy', 34)
+                    .attr('cy', 29)
                     .attr('r', 7)
                     .attr('class', 'outer-handle');
                 selection.enter()
                     .select('.w')
                     .append('circle')
-                    .attr('cy', 34)
+                    .attr('cy', 29)
                     .attr('r', 4)
                     .attr('class', 'inner-handle');
             })
@@ -85,15 +85,31 @@
             var yExtent = fc.util.extent()
                 .fields(['low', 'high'])(filteredData);
 
+            var brushHide = false;
+
             navChart.xDomain(fc.util.extent().fields('date')(model.data))
                 .yDomain(yExtent);
 
             brush.on('brush', function() {
+                var width = selection.select('.plot-area').select('.extent').attr('width');
+
+                // Hide the bar if the extent has no length
+                if (width > 0) {
+                    brushHide = false;
+                } else {
+                    brushHide = true;
+                }
+
+                nav.setHide(selection, brushHide);
+
                 if (brush.extent()[0][0] - brush.extent()[1][0] !== 0) {
                     dispatch[sc.event.viewChange]([brush.extent()[0][0], brush.extent()[1][0]]);
                 }
             })
             .on('brushend', function() {
+                brushHide = false;
+                nav.setHide(selection, brushHide);
+
                 if (brush.extent()[0][0] - brush.extent()[1][0] === 0) {
                     dispatch[sc.event.viewChange](sc.util.domain.centerOnDate(viewScale.domain(),
                         model.data, brush.extent()[0][0]));
@@ -116,6 +132,15 @@
         }
 
         d3.rebind(nav, dispatch, 'on');
+
+        nav.setHide = function(selection, brushHide) {
+            selection.select('.plot-area')
+                .selectAll('.e')
+                .classed('hidden', brushHide);
+            selection.select('.plot-area')
+                .selectAll('.w')
+                .classed('hidden', brushHide);
+        };
 
         nav.dimensionChanged = function(container) {
             layoutWidth = parseInt(container.style('width'));
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -22,27 +22,30 @@
         var navMulti = fc.series.multi()
             .series([area, line, brush])
             .decorate(function(selection) {
-                //The following selections append on the navigator handles to the navigator bars
-                selection.select('.e')
+                selection.enter()
+                    .select('.e')
                     .append('circle')
                     .attr('cy', 34)
                     .attr('r', 7)
-                    .classed('outer-handle', true);
-                selection.select('.e')
+                    .attr('class', 'outer-handle');
+                selection.enter()
+                    .select('.e')
                     .append('circle')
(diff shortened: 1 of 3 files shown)


src/assets/js/chart/nav.js
```diff
@@ -17,15 +17,40 @@
         var line = fc.series.line()
             .yValue(function(d) { return d.close; });
         var brush = d3.svg.brush();
-        var navMulti = fc.series.multi().series([area, line, brush])
+        var navMulti = fc.series.multi()
+            .series([area, line, brush])
+            .decorate(function(selection) {
+                //The following selections append on the navigator handles to the navigator bars
```
jleft: Remove comment

rjmcneill:
Ah my bad sorry!
I think I had that before, but with all the code I changed last night trying to get it working I might have lost those changes.
I'll get that sorted and update the PR

src/assets/js/chart/nav.js
```diff
@@ -17,15 +17,40 @@
...
+                selection.select('.e')
+                    .append('circle')
+                    .attr('cy', 34)
+                    .attr('r', 7)
+                    .classed('outer-handle', true);
```
jleft: `.attr('class', 'outer-handle');` same for classes below too please

src/assets/js/chart/nav.js
```diff
@@ -17,15 +17,40 @@
...
+                    .attr('r', 4)
+                    .classed('inner-handle', true);
+            })
             .mapping(function(series) {
                 if (series === brush) {
                     brush.extent([
                         [viewScale.domain()[0], navChart.yDomain()[0]],
                         [viewScale.domain()[1], navChart.yDomain()[1]]
                     ]);
+                } else {        // This stops the brush data being overwritten by the point data
```
jleft: Phew! Glad the fix wasn't too bad in the end - nice one :)

Comment on a new line below please

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
         stroke: @navigator-colour;
         stroke-width: @navigator-stroke;
       }
+      .n, .s, .ne, .nw, .se, .sw {
+        cursor: default !important;   // Removed cursor set by brush. Gave false impression of resizing ability
```
jleft: I'd just use `display: none;`

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
...
+          width: @stroke-width;
+          height: @navigator-handle-height;
+          fill: @navigator-stroke-colour;
+          stroke: @navigator-stroke-colour;
+          stroke-opacity: 0.75;
+          transform: translate(2px, 2px);
+          visibility: visible !important;
+        }
+      }
+      .outer-handle {
```
jleft: Can we move this into the `.e, .w {` block?

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
...
+        stroke: @primary-white;
+        stroke-width: @stroke-width;
+        fill: @primary-dark-blue;
+      }
+      .inner-handle {
```
jleft: Can we move this into the `.e, .w {` block?

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
...
+          fill: @navigator-stroke-colour;
+          stroke: @navigator-stroke-colour;
+          stroke-opacity: 0.75;
+          transform: translate(2px, 2px);
+          visibility: visible !important;
+        }
+      }
+      .outer-handle {
+        stroke: @primary-white;
+        stroke-width: @stroke-width;
```
jleft: Do we need to set the stroke-width? Or will `1 px` be inherited?

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
...
+        cursor: default !important;   // Removed cursor set by brush. Gave false impression of resizing ability
+      }
+      .e, .w {
+        > rect {
+          width: @stroke-width;
+          height: @navigator-handle-height;
+          fill: @navigator-stroke-colour;
+          stroke: @navigator-stroke-colour;
+          stroke-opacity: 0.75;
+          transform: translate(2px, 2px);
```
jleft: This is an attr, rather than a style - I don't think it will do anything?

src/assets/styles/navigator.less
```diff
@@ -23,6 +25,29 @@
...
+      .n, .s, .ne, .nw, .se, .sw {
+        cursor: default !important;   // Removed cursor set by brush. Gave false impression of resizing ability
+      }
+      .e, .w {
+        > rect {
+          width: @stroke-width;
+          height: @navigator-handle-height;
+          fill: @navigator-stroke-colour;
+          stroke: @navigator-stroke-colour;
+          stroke-opacity: 0.75;
```
jleft: Do we need to set the opacity? I can't see anything about this in the design spec, unless I'm missing something...

src/assets/styles/navigator.less
```diff
@@ -2,8 +2,10 @@
   min-height: @navigator-height;
 
   .extent {
+    height: @navigator-extent-height;
+    transform: translateY(1.25px);
     fill: @navigator-extent-colour;
-    stroke: @navigator-selection-border-colour;
+    stroke-opacity: 0;
```
jleft: `stroke: none;`

src/assets/styles/navigator.less
```diff
@@ -2,8 +2,10 @@
   min-height: @navigator-height;
 
   .extent {
+    height: @navigator-extent-height;
+    transform: translateY(1.25px);
```
jleft: This is an attr, rather than a style - I don't think it will do anything?

src/assets/styles/variables.less
```diff
@@ -82,6 +82,9 @@
 @navigator-colour: @primary-blue;
 @navigator-background-colour: @secondary-dark-blue;
 @navigator-stroke: 1px;
+@navigator-stroke-colour: @primary-white;
+@navigator-extent-height: 69px;
+@navigator-handle-height: 68px;
```
jleft: These heights didn't need to be specified before, why do we need them now?

src/assets/styles/navigator.less
rjmcneill: Yea, this works.
It's needed to account for changing the size of these bars.
I can make it an attribute if you'd rather have that?

src/assets/styles/variables.less
rjmcneill: The height of these is 70 by default, but with the translates (to stop them overlapping the outer rect of the full navigator) they overflow into the x-axis.  The overlap is automatically hidden, since it's behind the xaxis, but I thought it would be better to make the bars and extent smaller to avoid any issues we might have further down the line.

I don't mind though, your call @jleft

src/assets/styles/navigator.less
rjmcneill: Yea, this works.
It's needed to account for changing the size of the highlighted section.
I can make it an attribute if you'd rather have that @jleft ?

src/assets/styles/variables.less
Afredon: it would help to write it like
```calc(@height - @borderwidth) ```
with whatever variables needed to express where the number come from.

Comment (jleft) [inline src/assets/styles/variables.less]: I like @Afredon suggestions - if we can calculate the values from meaningful variables, then I think it would be a lot clearer; otherwise they're like magic numbers.

File: src/assets/styles/navigator.less (hunk 8 above)
Comment (jleft) [inline src/assets/styles/navigator.less]: Sorry, my head was in SVG land, I see these are CSS tranforms now!

If they work, then I'm fine with these.

File: src/assets/styles/variables.less (hunk 12 above)
Comment (rjmcneill) [inline src/assets/styles/variables.less]: The problem with this @Afredon is that the height (70px) is JavaScript side, and the stroke width is less side.
I know the height of the navigator stays at a constant 70px.
So I basically did this by eye.

Any way you can think around this?

Comment (Afredon) [inline src/assets/styles/variables.less]: I'd add a @navigator-bottom-margin: 30px in less, even though we don't use it for now, as the margin comes from javascript and then calc(@navigator-height - @navigator-bottom-margin - @yourreasontocrop1px)
We're almost done with the javascript margin which will disappear once the head is also handled with Flex, so it will be easy to finish the cleanup if we have an explicit variable for that 30px

Comment (rjmcneill) [inline src/assets/styles/variables.less]: Yea of course.
That makes perfect sense.
I think this is a pretty clean way of handling it for now.
Then we can remove the 'from javascript' part when you're actually using it

**Comment by rjmcneill** [comment]
@Afredon suggested adding a variable for the navigator bottom margin to make sure it's clear that this is represented in less temporarily as well as JS side.
I'll add that in if there are any further comments on this PR

**Comment by Afredon** [comment]
comment :)

File: src/assets/styles/variables.less [hunk 13]
```diff
@@ -78,11 +78,16 @@
 
 // Navigator
 @navigator-height: 100px;
+@navigator-bottom-margin-from-javascript: 30px;   //Temporary variable until fully pored to flex
```
jleft: `pored` -> `ported`?

src/assets/styles/variables.less
```diff
@@ -78,11 +78,16 @@
...
 @navigator-extent-colour: @primary-white;
 @navigator-colour: @primary-blue;
 @navigator-background-colour: @secondary-dark-blue;
 @navigator-stroke: 1px;
+@navigator-stroke-colour: @primary-white;
 @navigator-selection-border-colour: @primary-white;
+// The following resize the extent and handle heights - needed to correctly size them inside of the navigator
+@navigator-extent-height: calc(@navigator-height - @navigator-bottom-margin-from-javascript - @stroke-width);
```
jleft: Don't need to use `calc` here. Less can perform the calculation during transpilation:

http://lesscss.org/less-preview/#%7B%22less%22%3A%22%40width%3A%2010px%3B%5Cn%40margin%3A%202px%3B%5Cn%5Cn.test%20%7B%5Cn%20%20width%3A%20(%40width%20-%20%40margin)%3B%5Cn%7D%22%7D

jleft:
I tried this out and noticed when clicking to drag out a new brush, on the initial click (before dragging), the position of the brush handles/rect isn't right - the top of the brush is located when the cursor is, rather than at the top of the navigator - is this something which can be fixed? I noticed the `rect`'s height attr seems to change - might be worth looking at?

brush-position

rjmcneill:
I'll certainly have a look at this James.
This might be something to do with moving from the full rectangle to just using the West and East parts of the brush.
I'll have a look shortly and update you after

src/assets/js/chart/nav.js
```diff
@@ -88,6 +133,15 @@
 
         d3.rebind(nav, dispatch, 'on');
 
+        nav.setHide = function(selection, brushHide) {
```
jleft: I don't think this should be public, same reasoning as here: https://github.com/ScottLogic/d3fc-showcase/pull/417/files#r45355778

PR #417, opened 2015-11-19
Fix to stop the graph breaking when zoomed in too far

### LINKED ISSUE #411: Zoomed in extent is unlimited
#168 

Don't worry about limiting the zoom to a specific zoom level, just ensure the user can't zoom in so far that the chart blows up.

rjmcneill:
Resolves #411


src/assets/js/behavior/zoom.js
```diff
@@ -54,14 +54,22 @@
...
                         } else if (zoomed && trackingLatest) {
                             domain = sc.util.domain.moveToLatest(domain, selection.datum().data);
                         }
-                        dispatch.zoom(domain);
+
+                        var lowDomain = domain[0].toString();
+                        var highDomain = domain[1].toString();
+
+                        // Check that the low and high domains are not the same value due to zoom level
+                        if (lowDomain.localeCompare(highDomain) !== 0) {
```
rjmcneill: Sorry, we can just use a direct comparison here.
This was a method I was using before I had it worked out properly.

``` javascript
if (lowDomain !== highDomain) {
```

djmiley: Would `.getTime()` comparison not work?

src/assets/js/behavior/zoom.js
```diff
@@ -100,6 +108,12 @@
             return zoom;
         };
 
+        // Resets zoomBehaviour
```
djmiley: Possibly keep it as a private function?

src/assets/js/behavior/zoom.js
rjmcneill: Yup, looks like that works as well.

src/assets/js/behavior/zoom.js
rjmcneill: Figured we might end up using this function in when we're limiting the zoom to a specific number of periods as well, so thought it might be better to pull this out into a separate function now.

Not bothered though, can put it back to how it was if everyone thinks that's best

jleft: I'd remove the comment - the code is self-documenting.

Separate function is fine; I think it's its visibility which is the issue. The function is effectively 'public', i.e. from anywhere in our application we could do:

```js
var zoom = sc.behavior.zoom();
zoom.resetBehaviour(); // this works!
```

I don't know if ever makes sense to do this?

I agree with with Drew that this could be private:

```js
// Also move its declaration to above function zoom(...
function resetBehaviour() {
    zoomBehavior.translate([0, 0]);
    zoomBehavior.scale(1);
}
```
This would make the function private (it's still available within `sc.behavior.zoom`, but not outside)

```js
var zoom = sc.behavior.zoom();
zoom.resetBehaviour(); // Uncaught TypeError: zoom.resetBehaviour is not a function(…)

Comment (rjmcneill) [inline src/assets/js/behavior/zoom.js]: Ah right, I know what you mean now.
Yes, that makes sense.
I had what Drew mean mixed up just.

Yea, that sounds like a better solution

**Comment by djmiley** [comment]
:+1: LGTM

**Comment by djmiley** [comment]
Might have to get it up-to-date.

**Comment by jleft** [comment]
:+1: - I'm happy to rebase this and merge now :)

File: src/assets/js/behavior/zoom.js [hunk 3]
```diff
@@ -54,14 +54,19 @@
...
                         if (maxDomainViewed) {
                             domain = xExtent;
+
                         } else if (zoomed && trackingLatest) {
                             domain = sc.util.domain.moveToLatest(domain, selection.datum().data);
                         }
-                        dispatch.zoom(domain);
+
+                        // Check that the low and high domains are not the same value due to zoom level
+                        if (domain[0].getTime() !== domain[1].getTime()) {
```
Afredon: I didn't see the requirement, but it feels weird we want equality to be at the 1ms level. Shouldn't we use a higher granularity?

jleft:
Thanks @rjmcneill - rebased and merged in #429

rjmcneill:
Thanks @jleft :)

PR #418, opened 2015-11-19
Navbar highlight

Afredon:


src/assets/styles/navigator.less
```diff
@@ -6,7 +6,7 @@
     .extent {
       fill: @navigator-extent-colour;
       stroke: @navigator-selection-border-colour;
-      fill-opacity: 0.125;
+      fill-opacity: 0;
```
rjmcneill: Can we take away:

```javascript
fill: @navigator-extent-colour;
fill-opacity: 0;
```

and just have

``` javascript
fill: none
```

Afredon: Totally, that'll be better as we can cheat-style extent through highlighted area now.

src/assets/js/chart/nav.js
```diff
@@ -6,30 +6,161 @@
...
+                    iMax = iMid;
+                }
+            }
+            // deferred equality
+            return getLeftValue ? iMin : iMax;
+        };
+
+        var findIntervalIndexes = function(data, selectedDates) {
+            // returns the indexes of the widest interval in data than is included in [left;right]
+            var leftHighlightIndex = binarySearch(data, selectedDates.left, false);
```
tlsim: Could you get the same result using `d3.bisect`?

src/assets/js/chart/nav.js
```diff
@@ -6,30 +6,161 @@
...
+            output.left = data.slice(0, highlightIndexes.left);
+            output.highlight = data.slice(highlightIndexes.left, highlightIndexes.right + 1);
+            output.right = data.slice(highlightIndexes.right + 1);
+
+            addInterpolatedPoints(selectedDates, output);
+
+            return output;
+        };
+
+        var navMulti = fc.series.multi().series([areaLeft, areaHighlight, areaRight, line, brush])
```
tlsim: Might make sense to have a new series component to manage the area splitting.

src/index.html.tpl
```diff
@@ -33,7 +33,18 @@
                     <svg id="x-axis-container"></svg>
                 </div>
                 <div id="navbar-row">
-                    <svg id="navbar-container"></svg>
+                    <svg id="navbar-container">
+                        <defs>
+                            <linearGradient id="gradient-highlight" x1="0%" x2="0%" y1="0%" y2="100%">
```
tlsim: Prefer creating these elements using d3.

src/assets/js/chart/nav.js
```diff
@@ -6,30 +6,161 @@
...
+        var navChart = fc.chart.cartesian(fc.scale.dateTime(), yScale)
             .yTicks(0)
             .margin({
                 bottom: 40
             });
 
         var viewScale = fc.scale.dateTime();
+        var areaData = {};
+
+        var forcePathTop = function(path) {
```
jleft: As a general point, where possible, I'd use named functions rather than anonymous functions. Mainly because their names will appear in stack traces.

```js
function forcePathTop(path) { ...
```

tlsim:
Some good progress @afredon. I think this needs rebasing to see if it works with the new navbar handles. Is there any particular advantage of this approach of splitting the area into 3 areas over, say, drawing a new SVG area with the gradient styling, clipped to make only the region inside the brush extent visible? This might remove some of the complication of searching for indexes and interpolating.

Afredon:
I had 3 possible solutions, the area split, drawing over the brush, or clipping at the brush.
I went with this one as it was using purely d3/d3fc, but I was intrigued on trying the clipping too.

src/assets/js/chart/nav.js
Afredon: Ahhhh, I thought it was just 2 different syntaxes for the same thing, I didn't realise I was declaring anonymous functions. I will change that, I prefer named ones too.

src/index.html.tpl
Afredon: I didn't think of d3 here, I will change that :)

src/assets/js/chart/nav.js
Afredon: After looking at d3.bisect, it looks like yes we can get the same result, with slightly lower performances though (2 passes instead of one per point and no deferred equality)

src/assets/js/chart/nav.js
Afredon: It does make sense and it's interesting, but I don't know if I'll have enough time for it.

src/index.html.tpl
Afredon: Done

src/assets/js/chart/nav.js
Afredon: done

src/assets/styles/navigator.less
```diff
@@ -20,6 +19,12 @@
       .multi {
         .area {
           fill: @gridline-colour;
+          &.highlight {
+              fill: url('/#gradient-highlight');
```
jleft: Change to relative path -> `url('#gradient-highlight')`

src/assets/styles/navigator.less
```diff
@@ -20,6 +19,12 @@
...
+          }
+          &.unselected {
+              fill: url('/#gradient-unselected');
```
jleft: Change to relative path -> `url('#gradient-unselected')`

Afredon: That's what I had initially, but it took me a while that was the reason firefox never picks up the gradients. Firefox works only with the "/"

jleft: Ah, okay. Seems like the behaviour is different when specifying in the CSS: http://stackoverflow.com/questions/15123953/svg-fill-url-behaving-strangely-in-firefox

If we use an absolute link then it won't work if the site is run from a path which isn't `/`; e.g. when deploying to gh-pages, it's `/d3fc-showcase/`

Afredon: It might work without the / if I create the def in the same "g" than where the path is. Let me try.
Otherwise it's going to be a pain to deploy :(

Afredon: For rectangles, it works fine on this example: http://vanseodesign.com/web-design/svg-linear-gradients/

jleft: Could be because they're in-lining the fill style, rather than using it in the stylesheet?

Afredon: Maybe. I tried to inline the fill="url(...)" in my path but it has no effect.

rjmcneill:
Not sure if this works in comments but...
Closes #343

jleft:
Thank you for this solution @Afredon, there are some really interesting idea here :smiley:

However, we decided to use #526 (which uses a mask) - the changes made were small and maintainable, and its performance seems good enough.

PR #420, opened 2015-11-19
Travis integration for Slack

jleft:


jleft:
Ok, let's try this...

Issue #332, opened 2015-10-27
Re-evaluate data components 


jleft:

jleft:
#67

rjmcneill:
Added further coinbase products in #437

tlsim:
Refer to label: hardening/data.

Issue #406, opened 2015-11-18
Investigate impact of using a larger data set


jleft:
Initially try with ~3000 data points.

+ What is the impact?
+ Would it be easy to use a sampler? Would this be worthwhile?

jdhodges431:
It's definitely suffering with 3000 data points. At 1000 it's noticeably slow. I think it's worth looking into using a sampler, we could set a cap for the amount of data points we can manage at once and have the sampler group the data into buckets in order to stay below this threshold.

Afredon:
Try and compare where and how performance degrades. Change of scope like that can lead to detect O(n2) issues and better know which parts of the code are affected primarily.

jleft:
@jdhodges431 - thanks for looking into this. I remember there was a spreadsheet which had some results for profiling with different size datasets (I think)? Anything useful/interesting come out of this?

jdhodges431:
The profiling results seemed to show a logarithmic complexity w.r.t the dataset size:

![image](https://cloud.githubusercontent.com/assets/14218569/12445148/9130b3f2-bf5a-11e5-9a83-23ddeb7a14a0.png)

jleft:
Thanks, that looks pretty promising! What are the units for the axes?

jdhodges431:
On further inspection it looks linear. Unsure how I got the original information...

![image](https://cloud.githubusercontent.com/assets/14218569/12450095/9909e05c-bf79-11e5-93ae-77a05143bb7d.png)

PR #421, opened 2015-11-20
Refactor data

tlsim:
Here's an attempt to refactor some of the data code, for #332 

* Makes `coinbase.webSocket` interface use d3 events rather than a callback. This is closer to the native `WebSocket` interface.
* Adds a more generic component for making OHLC buckets from a component which produces trade data (in our case the webSocket).
* Coinbase webSocket instance is now in dataInterface, which might help towards improved handling of ws errors / status.


src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,55 @@
...
-            dataInterface.invalidate();
-            historicFeed.granularity(period);
-            ohlcConverter.period(period);
-            updateHistoricFeedDateRangeToPresent(period);
-            var currentData = [];
             historicFeed(callbackGenerator(function(err, data) {
                 if (!err) {
-                    currentData = data.reverse();
-                    ohlcConverter(liveCallback(currentData), currentData[currentData.length - 1]);
+                    var currentData = d3.map(data, function(d) {return d.date; });
```
tlsim: I'm wondering if the conversion to d3.map and sorting by date of the returned data should be internal to `collectOhlc` so that we can just pass arrays around?

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,125 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.data.collectOhlc = function() {
```
tlsim: Any suggestions for a better name would be appreciated!

src/assets/js/app.js
```diff
@@ -195,29 +195,25 @@
 
         function initialiseDataInterface() {
             return sc.data.dataInterface()
```
jleft: Nice - this looks much tidier :)

src/assets/js/data/collectOhlc.js
jleft: Naming is hard :( I can't think of a better name at the moment!

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,125 @@
...
+
+        var granularity = 60;
+        var tradeSource = sc.data.feed.coinbase.webSocket();
+        var tradeEventType = 'message';
+        var openEventType = 'open';
+        var price = function(d) { return +d.price; };
```
jleft: Stylistically I prefer `Number(d.price)` - I think its intent is a bit more obvious - not too fussed either way though.

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,125 @@
...
+        var volume = function(d) { return +d.volume; };
+        var date = function(d) { return d.date; };
+
+        var data = d3.map();
+        var event = d3.dispatch('update');
```
jleft: We tend to call this `dispatch`, rather than `event` - I think we should be consistent. I don't have a strong preference about which name we use though.

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,125 @@
...
+
+        var updateOhlc = function(trade) {
```
jleft: Prefer named function:

```js
function updateOhlc(trade) { ...
```

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,125 @@
...
+                    high: tradePrice,
+                    low: tradePrice,
+                    close: tradePrice,
+                    volume: tradeVolume
+                });
+            }
+            event.update(data.values());
+        };
+
+        var getBucketStart = function(date) {
```
jleft: Prefer named function:

```js
function getBucketStart(date) { ...
```

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,55 @@
     sc.data.dataInterface = function() {
         var historicFeed = fc.data.feed.coinbase();
         var callbackGenerator = sc.data.callbackInvalidator();
-        var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
+        var collectOhlc = sc.data.collectOhlc()
+          .date(function(d) {return new Date(d.time);})
+          .volume(function(d) {return +d.size;});
         var dataGenerator = fc.data.random.financial();
-        var dispatch = d3.dispatch(sc.event.messageReceived, sc.event.dataLoaded);
+        var coinbaseWebSocket = sc.data.feed.coinbase.webSocket();
```
jleft: It's interesting that both `dataInterface` and `collectOhlc` have an instance of the websocket by default.

I think it would be possible to create an instance of `collectOhlc` without passing it a websocket and it would work, but we'd wouldn't be able to subscribe to the websocket error event.

A few alternatives:
+ Not have a default websocket instance created in `collectOhlc` (it blows up if we don't pass it a websocket) - could possibly add this as a param to the factory function?
+ Or, remove the instance in `dataInterface` and expose the websocket error event through `collectOhlc`.

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,55 @@
...
+        var dispatch = d3.dispatch(sc.event.dataUpdate, sc.event.dataLoaded,
+          sc.event.dataLoadError, sc.event.webSocketError);
         var candlesOfData = 200;
```
jleft: Maybe something which could be configurable, with #406 in mind? Probably not needed now though...

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,55 @@
...
+            collectOhlc.granularity(granularity);
+            collectOhlc.tradeSource(coinbaseWebSocket);
 
-        function dataInterface(period) {
-            dataInterface.invalidate();
-            historicFeed.granularity(period);
-            ohlcConverter.period(period);
-            updateHistoricFeedDateRangeToPresent(period);
-            var currentData = [];
             historicFeed(callbackGenerator(function(err, data) {
```
jleft: I'd update `err` -> `error` for consistency

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,55 @@
...
+            });
+            coinbaseWebSocket.on('error', function(error) {
+                // Note that the 'close' event is potentially more useful for error info.
+                dispatch[sc.event.webSocketError](error);
+            });
+            collectOhlc();
         }
 
+        // TODO: remove? - #407
         dataInterface.generateDailyData = function() {
```
jleft: I've never been to keen on this function, just an idea...

Maybe a 'source' param could be added to dataInterface:

 ```js
dataInterface(source, granularity) {
if (source === sc.data.source.generator) { // (could have an object, like sc.event, which stores our data source types)
    // generate fake data
} else if (source === sc.data.source.coinbase) {
    // bitcoin data...
} else {
    // Throw error
}
```

Any thoughts?

File: src/assets/js/data/feed/coinbase/webSocket.js [hunk 12]
```diff
@@ -1,56 +1,57 @@
...
+        var socket;
 
-        function webSocket(cb) {
-            webSocket.close();
-            coinbaseSocket = new WebSocket('wss://ws-feed.exchange.coinbase.com');
-            var msg = {
-                type: 'subscribe',
+        var webSocket = function() {
+            var url = 'wss://ws-feed.exchange.coinbase.com';
+            var subscribe = {
```
jleft: Could move `url` and `subscribe` out of this function and into the factory function?

src/assets/js/data/feed/coinbase/webSocket.js
```diff
@@ -1,56 +1,57 @@
...
+                'type': 'subscribe',
                 'product_id': product
             };
 
-            coinbaseSocket.onopen = function(event) {
-                coinbaseSocket.send(JSON.stringify(msg));
-                cb(event, null);
-            };
+            socket = new window.WebSocket(url);
```
jleft: Just curious why `window.WebSocket` instead of `WebSocket`?

src/assets/js/data/feed/coinbase/webSocket.js
```diff
@@ -1,56 +1,57 @@
...
-                    cb(event, datum);
-                }
+            socket.onopen = function(event) {
+                socket.send(JSON.stringify(subscribe));
+                d3event.open(event);
             };
-
-            coinbaseSocket.onerror = function(event) {
-                cb(event, null);
+            socket.onerror = function() {
```
jleft: Do we get an event object with onerror? If we don't, then I'm not sure if there's much value in sending the string with the event.

src/assets/js/event.js
```diff
@@ -4,8 +4,10 @@
     sc.event = {
         crosshairChange: 'crosshairChange',
         viewChange: 'viewChange',
-        messageReceived: 'messageReceived',
+        dataUpdate: 'dataUpdate',
```
jleft: Just wondering if the difference between the names of `dataUpdate` and `dataLoaded` is a bit subtle? Can't really think of a better name at the moment, but maybe `dataUpdate` could be something like `newTrade`?

src/assets/js/data/feed/coinbase/webSocket.js
```diff
@@ -1,56 +1,57 @@
...
     'use strict';
+
+    // https://docs.exchange.coinbase.com/#websocket-feed
+
     sc.data.feed.coinbase.webSocket = function() {
+
         var product = 'BTC-USD';
-        var msgType = 'match';
-        var coinbaseSocket = null;
+        var d3event = d3.dispatch('open', 'close', 'error', 'message');
```
jleft: Maybe `d3event` -> `dispatch`?

jleft:
Looking good, thanks @tlsim - definitely looks like a step in the right direction.

src/assets/js/data/dataInterface.js
tlsim: I like option 1 - will go with that.

src/assets/js/data/dataInterface.js
tlsim: Yeah this needs more thought. A nicer interface for this component would include configuration for `start`, `end` and `granularity`, with the number of 'candles' dependent on those params. The historic feed has a limit of 200 candles returned per request however, so we'll need a better way to handle this limitation.

src/assets/js/data/feed/coinbase/webSocket.js
tlsim: Some linting tool in some other project complained :) I'll change to just `WebSocket`.

src/assets/js/data/feed/coinbase/webSocket.js
tlsim: Not sure. [Docs](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) say 'a simple event named error', giving something like `{ "type": "error" }`?

I'm also using this event to propagate the message received with type 'error' https://docs.exchange.coinbase.com/#error

Would it be acceptable for the d3 'error' event to respond with either type?

src/assets/js/data/dataInterface.js
tlsim: I agree. a configurable source is better.

Afredon: When I was splitting the products, I thought ultimately we could also store the source/provider in the objects to remove that branching in your example. Not sure if we need at product or product x period level though. I imagine it is possible to get the 5 seconds feed from one provider and the 1 day feed from another.

src/assets/js/data/feed/coinbase/webSocket.js
tlsim: I don't see this component as reusable with anything other than the coinbase API. I'll make this change but I think it makes sense to keep the defaults.

src/assets/js/data/dataInterface.js
tlsim: I'll hold off making any change for now so we can discuss more.

src/assets/js/data/feed/coinbase/webSocket.js
jleft: Ah, sorry - I wasn't thinking of this change with a view to making them configurable; just thought it would look a little neater. I also don't see it being configurable, so would be happy to just use the default currently there.

rjmcneill:
LGTM :+1: 
Going to be fun integrating it with the coinbase product stuff I've been doing :stuck_out_tongue_closed_eyes:

src/assets/js/data/collectOhlc.js
```diff
@@ -0,0 +1,118 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.data.collectOhlc = function() {
+
+        var granularity = 60;
+        var tradeEventType = 'message';
+        var openEventType = 'open';
```
jleft: Unused?

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,54 @@
     sc.data.dataInterface = function() {
         var historicFeed = fc.data.feed.coinbase();
         var callbackGenerator = sc.data.callbackInvalidator();
-        var ohlcConverter = sc.data.feed.coinbase.ohlcWebSocketAdaptor();
         var dataGenerator = fc.data.random.financial();
-        var dispatch = d3.dispatch(sc.event.messageReceived, sc.event.dataLoaded);
+        var coinbaseWebSocket = sc.data.feed.coinbase.webSocket();
```
jleft: I'm still struggling with the relationship between the WebSocket and `collectOhlc`; I find it a bit confusing having the instance of the WebSocket both here, in `dataInterface`, and in `collectOhlc`.

For example, a listener for the WebSocket's errors is registered in `dataInterface`, but we close the WebSocket by calling `collectOhlc.close()`, when we could just call `coinbaseWebSocket.close()` in `dataInterface`?

I wonder if `collectOhlc` even needs to know about the WebSocket? The WebSocket could be created in data interface and event listeners set up there, then whenever a message is received, it could be passed to `collectOhlc` which would bucket the trade and return the updated data? Then `collectOhlc` would only have a single responsibility of bucketing trades. (This might also remove the need to use events in `collectOhlc`).

Happy to discuss :)

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,54 @@
...
+        var collectOhlc = sc.data.collectOhlc(coinbaseWebSocket)
```
jleft: `coinbaseWebSocket` is being passed in here, to the factory function, but there is no matching param on `sc.data.collectOhlc`. Also see line 49, where the same websocket instance is being passed into the component.

Perhaps just passing it into the factory function would work? (If we are to take this approach - see comment above.)

src/assets/js/data/dataInterface.js
```diff
@@ -4,52 +4,54 @@
...
+                } else {
+                    dispatch[sc.event.dataLoadError](error);
                 }
-                dispatch[sc.event.dataLoaded](err, currentData);
             }));
+            collectOhlc.on('update', function(data) {
+                dispatch[sc.event.newTrade](dateSort(data));
+            });
+            coinbaseWebSocket.on('error', function(error) {
+                // Note that the 'close' event is potentially more useful for error info.
```
jleft: If the close event is also useful, should we propagate it through `dataInterface` to app?

tlsim: I think that would make sense. #446

src/assets/js/data/dataInterface.js
tlsim: Nice idea! - I'll update.

src/assets/js/data/dataInterface.js
```diff
@@ -4,67 +4,75 @@
...
-            dispatch[sc.event.dataLoaded](null, dataGenerated);
+            var data = dataGenerator(candlesOfData);
+            dispatch[sc.event.dataLoaded](data);
         };
 
-        dataInterface.invalidate = function() {
-            ohlcConverter.close();
+        function invalidate() {
+            coinbaseWebSocket.close();
+            data = [];
```
tlsim: This array is needed in order to update the data when a new trade arrives, now that `collectOhlc` does not depend on `coinbaseWebsocket`. I'm seeing when switching between periods (above the Coinbase rate limit) that it can become inconsistent (wrong period, empty array). The chart doesn't cope well with trying to display an empty data array - we get console errors with invalid `<path>` attributes.

jleft:
Thanks @tlsim - looks like a great improvment. Happy to see this merged now and approach the other aspects discussed later (Quandl, adapting the data generator to have an interface consistent with the other data sources etc.).

:+1: for tests too!

Rebasing and merging in #455

PR #427, opened 2015-11-24
Cache node_modules on Travis

jleft:
Another speed improvement for the build (only relevant when run on Travis)


djmiley:
LGTM

jleft:
I've read in a few places that this can be more trouble than it's worth; there can be issues when dep versions are updated.

I'm going to close this now, but happy to reopen if anyone is keen to try this.

PR #428, opened 2015-11-24
Small fixes for IE and FF

### LINKED ISSUE #422: Primary chart render console error in IE
`Unable to get property 'datum' of undefined or null reference`

```
.on('trackingmove', function(crosshairData) {
    dispatch.crosshairChange(crosshairData[0].datum); // <- here
})
```

Afredon:
Fixes #422 and an issue with the navbar vertical bars width in Firefox

### DIFF
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -22,14 +22,12 @@
         var navMulti = fc.series.multi()
             .series([area, line, brush])
             .decorate(function(selection) {
-                var eastResizeRect = selection.select('.resize.e>rect');
                 var westResizeRect = selection.select('.resize.w>rect');
+                var eastResizeRect = selection.select('.resize.e>rect');
                 westResizeRect.attr('width', '2px')
-                    .attr('x', '-2')
-                    .attr('stroke-width', '1px');
+                    .attr('x', '-1');
                 eastResizeRect.attr('width', '2px')
-                    .attr('x', '-2')
-                    .attr('stroke-width', '1px');
+                    .attr('x', '-1');
 
                 selection.enter()
                     .select('.e')
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -22,33 +22,34 @@
         var navMulti = fc.series.multi()
             .series([area, line, brush])
             .decorate(function(selection) {
-                var westResizeRect = selection.select('.resize.w>rect');
-                var eastResizeRect = selection.select('.resize.e>rect');
-                westResizeRect.attr('width', '2px')
-                    .attr('x', '-1');
-                eastResizeRect.attr('width', '2px')
-                    .attr('x', '-1');
+                var enter = selection.enter();
 
-                selection.enter()
-                    .select('.e')
+                // overload d3 styling for the brush sides
+                // as Firefox does not react properly to setting these through less file.
+                var westResizeRect = enter.select('.resize.w>rect');
+                var eastResizeRect = enter.select('.resize.e>rect');
+                westResizeRect.attr('width', 2)
+                    .attr('x', -1);
+                eastResizeRect.attr('width', 2)
+                    .attr('x', -1);
+
+                // Adds the handles to the brush sides
+                enter.select('.e')
                     .append('circle')
                     .attr('cy', 29)
                     .attr('r', 7)
                     .attr('class', 'outer-handle');
-                selection.enter()
-                    .select('.e')
+                enter.select('.e')
                     .append('circle')
                     .attr('cy', 29)
                     .attr('r', 4)
                     .attr('class', 'inner-handle');
-                selection.enter()
-                    .select('.w')
+                enter.select('.w')
                     .append('circle')
                     .attr('cy', 29)
                     .attr('r', 7)
                     .attr('class', 'outer-handle');
-                selection.enter()
-                    .select('.w')
+                enter.select('.w')
                     .append('circle')
                     .attr('cy', 29)
                     .attr('r', 4)
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -109,7 +109,7 @@
 
                 setHide(selection, brushHide);
 
-                if (brush.extent()[0][0] - brush.extent()[1][0] !== 0) {
+                if (!brush.empty()) {
                     dispatch[sc.event.viewChange]([brush.extent()[0][0], brush.extent()[1][0]]);
                 }
             })
@@ -117,7 +117,7 @@
                 brushHide = false;
                 setHide(selection, brushHide);
 
-                if (brush.extent()[0][0] - brush.extent()[1][0] === 0) {
+                if (brush.empty() === 0) {
                     dispatch[sc.event.viewChange](sc.util.domain.centerOnDate(viewScale.domain(),
                         model.data, brush.extent()[0][0]));
                 }
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -22,6 +22,15 @@
         var navMulti = fc.series.multi()
             .series([area, line, brush])
             .decorate(function(selection) {
+                var eastResizeRect = selection.select('.resize.e>rect');
+                var westResizeRect = selection.select('.resize.w>rect');
+                westResizeRect.attr('width', '2px')
+                    .attr('x', '-2')
+                    .attr('stroke-width', '1px');
+                eastResizeRect.attr('width', '2px')
+                    .attr('x', '-2')
+                    .attr('stroke-width', '1px');
+
                 selection.enter()
                     .select('.e')
                     .append('circle')
@@ -128,7 +137,8 @@
                     dispatch[sc.event.viewChange](domain);
                 });
 
-            navbarContainer.select('.plot-area').call(zoom);
+            navbarContainer.select('.plot-area')
+                .call(zoom);
         }
 
         d3.rebind(nav, dispatch, 'on');
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -40,37 +40,22 @@
                     .attr('height', extentHeight)
                     .attr('y', backgroundStrokeWidth / 2);
 
-                // overload d3 styling for the brush sides
+                // overload d3 styling for the brush handles
                 // as Firefox does not react properly to setting these through less file.
-                var westResizeRect = enter.select('.resize.w>rect');
-                var eastResizeRect = enter.select('.resize.e>rect');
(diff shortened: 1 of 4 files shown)


src/assets/js/chart/nav.js
```diff
@@ -22,6 +22,13 @@
         var navMulti = fc.series.multi()
             .series([area, line, brush])
             .decorate(function(selection) {
+                var westResizeRect = selection.select('.resize.w>rect');
+                var eastResizeRect = selection.select('.resize.e>rect');
```
jleft: Can we use the enter selection for both the east and west? e.g.

```js
var eastResizeRect = selection.enter()
    .select('.resize.e>rect');
```

src/assets/js/chart/nav.js
```diff
@@ -22,6 +22,13 @@
...
+                westResizeRect.attr('width', '2px')
```
jleft: Do we need the `px`? I thought this was only relevant when setting a style, rather than an attr?

src/assets/js/chart/nav.js
```diff
@@ -22,6 +22,13 @@
...
+                    .attr('x', '-1');
```
jleft: Should we also be adjusting the `y` coordinate? Does it need to be shifted by `1` or `2`?

Afredon: no, it's just to override d3 which sets width to 6 and x to -3. I didn't notice anything on y, but if I missed that yes we can adjust.

src/assets/js/chart/primary.js
```diff
@@ -84,7 +84,9 @@
              .xLabel('')
              .yLabel('')
              .on('trackingmove', function(crosshairData) {
-                 dispatch.crosshairChange(crosshairData[0].datum);
+                 if (crosshairData.length > 0) {
```
jleft: I'd be tempted to let app.js handle this; we could simplify primary to just:

```js
.on('trackingmove', dispatch.crosshairChange)
.on('trackingend', dispatch.crosshairChange);
```

And update app.js to check if we have data or not, something like this:

```js
function onCrosshairChange(data) {
    var dataPoint = data.length ? data[0].datum : null;
    legendModel.data = dataPoint;
    render();
}
```

src/assets/js/chart/nav.js
Afredon: I think we don't need it. I took the habit to put it when talking about sizes, but yeah for an attribute it seems to make no difference.

src/assets/js/chart/nav.js
jleft: I was just basing this on the removed style - `transform: translate(2px, 2px);` - made me wonder if we needed to shift the y too.

src/assets/js/chart/nav.js
Afredon: Ah that should work yes. but we would need to save it. 
```var enter = selection.enter()
enter.select(...```

Comment (jleft) [inline src/assets/js/chart/nav.js]: Yup, good idea, that's worth doing, especially as it's used for the circles too.

File: src/assets/js/chart/primary.js (hunk 4 above)
Comment (Afredon) [inline src/assets/js/chart/primary.js]: I'd be glad to do so, it's closer to the separation of things I prefer.
I think we need to skip the render() when there is no data though, I'll double check.

File: src/assets/js/chart/nav.js (hunk 3 above)
Comment (Afredon) [inline src/assets/js/chart/nav.js]: Ah, I was wondering why there was a y component actually. I didn't see a need for it.

Comment (jleft) [inline src/assets/js/chart/nav.js]: I'll check the code out and have a look. I expect that if the `rects` aren't looking quite right, then `shape-rendering="crispEdges"` might improve how they look.

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering

Comment (Afredon) [inline src/assets/js/chart/nav.js]: Ok I need to do something to y too, you're right.
This is what is in d3:
```js
.append("rect")
          .attr("x", function(d) { return /[ew]$/.test(d) ? -3 : null; })
          .attr("y", function(d) { return /^[ns]/.test(d) ? -3 : null; })
```

so a .y(0) might be necessary

Afredon: Ahah I almost removed the crispEdges as it generates issues too from what I read in some fora. But I managed to get a correct appearance on most cases without it. Also it isn't supported by Edge, so we can't rely on it anyway.

src/assets/js/chart/nav.js
Afredon: done.

src/assets/js/chart/nav.js
Afredon: px removed.

src/assets/js/chart/primary.js
Afredon: On second thought, I'm not really happy either with the fact has to know the internals of the crosshair to know what to do, while the current solution keep the change invisible to app.js
Alternative would be to send an undefined point when data.length === 0 in onTrackingMove and let app.js decide what to do with that, instead of blocking the event entirely

src/assets/js/chart/nav.js
```diff
@@ -22,26 +22,34 @@
...
-                selection.enter()
-                    .select('.e')
+                var enter = selection.enter();
+
+                // overload d3 styling for the brush sides
+                // as Firefox does not react properly to setting these through less file.
+                var westResizeRect = enter.select('.resize.w>rect');
+                var eastResizeRect = enter.select('.resize.e>rect');
+                westResizeRect.attr('width', 2)
+                    .attr('x', -1);
```
jleft: After checking this code out, I think the the y-positioning does need adjusting, it seems these `rect` elements are a pixel higher than they were previously:

brush-rect-y

src/assets/js/chart/nav.js
```diff
@@ -110,7 +118,7 @@
                 brushHide = false;
                 setHide(selection, brushHide);
 
-                if (brush.extent()[0][0] - brush.extent()[1][0] === 0) {
+                if (brush.empty() === 0) {
```
jleft: Does this work? I thought `brush.empty()` would only return `true` or `false`?

Afredon: Mhh, it works as true === 0, but it's unintentional. I didn't replace all I want.

src/assets/js/chart/nav.js
Afredon: 2 pixels as it was before are necessary indeed. I set the height at 60 and it ends with only 58.
I'd like to understand why though.

Afredon: That put us with the exact same effect than the transform(2,2) from previously, as we replaced the default x-3 by -1, and put back the y+2

src/assets/js/chart/primary.js
jleft: I'm not too keen on primary changing arguments which are sent with the event. I think it could be hard to understanding what's happening if an event is propagating through many components and the components are changing the arguments.

Is there any issue with letting the legend handle the crosshair data? I can't see a reason not to pass it all of the crosshair data, it would definitely simplify primary and app too:

primary:

```js
.on('trackingmove', dispatch.crosshairChange)
.on('trackingend', dispatch.crosshairChange)
```

app:

```js
function onCrosshairChange(data) {
    legendModel.data = data;
    render();
}
```

src/assets/js/chart/nav.js
jleft: `true === 0 // is false`

src/assets/js/chart/primary.js
Afredon: Technically no issue, my problem here is coupling. Legend will require to receive data formatted by crosshair instead of just a single point which would have been more independent from something else.

src/assets/js/chart/nav.js
Afredon: oops, yes. So there was no logical reason it was working, which raises new questions as to why it was still apparently working.

src/assets/js/chart/primary.js
Afredon: Thinking on it again, the coupling is a problem. There is no reason why Legend should receive an array of data instead of a single point. 
For primary changing the arguments sent, I'm not sure I would count that as a change. It's merely saying if there is no data, send undefined instead of digging into the non existing data. trackingend is doing something similar by replacing whatever data we have by undefined.

src/assets/js/chart/nav.js
Afredon: Fixed

src/assets/js/chart/nav.js
Afredon: new update details what's going on and solve issues. It should be nice on all browsers now.

src/assets/js/chart/nav.js
```diff
@@ -22,28 +34,44 @@
...
+                var enter = selection.enter();
+
+                selection.select('.extent')
+                    .attr('height', extentHeight)
+                    .attr('y', backgroundStrokeWidth / 2);
+
+                // overload d3 styling for the brush sides
+                // as Firefox does not react properly to setting these through less file.
+                var westResizeRect = enter.select('.resize.w>rect');
+                var eastResizeRect = enter.select('.resize.e>rect');
```
jleft: Just FYI, an alternaitve, which is a bit more concise, is using a `,` to select mutliple elements:

```js
 // N.B. selectAll() used over select() to get both elements
enter.selectAll('.resize.w>rect, .resize.e>rect')
    .attr('width', handleBarWidth)
    .attr('x', -handleBarWidth / 2)
    .attr('height', barHeight)
    .attr('y', borderWidth);
```

src/assets/js/chart/nav.js
```diff
@@ -22,28 +34,44 @@
...
+                westResizeRect.attr('width', handleBarWidth)
+                    .attr('x', -handleBarWidth / 2)
+                    .attr('height', barHeight)
```
jleft: If we want to override the `height`, then it needs to be done on update, not enter; otherwise, the D3 brush will overwrite our value after the initial render (they're only only pixel different - `59` vs `60`).

https://github.com/mbostock/d3/blob/5b981a18db32938206b3579248c47205ecc94123/src/svg/brush.js#L142

jleft:
Thanks @Afredon - just a couple more comments, then this is good to go :)

src/assets/js/chart/nav.js
Afredon: Ah thanks, I didn't know we could do that, it's definitely better

src/assets/js/chart/nav.js
Afredon: Good catch, I completely missed that!

jleft:
Thanks @Afredon, once this is rebased, I'm happy to merge.

PR #434, opened 2015-11-27
Added overlay for components that sit on top of chart. Moved legend i…

jdhodges431:
…nto overlay and added edit indicator elements.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -23,6 +23,7 @@ export default function() {
         xAxis: chartsContainer.select('#x-axis-container'),
         navbar: chartsContainer.select('#navbar-container'),
         overlay: overlay,
+        overlaySecondaries: overlay. selectAll('.overlay-secondary-container'),
         legend: appContainer.select('#legend'),
         suspendLayout: function(value) {
             var self = this;
@@ -74,6 +75,7 @@ export default function() {
     var navResetModel = model.chart.navigationReset();
     var headMenuModel = model.menu.head([generated, bitcoin], generated, day1);
     var legendModel = model.chart.legend(generated, day1);
+    var overlayModel = model.menu.overlay();
 
     var charts = {
         primary: undefined,
@@ -127,10 +129,7 @@ export default function() {
             .call(selectors);
 
         containers.overlay
-            .datum({
-                primaryModel: primaryChartModel,
-                secondaryModel: charts.secondaries
-            })
+            .datum(overlayModel)
             .call(overlay);
 
         if (layoutRedrawnInNextRender) {
@@ -169,6 +168,34 @@ export default function() {
         secondaryChartModel.trackingLatest = trackingLatest;
         navModel.trackingLatest = trackingLatest;
         navResetModel.trackingLatest = trackingLatest;
+        render();
+    }
+
+    function onPrimaryIndicatorChange(indicator) {
+        indicator.isSelected = !indicator.isSelected;
+        primaryChartModel.indicators =
+            selectorsModel.indicatorSelector.indicatorOptions.filter(function(option) {
+                return option.isSelected;
+            });
+        overlayModel.primaryIndicators = primaryChartModel.indicators;
+        render();
+    }
+
+    function onSecondaryChartChange(_chart) {
+        _chart.isSelected = !_chart.isSelected;
+        charts.secondaries =
+            selectorsModel.indicatorSelector.secondaryChartOptions.filter(function(option) {
+                return option.isSelected;
+            });
+        // TODO: This doesn't seem to be a concern of menu.
+        charts.secondaries.forEach(function(chartOption) {
+            chartOption.option.on(event.viewChange, onViewChange);
+        });
+        overlayModel.secondaryIndicators = charts.secondaries;
+
+        // TODO: Remove .remove! (could a secondary chart group component manage this?).
+        containers.secondaries.selectAll('*').remove();
+        updateLayout();
         render();
     }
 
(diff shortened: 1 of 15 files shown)


jdhodges431:
closes #416

jdhodges431:
Currently rebasing this with develop so dropdowns are displayed in the overlay.

src/assets/js/app.js
```diff
@@ -7,6 +7,7 @@
```
jleft: Nice :) this looks well organised and easy to follow

src/assets/js/app.js
```diff
@@ -117,6 +121,10 @@
                 .datum(selectorsModel)
                 .call(selectors);
 
+            containers.overlay
+                .datum(overlayModel)
```
jleft: Move one line up: `containers.overlay.datum(overlayModel)`

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.menu.generator.editIndicatorGroup = function() {
+        var dispatch = d3.dispatch('click');
```
jleft: As this is a mechanism for exposing the event to other components in the application, rather than dispatching an event called `click`, we can dispatch an event which is named in a more descriptive and meaningful way. Perhaps `indicatorChange` would work? 

Also worth looking at event.js - the string which represents this event can be stored there: the dispatch would then look something like:

```js
var dispatch = d3.dispatch(sc.event.indicatorChange);
```

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+
+        function editIndicatorGroup(selection) {
+            selection.each(function(model) {
+                var selection = d3.select(this);
+
+                var div = selection
+                    .selectAll('div')
+                    .data(model.selectedIndicators);
```
jleft: I'm not sure that we should be keying on index here (which is the default behaviour), would it make sense to key on `valueString`?

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+
+                div.enter()
+                    .append('div')
+                    .attr('class', 'edit-indicator')
+                    .each(function(d) {
```
jleft: Can we avoid using `each` in favour of a more 'fluent' style?

```js
// Similar to this (not sure if it would actually work!)
var containersEnter = div.enter()
    .append('div')
    .attr('class', 'edit-indicator');

containersEnter.append('span')
    .attr('class', 'icon sc-icon-delete')
    .on('click', dispatch.click);

containersEnter.append('span')
    .attr('class', 'indicator-label');

// etc.
```

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+                        selection.append('span')
+                            .attr('class', 'indicator-label');
+                    });
+
+                div.select('.indicator-label')
+                    .text(function(d) {
+                        return d.displayString;
+                    });
+
+                div.select('.icon').on('click', dispatch.click);
```
jleft: Can the click event handler just be added on the `enter()` selection?

`.on(...` on a new line

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+
+                div.exit().remove();
```
jleft: `.remove()` on new line

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+            });
+        }
+
+        d3.rebind(editIndicatorGroup, dispatch, 'on');
+        return editIndicatorGroup;
```
jleft: Newline before `return editIndicatorGroup;`

src/assets/js/menu/generator/editIndicatorGroup.js
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+})(d3, fc, sc);
```
jleft: Add newline to end of file (this will be enforced when ESLint lands!)

src/assets/js/util/getSVG.js
```diff
@@ -0,0 +1,9 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.util.getSVG = function(path, callback) {
```
jleft: Remove - no longer needed now that we have the SVG sprites.

src/assets/styles/overlay.less
```diff
@@ -0,0 +1,81 @@
+#overlay {
```
jleft: This is similar to `flex.less` - can we create some common classes to reduce duplication? For example:

flex.less

```less
#x-axis-row {
  flex: 0 0 @x-axis-height;
  display: flex;
  flex-direction: row;
}
```

overlay.less

```less
#overlay-x-axis-row {
  flex: 0 0 20px;
  display: flex;
  flex-direction: row;
}
```

We could create a single class to be shared amongst both:

```less
.x-axis-row {
  flex: 0 0 @x-axis-height;
  display: flex;
  flex-direction: row;
}
```

src/assets/styles/overlay.less
```diff
@@ -0,0 +1,81 @@
...
+  #overlay-primary-container {
+    flex: 5;
+    display: flex;
+    flex-direction: column;
+  }
+
+  #overlay-head {
+    display: flex;
+    flex-direction: column;
+    flex: 0 0 40px
```
jleft: Could `@head-height` be useful?

src/assets/styles/overlay.less
```diff
@@ -0,0 +1,81 @@
...
+
+  .overlay-secondary-container {
+    flex: 0;
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+  }
+
+  #overlay-x-axis-row {
+    flex: 0 0 20px;
```
jleft: Can use a variable: `flex: 0 0 @x-axis-height`

src/assets/styles/overlay.less
```diff
@@ -0,0 +1,81 @@
...
+    display: flex;
+    flex-direction: row;
+  }
+
+  #overlay-navbar-container {
+    flex: 0 0 100px;
```
jleft: Can use a variable: `flex: 0 0 @navigator-height`

jleft:
Rebased in #471

PR #435, opened 2015-11-30
Add dropdown selectors for series and indicators; add SVG icon sprite generation

### LINKED ISSUE #316: Drop-downs/components

### LINKED ISSUE #318: Remove slide-out menu from mobile version

### LINKED ISSUE #320: Series/indicator buttons
No need to add the icons for the first iteration - this can be done later.

Feel free to split this into two issues (series buttons/dropdown, indicators buttons/dropdown), if it helps.

jleft:
- Resolves #316
- Resolves #318
- Resolves #320 
- Closes #387 (rebased and integrated SVG sprite generation)


src/index.html.tpl
```diff
@@ -12,23 +12,22 @@
...
+            <div id="selectors">
+                <div id="series-dropdown" class="dropdown selector-dropdown"></div>
+                <div id="indicator-dropdown" class="dropdown selector-dropdown"></div>
+            </div>
             <div id="legend"></div>
             <div id="charts-container">
                 <svg id="primary-container"></svg>
                 <svg class="secondary-container"></svg>
                 <svg class="secondary-container"></svg>
-                <svg class="secondary-container"></svg>
```
djmiley: Why have you removed this?

src/assets/js/app.js
```diff
@@ -59,7 +59,7 @@
 
         var primaryChartModel = sc.model.primaryChart(generated);
         var secondaryChartModel = sc.model.secondaryChart(generated);
-        var sideMenuModel = sc.model.menu.side();
+        var selectorsModel = sc.model.menu.selectors();
```
djmiley: Could there be a more expressive name than `selectors` ? I'm not entirely sure what it would be, maybe `displaySelectors` ? 
My main concern is that `selectors` does not imply to me that we don't select the product in this menu, which may also motivate the renaming of `headMenu` possibly?

djmiley:
Looks good, just a couple of thoughts about it.

src/index.html.tpl
jdhodges431: The volume chart is absent from the design spec and we don't have an icon for it yet. Might be worth getting in touch with the UX person about it.

jleft: Yeah, I was going to suggest creating a separate issue to add the volume chart back in.

src/assets/js/app.js
jleft: I agree, I'm not convinced by the naming either. It was discussed here: https://github.com/ScottLogic/d3fc-showcase/pull/387#discussion_r44795947

I think we could probably do with reviewing the naming, bearing in mind the UX changes, something we could do in a separate issue?

jleft:
> Looks good, just a couple of thoughts about it.

Thanks for the review @djmiley :)

PR #437, opened 2015-11-30
Added further coinbase products

### LINKED ISSUE #67: Add component to retrieve Coinbase product list
https://docs.exchange.coinbase.com/#get-products

rjmcneill:
Closes #67

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -7,6 +7,7 @@ import menu from './menu/menu';
 import util from './util/util';
 import event from './event';
 import dataInterface from './data/dataInterface';
+import coinbaseProducts from './data/feed/coinbase/products';
 
 export default function() {
 
@@ -55,16 +56,11 @@ export default function() {
         seconds: 60,
         d3TimeInterval: {unit: d3.time.minute, value: 1},
         timeFormat: '%H:%M'});
-
     var generated = model.data.product({
+        family: 'generated',
         display: 'Data Generator',
         volumeFormat: '.3s',
         periods: [day1]
-    });
-    var bitcoin = model.data.product({
-        display: 'Bitcoin',
-        volumeFormat: '.2f',
-        periods: [minute1, minute5, hour1]
     });
 
     var primaryChartModel = model.chart.primary(generated);
@@ -73,7 +69,7 @@ export default function() {
     var xAxisModel = model.chart.xAxis(day1);
     var navModel = model.chart.nav();
     var navResetModel = model.chart.navigationReset();
-    var headMenuModel = model.menu.head([generated, bitcoin], generated, day1);
+    var headMenuModel = model.menu.head([generated], generated, day1);
     var legendModel = model.chart.legend(generated, day1);
     var overlayModel = model.menu.overlay();
 
@@ -269,9 +265,10 @@ export default function() {
                 loading(true);
                 updateModelSelectedProduct(product.option);
                 updateModelSelectedPeriod(product.option.periods[0]);
-                if (product.option === bitcoin) {
+                if (product.option.family === 'bitcoin') {
+                    _dataInterface.setNewProduct(product.option.display);
                     _dataInterface(product.option.periods[0].seconds);
-                } else if (product.option === generated) {
+                } else if (product.option.family === 'generated') {
                     _dataInterface.generateDailyData();
                 }
                 render();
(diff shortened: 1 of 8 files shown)


src/assets/js/menu/head.js
```diff
@@ -20,8 +20,16 @@
         var head = function(selection) {
             selection.each(function(model) {
                 var selection = d3.select(this);
-
                 var products = model.products;
+                var selectedIndex = -1;
+
+                for (var i = 0; i < products.length; i++) {
```
jdhodges431: I've done something similar with map and indexof elsewhere. In that style it would look like this:

    var selectedIndex = products.map(function(product) {
        return product.display === model.selectedProduct.display;
    }).indexOf(true);

Wondering which is preferable.

Afredon: I don't see where selectedIndex is used, which would make irrelevant the best way to calculate it.

rjmcneill: Good spot @jdhodges431 
I accidentally pushed this back.
I had this debug code in to allow me to get coding on the rest of the stuff, and forgot to take it out.
Replaced this with:

``` JavaScript
selectedIndex: products.indexOf(model.selectedProduct)
```
Later on in the file.  My bad!

src/assets/js/data/feed/coinbase/products.js
```diff
@@ -0,0 +1,30 @@
+(function(d3, fc, sc) {
+    'use strict';
+
+    sc.data.feed.coinbase.products = function(minute1, minute5, hour1, day1, callback) {
+        var coinbaseProducts = [];
+        var product = null;
+
+        d3.json('https://api.exchange.coinbase.com/products', function(error, reponse) {
```
Afredon: I'm glad you start using French in your code, but wouldn't that be "response" in English ? Otherwise, I'm glad with "Erreur" for the first parameter :)

src/assets/js/app.js
```diff
@@ -235,9 +243,13 @@
         function initialiseHeadMenu(dataInterface) {
             return sc.menu.head()
                 .on(sc.event.dataProductChange, function(product) {
+                    var productSubStr = product.option.display.substring(0,3);
+
                     updateModelSelectedProduct(product.option);
                     updateModelSelectedPeriod(product.option.periods[0]);
-                    if (product.option === bitcoin) {
+
+                    if (productSubStr === 'BTC') {
```
Afredon: You're using the display name to extract an information semantically different. It would be better to explicit the information you are looking for when creating the product, for example in a field named Family or Category, or directly associate with a dataInterface method.

rjmcneill: Yea @Afredon, that does sound like a better way of doing it.
I'll add that to my future update

src/assets/js/data/feed/coinbase/products.js
rjmcneill: Hah, good spot @Afredon :stuck_out_tongue_closed_eyes:

src/assets/js/app.js
```diff
@@ -301,6 +298,23 @@ export default function() {
...
+
+    function insertProductsIntoHeadMenuModel(error, bitcoinProducts) {
+        if (error) {
+            console.log('Error getting coinbase products: ' + error); // TODO: something more useful for the user!
+        }
+        // Add the newly received products to the product list
+        headMenuModel.products = headMenuModel.products.concat(bitcoinProducts);
+        render();
+
+        containers.app.select('.head-menu')
```
tlsim: Can remove this now that `render` is called above.

test/data/coinbaseProductSpec.js
```diff
@@ -0,0 +1,42 @@
...
+var min5 = 300;
+var hr1 = 3600;
+var day1 = 86400;
+
+describe('data/coinbase/products', function() {
+    it('should assign the correct periods to BTC-USD', function() {
+        var btcusd = [{ id: 'BTC-USD' }];
+        var productResponse = formatProducts(min1, min5, hr1, day1, btcusd);
+        expect(productResponse[0].periods.length).toEqual(4);
+        expect(productResponse[0].periods[0]).toEqual(60);
```
tlsim: It's fine to use the variables `min1` etc. inside `toEqual`.

src/assets/js/data/feed/coinbase/products.js
```diff
@@ -0,0 +1,35 @@
...
+    d3.json('https://api.exchange.coinbase.com/products', function(error, response) {
+        if (error) {
+            callback(error);
+        } else {
+            callback(error, formatProducts(minute1, minute5, hour1, day1, response));
+        }
+    });
+}
+
+export function formatProducts(minute1, minute5, hour1, day1, response) {
```
tlsim: I think moving this function to a file of its own is a good idea - it's a bit clearer what is getting tested, and we can be consistent in using `export default` everywhere.

src/assets/js/app.js
rjmcneill: Oops...
Yea, I went back to updating tests when you left, forgot to remove that! :smile:

rjmcneill:
Thanks @tlsim 
Updated after your comments!

tlsim:
Cheers @rjmcneill - Looks good. I moved all the coinbase stuff to /data/coinbase in #477.

rjmcneill:
Makes sense @tlsim, thanks for the update :+1:

Issue #390, opened 2015-11-13
Create top-level renderering components


jleft:
#331 

This should also tidy up `sc.app`.

Current:
![current](https://cloud.githubusercontent.com/assets/2017615/11152422/402d4e12-8a2a-11e5-9483-7b164bd5f4b8.PNG)

Proposed (unsure on naming):
![proposed](https://cloud.githubusercontent.com/assets/2017615/11152425/443ac5fc-8a2a-11e5-8f9a-d92a1dc4712c.PNG)

jleft:
I've self-assigned this for now, but happy to discuss splitting this up into smaller issues if anyone would like to tackle part of this.

PR #439, opened 2015-12-01
Add loading message and hide chart when data is being retrieved

jleft:
I think the loading message could be converted into a component at some point, but I'm not convinced that it's worth doing until #390 is complete.


djmiley:
:+1:

jleft:
Thanks @djmiley @Afredon :)

Issue #410, opened 2015-11-18
Better way to key on renderers


jleft:
Solve the problem that keys are ultimately strings, so we need some way to differentiate between renderers of the same type (e.g. when using multi).

Can't use an index as we would end up reusing containers, which won't work (a component needs to be called on an empty container).

Using identity doesn't work if multiple renderers of the same type are used, because they're the same type.

Look at add an id property to renderers and key on this.

jleft:
Resolved in #441

PR #441, opened 2015-12-01
Refactor models and add UID component to generate IDs for keys

jleft:
Still a WIP, but thought I'd put this up for some feedback. #410 

Some areas to consider:
- Should we be consistent with how we pass configuration values to the models? If so, which approach should we use? Currently, it's either several params (e.g. `sc.model.menu.option`) or a single config object (e.g. `sc.model.data.product`).
- I think adding an `id` property to the renders, so we can key on the `id` is good (or at least better than `isLine`) and should make future work (e.g. additive indicators) easier. Not so keen on the implementation - there is repetition. Perhaps we should make own series/indicator components which wrap the d3fc components and add the `id` to them?
- Should models have default values?
- Should we instantiate series/indicator components outside of `sc.model.whatever`? We could do this in a similar way to period and products, which are instantiated in app.js.


jdhodges431:
We could use config objects if the number of parameters becomes unreasonably long. This doesn't seem to be the case at the moment - could this be an issue in the future?

tlsim:
I'm happy to see this merged now and we can decide on the points raised later.

djmiley:
I agree with @tlsim

jleft:
> We could use config objects if the number of parameters becomes unreasonably long. This doesn't seem to be the case at the moment - could this be an issue in the future?

Yeah, definitely a good idea - a long list of unnamed params is not nice! Agreed, something to bear in mind for the future.

PR #449, opened 2015-12-02
Replace JSHint and JSCS with ESLint

### LINKED ISSUE #447: Add ESLint; remove JSHint and JSCS

jleft:
Resolves #447 

Rules are very similar to d3fc's, with a few slight changes:
+ Warning not error for `no-console` (allow console logs for now)
+ Warning not error for `no-warning-comments` (allow TODO comments for now)
+ Allow `this` to be assigned to `var self`: `"consistent-this": [2, "self"]`
+ Add `sc` as a global
+ Add `window` as a global in _app.js_ and _layout.js_


test/.eslintrc
```diff
@@ -0,0 +1,17 @@
+{
+    "env": {
+        "browser": true
+    },
+    "globals": {
+        "expect": true,
+        "fc": true,
+        "sc": true,
+        "it": true,
+        "describe": true,
```
tlsim: Could you use `env` for the jasmine globals? http://eslint.org/docs/user-guide/configuring.html#specifying-environments

tlsim:
:+1:

test/.eslintrc
jleft: Nice, that works - I've updated

PR #452, opened 2015-12-03
Setup Karma  for unit testing

### LINKED ISSUE #448: Add Karma

jleft:
Resolves #448

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -58,11 +58,17 @@ module.exports = function(grunt) {
         },
 
         watch: {
-            files: ['<%= meta.srcFiles %>'],
-            tasks: ['build:development'],
             options: {
                 atBegin: true,
                 livereload: true
+            },
+            dev: {
+                files: ['<%= meta.srcFiles %>'],
+                tasks: ['build:development']
+            },
+            devTest: {
+                files: ['<%= meta.ourJsFiles %>'],
+                tasks: ['build:development', 'karma:chromeBackground:run']
             }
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -454,7 +442,7 @@ module.exports = function(grunt) {
     grunt.registerTask('deploy', ['buildAndTest', 'gh-pages:origin']);
     grunt.registerTask('deploy:upstream', ['buildAndTest', 'gh-pages:upstream']);
 
-    grunt.registerTask('buildAndTest', ['build', 'test']);
+    grunt.registerTask('buildAndTest', ['build', 'test:phantom']);
     grunt.registerTask('dev', ['connect:watch', 'watch']);
```
tlsim: Might be nice to have `grunt dev` start a Karma server (Chrome only?) with `grunt watch` triggering running tests. https://github.com/karma-runner/grunt-karma#karma-server-with-grunt-watch

jleft: I'm not sure about this, at least with the current set of tests. The tests only cover a few utilities, so unless we're changing these utilities, then it's unlikely that the tests need to be run so frequently when developing. The `grunt test:chrome` task will watch for any changes and re-run the tests. Happy to combine the two into another task, if that'd be useful?

jleft: I've added the task `devTest` - let me know what you think :)

tlsim: :+1:

PR #453, opened 2015-12-03
 Move to ES6 modules and Rollup

### LINKED ISSUE #329: Move to ES6 modules and Rollup

tlsim:
Resolves #329 
Ready to try out at least!
Still to do: fix tests, linting - wait for #449 and #452 first.


Gruntfile.js
```diff
@@ -264,15 +247,38 @@ module.exports = function(grunt) {
...
                 }
             },
             production: {
+                files: {
+                    'dist/assets/js/app.js': ['src/assets/js/main.js']
+                },
+                options: {
+                    useStrict: false, // 'use-strict' is not supported by D3. d3#1755
+                    plugins: [
+                        require('rollup-plugin-npm')({
```
jleft: Just to check my understanding... is this used to bundle up our dependencies? Does it bundle the deps from their source?

I also noticed that Bootstrap is referenced in `vendorJsFiles` and as a dep in package.json - does this mean it will be included twice?

Are there any advantages to doing this over the previous approach (concat the already-minified vendor dist files to our bundled and minified code)? I expect using rollup and uglify with the vendor files will increase the build time (particularly with uglify, as this was particularly slow before #426)

Gruntfile.js
```diff
@@ -295,12 +301,22 @@ module.exports = function(grunt) {
...
+                    'test/**/*.js': ['browserify']
+                },
                 exclude: ['src/assets/js/main.js'],
                 files: [
                     '<%= meta.vendorJsFiles %>',
                     '<%= meta.srcJsFiles %>',
                     '<%= meta.testJsFiles %>'
-                ]
+                ],
+                browserify: {
```
jleft: Could be interesting to investigate the other options - Node/Jasmine/jsdom etc. but happy to roll with this for now :)

jleft:
Looks good to me - thanks @tlsim! A few questions above, but I'd be happy to see this merged :+1:

PR #454, opened 2015-12-03
Removes padYDomain (replaced by fc.util.extent()) and exposes extentAccessor to remove findTotalYExtent in primary

djmiley:


djmiley:
https://github.com/ScottLogic/d3fc/issues/806

src/assets/js/chart/primary.js
```diff
@@ -30,48 +30,14 @@ function produceAnnotatedTickValues(scale, annotation) {
...
-        }
-        if (movingAverageShown) {
-            var movingAverageExtent = fc.util.extent()
-              .fields('movingAverage')(visibleData);
-            extent[0] = d3.min([movingAverageExtent[0], extent[0]]);
-            extent[1] = d3.max([movingAverageExtent[1], extent[1]]);
-        }
-        if (!(movingAverageShown || bollingerBandsShown)) {
-            throw new Error('Unexpected indicator type');
+function findExtentAccessors(multiSeries) {
```
tlsim: Good stuff @djmiley, this is much neater.

src/assets/js/chart/primary.js
```diff
@@ -30,48 +30,14 @@ function produceAnnotatedTickValues(scale, annotation) {
...
+    var extentAccessors = [];
+    for (var i = 0; i < multiSeries.length; i++) {
```
tlsim: Could be even shorter with array `reduce`.

src/assets/js/initialiseModel.js
```diff
@@ -41,33 +41,45 @@ export default function() {
...
+          'candlestick',
+          candlestick,
+          'sc-icon-candlestick-series');
         candlestickOption.isSelected = true;
+        candlestickOption.option.config.extentAccessor = ['high', 'low'];
 
         var ohlc = fc.series.ohlc();
         ohlc.id = util.uid();
+        var ohlcOption = model.menu.option('OHLC', 'ohlc', ohlc, 'sc-icon-ohlc-series');
+        ohlcOption.option.config.extentAccessor = ['high', 'low'];
```
tlsim: I don't see any issue with storing `extentAccessor` on the `option` itself in a similar way to `id` above. Can then go without `config`.

djmiley: This is related to my attempts to modify indicators, where I stored their options on an `option.config`. I'd be very happy to do this if this is better. :)

tlsim: Ah I hadn't thought about that. Would we need the extent accessors to be user configurable? If not, `config` might not be the best place (if we go down this route for #73). But at the moment, the only advantage is that `option.extentAccessor` is a bit shorter.
Related to this, we use the `option` model for both series and indicators, but presumably we only need indicators to be configurable. We could subclass `option`? Some way to avoid `option.option` would be nice too - the `option` property could be renamed 'component' or something, so that `indicator.component` would be the underlying d3fc component.

djmiley: I like the sound of that! @jleft How do you feel about renaming the `option` parameter in `model.menu.option` to `component`.

djmiley: Yes, I agree, not putting it in the `config` object makes more sense for the `extentAccessor` currently, and presumably, at all future times.

jleft: > @jleft How do you feel about renaming the `option` parameter in `model.menu.option` to `component`.

:+1: I think that's a better name. I'd say it's low-priority for now, unless anyone disagrees? 

Although longer term, I'm not sure that we should be storing the component on the menu option object... that's a conversation for later though!

src/assets/js/initialiseModel.js
```diff
@@ -41,33 +41,45 @@ export default function() {
...
 
         var ohlc = fc.series.ohlc();
         ohlc.id = util.uid();
+        var ohlcOption = model.menu.option('OHLC', 'ohlc', ohlc, 'sc-icon-ohlc-series');
+        ohlcOption.option.extentAccessor = ['high', 'low'];
 
         var line = fc.series.line();
         line.id = util.uid();
+        var lineOption = model.menu.option('Line', 'line', line, 'sc-icon-line-series');
+        lineOption.option.extentAccessor = 'close';
```
jleft: Just wondering if something like this is a possibility? Where we can access access the properties directly from the component's methods:

```js
ohlcOption.option.extentAccessor = function() { return ohlc.yValue(); };
```

src/assets/js/chart/primary.js
jleft: :+1: Agreed, this is great, much simpler. Defeating the cyclomatic complexity! :)

Possibly rename to `getExtentAccessors`?

src/assets/js/chart/primary.js
```diff
@@ -30,48 +30,14 @@ function produceAnnotatedTickValues(scale, annotation) {
...
-        if (movingAverageShown) {
-            var movingAverageExtent = fc.util.extent()
-              .fields('movingAverage')(visibleData);
-            extent[0] = d3.min([movingAverageExtent[0], extent[0]]);
-            extent[1] = d3.max([movingAverageExtent[1], extent[1]]);
-        }
-        if (!(movingAverageShown || bollingerBandsShown)) {
-            throw new Error('Unexpected indicator type');
+function findExtentAccessors(multiSeries) {
+    return multiSeries.reduce(function(mem, series) {
```
jleft: Why name the argument `mem`?

src/assets/js/chart/primary.js
```diff
@@ -213,9 +179,11 @@ export default function() {
 
         // Scale y axis
         var visibleData = util.domain.filterDataInDateRange(primaryChart.xDomain(), model.data);
-        var yExtent = findTotalYExtent(visibleData, currentSeries, currentIndicators);
-        // Add percentage padding either side of extreme high/lows
-        var paddedYExtent = util.domain.padYDomain(yExtent, 0.04);
+        // // Add percentage padding either side of extreme high/lows
```
jleft: `// //` -> `//`

src/assets/js/initialiseModel.js
```diff
@@ -77,21 +89,28 @@ export default function() {
         var secondary = chart.secondary;
 
         var movingAverage = fc.series.line()
-            .decorate(function(select) {
-                select.enter()
-                    .classed('movingAverage', true);
-            })
-            .yValue(function(d) { return d.movingAverage; });
+          .decorate(function(select) {
```
jleft: Indentation doesn't look right here: 2 -> 4 spaces

src/assets/js/model/menu/option.js
```diff
@@ -1,4 +1,7 @@
 export default function(displayString, valueString, option, icon, isPrimary) {
+
+    option.config = {};
```
jleft: Is this irrelevant now?

jleft:
:+1: LGTM, just a few minor comments

jleft:
:+1: thanks @djmiley

@tlsim - if you don't have any further comments, then feel free to merge :)

src/assets/js/chart/primary.js
```diff
@@ -30,48 +30,14 @@ function produceAnnotatedTickValues(scale, annotation) {
...
-            var movingAverageExtent = fc.util.extent()
-              .fields('movingAverage')(visibleData);
-            extent[0] = d3.min([movingAverageExtent[0], extent[0]]);
-            extent[1] = d3.max([movingAverageExtent[1], extent[1]]);
-        }
-        if (!(movingAverageShown || bollingerBandsShown)) {
-            throw new Error('Unexpected indicator type');
+function getExtentAccessors(multiSeries) {
+    return multiSeries.reduce(function(extentAccessors, series) {
+        if (series.extentAccessor) {
```
jdhodges431: How about:

`return extentAccessors.concat(series.extentAccessor || []);`

PR #462, opened 2015-12-07
Navbar hidden for small and x-small devices

### LINKED ISSUE #400: Hide navigator

rjmcneill:
Closes #400


jleft:
I don't think it was clear in the designs if the navbar should also be hidden on landscape tablet (based on v2, I think it should). We can seek clarification and raise a bug if this needs to changed.

:+1:

rjmcneill:
@jleft yea, I assumed it was supposed to be since there will be even less vertical space.
But like you say, can investigate it further after UX feedback

PR #467, opened 2015-12-10
Add data adaptors and integrate Quandl

### LINKED ISSUE #409: Investigate using Quandl as a data source
If we encounter problems with a Bitcoin API(s), could Quandl be a suitable alternative?

http://d3fc.io/components/data/quandl.html

jleft:
~~Thought this was worth putting up as a WIP, I'll integrate with #437 once it's merged.~~ Done.

Any thoughts or comments welcome!

Resolves #409


src/assets/js/app.js
```diff
@@ -56,11 +66,24 @@ export default function() {
         seconds: 60,
         d3TimeInterval: {unit: d3.time.minute, value: 1},
         timeFormat: '%H:%M'});
+
+    var generatedSource = model.data.source(dataGeneratorAdaptor());
+    var bitcoinSource = model.data.source(coinbaseAdaptor(), webSocket());
+    var quandlSource = model.data.source(quandlAdaptor());
+
     var generated = model.data.product({
-        family: 'generated',
```
jleft: `source` replaces `family`

src/assets/js/data/coinbase/formatProducts.js
```diff
@@ -1,24 +1,23 @@
 import model from '../../model/model';
 
-export default function(minute1, minute5, hour1, day1, response) {
-    var products = response.map(function(product) {
-        if (product.id !== 'BTC-USD') {
+export default function(products, source, defaults, exceptions) {
```
jleft: Thought this change would make it easier to maintain if we want to change the defaults or add another exception - reduces the potential for lots of branching

src/assets/js/data/coinbase/getProducts.js
```diff
@@ -1,13 +1,12 @@
 import d3 from 'd3';
 import model from '../../model/model';
-import formatProducts from './formatProducts';
 
-export default function(minute1, minute5, hour1, day1, callback) {
+export default function(callback) {
```
jleft: Now has a single responsibility for getting the products; formatting is done separately.

Could be worth componentising, but this is so small that I didn't think it was worth it.

test/data/coinbase/formatProductsSpec.js
```diff
@@ -1,14 +1,18 @@
...
 
 var min1 = 60;
 var min5 = 300;
 var hr1 = 3600;
 var day1 = 86400;
+var defaults = [hr1, day1];
+var exceptions = d3.map();
+exceptions.set('BTC-USD', [min1, min5, hr1, day1]);
 
 describe('data/coinbase/formatProducts', function() {
```
jleft: I'm not convinced that we should be testing what, I think, is essentially some configuration settings for the application (the products which we have)?

Rather than checking that our configuration is correct, I think we should be testing the units of code and their behaviour - otherwise we're going to have to duplicate the configuration (to some extent) in both the application source and our tests. E.g. the final three tests are testing exactly the same behaviour just with different configuration values for the product.

I'd suggest that we should be testing:
+ that default values can assigned when a product is not listed as an exception 
+ the exceptions are applied if a product is listed as an exception.

I've kept this as before for now (apart for updating the params) but happy to update.

src/assets/js/app.js
```diff
@@ -41,6 +46,11 @@ export default function() {
         seconds: 86400,
         d3TimeInterval: {unit: d3.time.day, value: 1},
         timeFormat: '%b %d'});
+    var week1 = model.data.period({
+        display: 'Weekly',
+        seconds: 60 * 60 * 24 * 7,
```
rjmcneill: Should we have all our seconds worked out like this?
Could be clearer that way.

src/assets/js/app.js
```diff
@@ -244,17 +267,17 @@ export default function() {
...
                 resetToLatest();
                 updateLayout();
             })
-            .on(event.dataLoadError, function(err) {
+            .on(event.historicFeedError, function(err) {
                 console.log('Error getting historic data: ' + err); // TODO: something more useful for the user!
             })
-            .on(event.webSocketError, function(err) {
+            .on(event.streamingFeedError, function(err) {
                 console.log('Error loading data from websocket: ' + err);
```
rjmcneill: I know you didn't change this, but is it worth adding this beside it?

```
 // TODO: something more useful for the user!
```

src/assets/js/app.js
jleft: Nice spot :) updated

src/assets/js/app.js
jleft: Sounds good - done.

src/assets/js/data/adaptor/dataGenerator.js
```diff
@@ -0,0 +1,62 @@
...
+import fc from 'd3fc';
+
+export default function() {
+
+    var dataGenerator = fc.data.random.financial(),
+        allowedPeriods = [60 * 60 * 24],
+        candles,
+        end,
+        granularity,
+        product = null;
```
jdhodges431: For the sake of consistency, would it be better to use individual var statements rather than combined?

rjmcneill:
LGTM @jleft :+1:

jdhodges431:
Looks good :+1:

test/data/coinbase/formatProductsSpec.js
tlsim: I suppose what separates the current situation from pure configuration is that every product returned from the coinbase API will be added to the dropdown, and formatProducts contains logic to override the default periods for specified product ids. We could instead have a single productId/periods map for all products. There would be no override logic to test.

src/assets/js/data/adaptor/dataGenerator.js
jleft: Good question, it seems we vary our approach throughout the application. Opened an issue to cover this: #479

test/data/coinbase/formatProductsSpec.js
jleft: I think having the override logic is fine (unless this change has made formatProducts too generic?).

My main concern with the tests is that the last three are pretty much identical - unless we did specify the product(s) which we want to override in formatProducts (which is no longer the case), I can't think of a scenario where they would not all pass or all fail, as they're just testing that the defaults are set.

tlsim: Ok, we'll stay with that. You're welcome to improve the tests :smile:

src/assets/js/app.js
```diff
@@ -298,16 +316,19 @@ export default function() {
...
 
     function insertProductsIntoHeadMenuModel(error, bitcoinProducts) {
         if (error) {
-            console.log('Error getting coinbase products: ' + error); // TODO: something more useful for the user!
+            console.log('Error getting Coinbase products: ' + error); // TODO: something more useful for the user!
         } else {
-            // Add the newly received products to the product list
-            headMenuModel.products = headMenuModel.products.concat(bitcoinProducts);
+            var defaultPeriods = [hour1, day1];
+            var exceptionPeriods = d3.map();
```
tlsim: Maybe 'override' / 'overridden' instead of 'exception'?

src/assets/js/data/adaptor/dataGenerator.js
```diff
@@ -0,0 +1,62 @@
...
+    var dataGeneratorAdaptor = function(cb) {
+        end.setHours(0, 0, 0, 0);
+        var millisecondsPerDay = 24 * 60 * 60 * 1000;
+        dataGenerator.startDate(new Date(end - (candles - 1) * millisecondsPerDay));
+
+        var data = dataGenerator(candles);
+        cb(null, data);
+    };
+
+    dataGeneratorAdaptor.candles = function(x) {
```
tlsim: What's your thinking on the future interface for the adaptors? We could move to `start`, `end` and `granularity` instead of some constant number of candles from the present time.

test/data/coinbase/formatProductsSpec.js
jleft: Sounds good, thanks :smiley:

src/assets/js/app.js
jleft: Updated to `productPeriodOverrides`

src/assets/js/data/adaptor/dataGenerator.js
jleft: > What's your thinking on the future interface for the adaptors? We could move to start, end and granularity instead of some constant number of candles from the present time.

I'm not sure, but I've been thinking the same as your suggestion - seems like a much nicer interface to me :smiley:

tlsim: Scottlogic/d3fc#763 should help us here.

tlsim:
:+1:

PR #473, opened 2015-12-14
Overlay model updates when clear all button is clicked.

### LINKED ISSUE #472: Clear all button doesn't update overlay model
Indicators removed from chart when clear all button is clicked, but indicator labels are not.

jdhodges431:
Resolves #472.


jleft:
Nice work :+1:

PR #474, opened 2015-12-14
Selectors repositioned to head on mobile.

### LINKED ISSUE #403: Re-position menus

jdhodges:
Closes #403.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -22,8 +22,13 @@ export default function() {
         <div class="row head-menu head-row"> \
             <div class="col-md-12 head-sub-row"> \
                 <div id="product-dropdown" class="dropdown product-dropdown"></div> \
+                <div class="selectors"> \
+                    <div class="series-dropdown dropdown selector-dropdown"></div> \
+                    <div id="mobile-period-selector" class="dropdown"></div> \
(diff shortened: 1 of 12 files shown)


rjmcneill:
LGTM @jdhodges431 

One issue with it though is how squashed everything looks on mobile.
I think it's probably because of our addition of the product dropdown in the mobile designs that UX didn't have.  So I guess that's sort of blocked by them until we know what they want.
Screenshot shows this.
Also, the clear all button is pushed out into the margin now I think.

mobileview

jdhodges431:
Yeah may be able to make the buttons smaller, the spec doesn't give a size.

jleft:
I agree with @rjmcneill - it does look fairly squashed and losing the clear all button isn't ideal.

I'll ask for confirmation on how a product should be selected on mobile. For now, I'd suggest, that we replace the product selectors dropdown text with an icon on mobile, if that seems reasonable?

product-search

There's also quite a lot of padding in the Add Indicator button either side of the text - is this something which can be reduced? (At least on mobile.)

jdhodges431:
Sorry for wasting everyone's precious time with that first attempt. Hopefully this should be an improvement :pray:

PR #476, opened 2015-12-15
Debounce Coinbase historic API requests to avoid reaching rate limit

### LINKED ISSUE #468: Work around Coinbase rate limit
#464

rjmcneill:
Closes #468

### DIFF
--- a/src/assets/js/data/coinbase/historic/feedAdaptor.js
+++ b/src/assets/js/data/coinbase/historic/feedAdaptor.js
@@ -1,16 +1,18 @@
 import d3 from 'd3';
 import fc from 'd3fc';
+import debounce from '../../../util/debounce';
 
 export default function() {
+    var rateLimit = 1000;       // The coinbase API has a limit of 1 request per second
 
     var historicFeed = fc.data.feed.coinbase(),
         candles;
 
-    function coinbaseAdaptor(cb) {
+    var coinbaseAdaptor = debounce(function coinbaseAdaptor(cb) {
         var startDate = d3.time.second.offset(historicFeed.end(), -candles * historicFeed.granularity());
         historicFeed.start(startDate);
         historicFeed(cb);
-    }
+    }, rateLimit);
 
     coinbaseAdaptor.candles = function(x) {
         if (!arguments.length) {
--- a/src/assets/js/util/debounce.js
+++ b/src/assets/js/util/debounce.js
@@ -1,0 +1,38 @@
+// Inspired by underscore library implementation of debounce
+
+export default function(func, wait, immediate) {
+    var timeout;
+    var args;
+    var timestamp;
+    var result;
+
+    var later = function() {
+        var last = new Date().getTime() - timestamp;
+
+        if (last < wait && last >= 0) {
+            timeout = setTimeout(later.bind(this), wait - last);
+        } else {
+            timeout = null;
+            if (!immediate) {
+                result = func.apply(this, args);
+                args = null;
+            }
+        }
+    };
+
+    return function() {
+        args = arguments;
+        timestamp = new Date().getTime();
+        var callNow = immediate && !timeout;
+
+        if (!timeout) {
+            timeout = setTimeout(later.bind(this), wait);
+        }
+        if (callNow) {
+            result = func.apply(this, args);
+            args = null;
+        }
+
+        return result;
+    };
+}
--- a/src/assets/js/util/throttle.js
+++ b/src/assets/js/util/throttle.js
@@ -1,0 +1,55 @@
+// Inspired by underscore library implementation of throttle
+
+export default function(func, wait, options) {
+    var args;
+    var result;
+    var timeout = null;
+    var previous = 0;
+
+    if (!options) {
+        options = {};
+    }
+
+    var later = function() {
+        if (options.leading === false) {
+            previous = 0;
+        } else {
+            previous = new Date().getTime();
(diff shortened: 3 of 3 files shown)


rjmcneill:
Todo:

- Implement a throttle (removing underscore)
- Try throttling API call itself

rjmcneill:
Still WIP.
Will be good to go tomorrow :)

src/assets/js/data/coinbase/historic/feedAdaptor.js
```diff
@@ -1,26 +1,27 @@
...
 import fc from 'd3fc';
+import debounce from '../../../util/debounce';
 
 export default function() {
 
     var historicFeed = fc.data.feed.coinbase(),
         candles;
 
-    function coinbaseAdaptor(cb) {
+    var debouncedCoinbaseAdaptor = debounce(function coinbaseAdaptor(cb) {
```
jleft: I'd keep the name as `coinbaseAdaptor`

src/assets/js/data/coinbase/historic/feedAdaptor.js
```diff
@@ -1,26 +1,27 @@
...
         var startDate = d3.time.second.offset(historicFeed.end(), -candles * historicFeed.granularity());
         historicFeed.start(startDate);
         historicFeed(cb);
-    }
+    }, 1000);
```
jleft: `1000` looks like it could be an arbitrary/magic number - I'd pull it out as a variable and add a comment explaining this is to prevent us hitting the Coinbase rate limit of 1 request / second

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
+/* Inspired by underscore library implementation of debounce */
```
jleft: We tend to use `//` for comments

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default function(func, wait, immediate) {
+    var timeout, args, context, timestamp, result;
```
jleft: Looks like `context` is unused?

I'd separate each variable on to a line of their own.

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
...
+    var later = function() {
```
jleft: New line before this line to separate the variable and function declarations

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
...
+        var last = new Date().getTime() - timestamp;
+
+        if (last < wait && last >= 0) {
+            timeout = setTimeout(later, wait - last);
+        } else {
+            timeout = null;
+            if (!immediate) {
+                result = func.apply(this, args);
+                if (!timeout) {
```
jleft: I feel like I'm missing something obvious... but as `timeout = null` above, will this always evaluate to `true`?

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
...
+                    args = null;
+                }
+            }
+        }
+    }.bind(this);
```
jleft: See comment on `bind` below

src/assets/js/util/debounce.js
```diff
@@ -0,0 +1,36 @@
...
+        if (!timeout) {
+            timeout = setTimeout(later, wait);
+        }
+        if (callNow) {
+            result = func.apply(this, args);
+            args = null;
+        }
+
+        return result;
+    }.bind(this);
```
jleft: This doesn't affect its current usage (as it's only used for the Coinbase adaptor, which doesn't use `this`), but check the use of `bind(this)`, both in debounce and throttle.

I _think_ `this` will be `undefined` here, we could lead to some issues. For example, if we were to run the following code, a type error would be thrown, rather than `3` being logged:

```js
var obj = {
    x: 1,
    y: 2,
    sum: debounce(function() { 
        console.log(this.x + this.y);
    }, 100)
};

obj.sum(); // type error, when we'd expect 3 to be logged
```

Generally, the value of `this` is determined by how the function is _invoked_, not how it is _defined_. In this case `bind` will return a function which will always be invoked with the same value of `this`, which is `undefined`, leading to the type error (can't access properties of undefined).

src/assets/js/util/throttle.js
```diff
@@ -0,0 +1,55 @@
+/* Inspired by underscore library implementation of throttle */
```
jleft: `//`

src/assets/js/util/throttle.js
```diff
@@ -0,0 +1,55 @@
...
+            previous = new Date().getTime();
+        }
+
+        timeout = null;
+        result = func.apply(this, args);
+
+        if (!timeout) {
+            args = null;
+        }
+    }.bind(this);
```
jleft: See comment on `bind` above

src/assets/js/util/throttle.js
```diff
@@ -0,0 +1,55 @@
...
+
+            if (!timeout) {
+                args = null;
+            }
+        } else if (!timeout && options.trailing !== false) {
+            timeout = setTimeout(later, remaining);
+        }
+
+        return result;
+    }.bind(this);
```
jleft: See comment on `bind` above

jleft:
Nice work :+1:

PR #478, opened 2015-12-16
Add error message notifications for better error handling

### LINKED ISSUE #445: Error handling - visuals

### LINKED ISSUE #446: Error handling - data errors
#405

jleft:
Resolves #445 
Resolves #446


src/assets/js/notification/toast.js
```diff
@@ -0,0 +1,49 @@
...
+        selection.each(function(model) {
+            var container = d3.select(this);
+
+            var panel = panelDataJoin(container, [model]);
+            panel.enter().html('<div class="messages"></div>');
+
+            var toasts = toastDataJoin(panel.select('.messages'), model.messages);
+
+            var toastsEnter = toasts.enter();
+            toastsEnter.html(
```
jdhodges431: Should we be toastsEnter.append here rather than a chunk of hard coded html?

jleft: I tried that at first, but there was a _lot_ of code to create just static elements. It's also only on `enter`, so I don't think we'll have much of an issue performance-wise (at least compared to doing this on update).

d3fc's cartesian chart does something similar:
https://github.com/ScottLogic/d3fc/blob/master/src/chart/cartesian.js#L67-L94

jleft:
Updated

src/assets/js/app.js
```diff
@@ -267,18 +278,40 @@ export default function() {
...
+                }
+                var statusText = err.statusText || 'Unkown reason.';
+                var message = 'Error getting historic data: ' + statusText + responseText;
+
+                notificationMessagesModel.messages.unshift(model.notification.message(message));
+                render();
             })
-            .on(event.streamingFeedError, function(err) {
-                console.log('Error loading data from websocket: ' + err); // TODO: something more useful for the user!
+            .on(event.streamingFeedError, function(error, source) {
```
rjmcneill: Do we need to have this error event if we're not using it?
Maybe we could just remove it entirely if we're not worried about it?

rjmcneill:
LGTM @jleft :+1:

src/assets/js/app.js
```diff
@@ -267,18 +278,40 @@ export default function() {
...
+                // The WebSocket's error event doesn't contain much useful information,
+                // so the close event is used to report errors instead
+            })
+            .on(event.streamingFeedClose, function(closeEvent, source) {
+                // 1000 is normal closure
```
tlsim: I know we don't have any other streaming source yet, but we're handling websocket specific errors here. Will we need something similar to the notification formatters for streaming feeds?

src/assets/js/data/dataInterface.js
```diff
@@ -2,23 +2,21 @@ import d3 from 'd3';
...
+        event.newTrade,
         event.historicDataLoaded,
         event.historicFeedError,
-        event.newTrade,
-        event.streamingFeedError);
+        event.streamingFeedError,
+        event.streamingFeedClose);
 
     var _collectOhlc = collectOhlc()
         .date(function(d) {return new Date(d.time); })
```
tlsim: Configured like this, it's coinbase specific. Not sure how best to make generic - could the adaptors modify the trades received?

src/assets/js/notification/toast.js
tlsim: Looks ok to me - might this also be a nice way to move parts of index.html moved into the corresponding component?

src/assets/js/app.js
jleft: Good point, I was trying to make it a bit more generic, with the potential for data generator streaming (maybe I should have followed YAGNI instead).

> Will we need something similar to the notification formatters for streaming feeds?

Yes, that makes sense to me - I'll update to do this now as it should be a very small change.

jleft:
TODO:

- [ ] Add notification formatters for streaming feeds

src/assets/js/data/dataInterface.js
jleft: > Not sure how best to make generic - could the adaptors modify the trades received?

Yeah, that would work; a couple of approaches could be to:
- map the data from all sources into the same format (this would be my preference)
- expose the accessors for the data's properties on the data feeds and pass these for collectOhlc, for example:

```js
collectOhlc().date(streamingFeed.date)
```

tlsim: :+1: for option 1.

tlsim:
LGTM :+1:

jleft:
Thanks for the reviews :smiley:

src/assets/js/app.js
```diff
@@ -214,11 +236,13 @@ export default function() {
...
     }
 
-    function loading(isLoading) {
-        appContainer.select('#loading-message')
-            .classed('hidden', !isLoading);
+    function loading(isLoading, error) {
+        appContainer.select('#loading-status-message')
+            .classed('hidden', !(isLoading || error))
+            .select('.content')
+            .text(error ? error : 'Loading...');
```
jdhodges431: `.text(error || 'Loading...');` instead?

jleft: Good spot, stylistically, I think that would be more consistent.

jleft: https://github.com/ScottLogic/d3fc-showcase/pull/491

src/assets/js/app.js
```diff
@@ -267,19 +291,33 @@ export default function() {
...
+                try {
+                    var responseObject = JSON.parse(err.responseText);
+                    var formattedMessage = source.historicNotificationFormatter(responseObject);
+                    if (formattedMessage) {
+                        responseText = '. ' + formattedMessage;
+                    }
+                } catch (e) {
+                    responseText = '';
+                }
+                var statusText = err.statusText || 'Unkown reason.';
```
jdhodges431: Typo: Unkown => Unknown

jleft: #492

src/assets/js/app.js
```diff
@@ -322,15 +360,17 @@ export default function() {
 
     function insertProductsIntoHeadMenuModel(error, bitcoinProducts) {
         if (error) {
-            console.log('Error getting Coinbase products: ' + error); // TODO: something more useful for the user!
+            var statusText = error.statusText || 'Unkown reason.';
```
rjmcneill: Same typo here :)

Issue #456, opened 2015-12-04
Refactor models


jleft:
Continuing from #441:

- Should we be consistent with how we pass configuration values to the models? If so, which approach should we use? Currently, it's either several params (e.g. `sc.model.menu.option`) or a single config object (e.g. `sc.model.data.product`).
- I think adding an `id` property to the renders, so we can key on the `id` is good (or at least better than `isLine`) and should make future work (e.g. additive indicators) easier. Not so keen on the implementation - there is repetition. Perhaps we should make own series/indicator components which wrap the d3fc components and add the `id` to them?
- Should models have default values?
- Should we instantiate series/indicator components outside of `sc.model.whatever`? We could do this in a similar way to period and products, which are instantiated in app.js.

PR #483, opened 2015-12-17
Refactored the model. Objects initialised outside of model in initial…

jdhodges431:
…iseModel. Also removed config objects.

Relates to #456.


src/assets/js/initialiseModels.js
```diff
@@ -0,0 +1,121 @@
...
+        secondaryChart: model.chart.secondary(dataGenerator),
+        selectors: initSelectors(),
+        xAxis: model.chart.xAxis(periods.day1),
+        nav: model.chart.nav(),
+        navReset: model.chart.navigationReset(),
+        headMenu: headMenu,
+        legend: model.chart.legend(dataGenerator, periods.day1),
+        overlay: model.menu.overlay()
+    };
+}
```
djmiley: Would these not be better in separate files? Not sure what everyone else thinks. I like the use of `initialiseModel` though, but you could import into this file from other files.

src/assets/js/app.js
```diff
@@ -2,17 +2,13 @@
...
 import util from './util/util';
 import event from './event';
 import dataInterface from './data/dataInterface';
 import coinbaseProducts from './data/coinbase/getProducts';
-import coinbaseAdaptor from './data/adaptor/coinbase';
-import dataGeneratorAdaptor from './data/adaptor/dataGenerator';
-import quandlAdaptor from './data/adaptor/quandl';
-import webSocket from './data/coinbase/webSocket';
 import formatProducts from './data/coinbase/formatProducts';
+import initialiseModels from './initialiseModels';
```
tlsim: Super pedantic, but could change to `initialiseModel` just to emphasise that it returns single top level model object.

tlsim:
Looks good :+1: Just needs a rebase.

src/assets/js/initialiseModels.js
```diff
@@ -0,0 +1,143 @@
...
+            var line = fc.series.line();
+            line.id = util.uid();
+
+            var point = fc.series.point();
+            point.id = util.uid();
+
+            var area = fc.series.area();
+            area.id = util.uid();
+
+            var config = model.menu.dropdownConfig(null, false, true, true);
```
rjmcneill: It's not really clear what these arguments are actually for when just reading through the code.
Is it worth adding a comment or changing the approach?

rjmcneill:
LGTM @jdhodges431

src/assets/js/initialiseModels.js
jdhodges431: This was mentioned in #456 and #441. We went for passing several params rather than a config object when creating objects - this is a good example of one of the issues with passing params. However the alternative would involve passing a config object and returning exactly the same object back, which I think we agreed was fairly pointless.

src/assets/js/app.js
jdhodges431: Yeah, definitely a good idea.

src/assets/js/initialiseModel.js
```diff
@@ -0,0 +1,150 @@
...
+        primaryChart: model.chart.primary(products.generated),
+        secondaryChart: model.chart.secondary(products.generated),
+        selectors: initSelectors(),
+        xAxis: model.chart.xAxis(periods.day1),
+        nav: model.chart.nav(),
+        navReset: model.chart.navigationReset(),
+        headMenu: model.menu.head([products.generated, products.quandl], products.generated, periods.day1),
+        legend: model.chart.legend(products.generated, periods.day1),
+        overlay: model.menu.overlay(),
+        notificationMessage: model.notification.message,
```
jleft: I don't think this belongs here - it isn't an initialised model.

src/assets/js/initialiseModel.js
```diff
@@ -0,0 +1,150 @@
...
+                model.menu.option('Line', 'line', line, 'sc-icon-line-series'),
+                model.menu.option('Point', 'point', point, 'sc-icon-point-series'),
+                model.menu.option('Area', 'area', area, 'sc-icon-area-series')
+            ];
+
+            return model.menu.selector(config, options);
+        }
+
+        function initIndicatorSelector() {
+            function initIndicatorOptions() {
```
jleft: I think we tend not to nest functions (please correct me if I'm wrong!), excluding when declaring a module, so I don't think this is particularly consistent with our style.

As the nesting isn't that deep and it makes sense, I'm okay with it. Would be interested to see if anyone else has thoughts about this though.

src/assets/js/initialiseModel.js
```diff
@@ -0,0 +1,150 @@
...
+
+                var primaryIndicators = [
+                    model.menu.option('Moving Average', 'movingAverage',
+                        movingAverage, 'sc-icon-moving-average-indicator'),
+                    model.menu.option('Bollinger Bands', 'bollinger',
+                        bollingerBands, 'sc-icon-bollinger-bands-indicator')
+                ];
+
+                primaryIndicators.forEach(function(_option) {
+                    _option.isPrimary = true;
```
jleft: Worth adding the `isPrimary` and `isSecondary` properties to the option model?

jleft:
Looks really good, thanks @jdhodges431 - just a few comments above :)

Once these and adressed and this is rebased, then I'm happy to see this merged :+1:

src/assets/js/initialiseModel.js
tlsim: I agree. The non-nested version would be more readable.

jleft:
Cheers @jdhodges431

PR #484, opened 2015-12-18
Implements padding of domain

djmiley:


djmiley:
A little bug has been introduced, just a rebase here.

djmiley:
Bug fixed.

djmiley:
Couple more things to do to tidy up. Seperated the commits for ease of reverting. Would be happy to squash to one commit for merging.

src/assets/js/app.js
```diff
@@ -289,12 +300,14 @@ export default function(initialModel) {
                 loading(true);
                 updateModelSelectedProduct(product.option);
                 updateModelSelectedPeriod(product.option.periods[0]);
+                updateModelExtentPadding(1000 * product.option.periods[0].seconds / 2);
```
rjmcneill: Maybe change this to:
``` JavaScript
500 * product.option.periods[0].second
```
Unless you have another reason for doing it this way?

src/assets/js/app.js
```diff
@@ -289,12 +300,14 @@ export default function(initialModel) {
...
                 _dataInterface(product.option.periods[0].seconds, product.option);
                 render();
             })
             .on(event.dataPeriodChange, function(period) {
                 loading(true);
                 updateModelSelectedPeriod(period.option);
+                updateModelExtentPadding(1000 * period.option.seconds / 2);
```
rjmcneill: Maybe change this to:
``` JavaScript
updateModelExtentPadding(500 * period.option.seconds);
```
Unless you have another reason for doing it this way?

src/assets/js/behavior/zoom.js
```diff
@@ -41,13 +42,19 @@ export default function(width) {
...
 
-        var xExtent = fc.util.extent()
-          .fields('date')(selection.datum().data);
 
         zoomBehavior.x(scale)
           .on('zoom', function() {
-              var min = scale(xExtent[0]);
-              var max = scale(xExtent[1]);
+              var xExtent = fc.util.extent()
+                .fields('date')(selection.datum().data);
```
rjmcneill: 4 space indent?

src/assets/js/model/chart/nav.js
```diff
@@ -1,7 +1,8 @@
-export default function() {
+export default function(initialPeriod) {
     return {
         data: [],
         viewDomain: [],
+        padding: 1000 * initialPeriod.seconds / 2,
```
rjmcneill: padding: 500 * initialPeriod.seconds,

src/assets/js/model/chart/primary.js
```diff
@@ -2,11 +2,12 @@ import candlestickSeries from '../../series/candlestick';
 import option from '../menu/option';
 import util from '../../util/util';
 
-export default function(initialProduct) {
+export default function(initialProduct, initialPeriod) {
     var model = {
         data: [],
         trackingLatest: true,
         viewDomain: [],
+        padding: 1000 * initialPeriod.seconds / 2,
```
rjmcneill:
``` Javascript
padding: 500 * initialPeriod.seconds,
```

src/assets/js/model/chart/secondary.js
```diff
@@ -1,8 +1,9 @@
-export default function(initialProduct) {
+export default function(initialProduct, initialPeriod) {
     return {
         data: [],
         viewDomain: [],
         trackingLatest: true,
+        padding: 1000 * initialPeriod.seconds / 2,
```
rjmcneill:
``` Javascript
padding: 500 * initialPeriod.seconds,
```

src/assets/js/model/chart/xAxis.js
```diff
@@ -1,6 +1,7 @@
 export default function(initialPeriod) {
     return {
         viewDomain: [],
+        padding: 1000 * initialPeriod.seconds / 2,
```
rjmcneill:
``` Javascript
padding: 500 * initialPeriod.seconds,
```

rjmcneill:
@djmiley feel free to ignore these comments if you have a reason for doing it the way you have.

tlsim:
@djmiley - I like your approach over say, adding extra null data (which would likely complicate the y axis range calculation, latest close line, data-domain utils and more). 

It seems that since lots of our application logic assumes an unpadded domain, by maintaining the padded domain in app.js, we have to 'unpad' the domain more than we'd really like to at lower levels (zoom behavior, navigator brush), and generally worry more about whether a domain is padded or not.

I'm wondering if it might be possible to always work with an unpadded domain for app logic, and then only at the level of rendering the components actually pad the domain.
I got quite a promising result adding these lines to the primary chart render, for example:
```javascript
// Calculate padded domain.
var paddedXExtent = fc.util.extent()
    .fields(fc.util.fn.identity)
    .pad(0.1)(model.viewDomain); // Arbitrary pad, would get from model in reality.

var zoomScale = xScale.copy(); // Pass this to zoomBehavior.
zoomScale.domain(model.viewDomain); // zoomBehavior logic requires unpadded domain.
primaryChart.xDomain(paddedXExtent); // pad the xDomain for primary chart render only.
```
I'm happy to hear if this wouldn't work. What do you think?

src/assets/js/model/chart/nav.js
tlsim: Not sure the nav needs padding - the domain doesn't have to line up with anything else, unlike the secondary charts.

src/assets/js/app.js
djmiley: 1000 for milliseconds, / 2 for half at either end. 500 looks more of a magic number? I should probably add a comment?

src/assets/js/app.js
djmiley: 1000 for milliseconds, / 2 for half at either end. 500 looks more of a magic number? I should probably add a comment?

src/assets/js/model/chart/nav.js
djmiley: It needs padding for the zoom to work correctly.

djmiley: At least in the current implementation, I mean :)

src/assets/js/initialiseModel.js
```diff
@@ -142,11 +142,12 @@ export default function() {
     return {
         periods: periods,
         sources: sources,
-        primaryChart: model.chart.primary(products.generated),
-        secondaryChart: model.chart.secondary(products.generated),
-        selectors: initSelectors(),
+        products: products,
```
djmiley: Remove products?

src/assets/js/chart/nav.js
```diff
@@ -139,7 +139,13 @@ export default function() {
 
         var brushHide = false;
 
-        navChart.xDomain(fc.util.extent().fields('date')(model.data))
+        // var dataExtent = fc.util.extent().fields('date')(model.data);
```
djmiley: Remove commented text

djmiley:
@tlsim I'm confident that this would work, and would make it neater somewhat. Thanks for the suggestion!

PR #496, opened 2016-01-04
Initialise websocket after historic data received

### LINKED ISSUE #470: Live data is collected when historic data fetch fails.
Causes errors first seen in #464 
Behaviour changed in #455. 
Revert to only initialising websocket on historic data success?

rjmcneill:
Closes #470

### DIFF
--- a/src/assets/js/data/dataInterface.js
+++ b/src/assets/js/data/dataInterface.js
@@ -35,6 +35,22 @@ export default function() {
         });
     }
 
+    function handleStreamingFeedEvents() {
+        if (source.streamingFeed != null) {
+            source.streamingFeed.on('message', function(trade) {
+                _collectOhlc(data, trade);
+                dispatch[event.newTrade](data, source);
+            })
+            .on('error', function(streamingFeedError) {
+                dispatch[event.streamingFeedError](streamingFeedError, source);
+            })
+            .on('close', function(closeEvent) {
+                dispatch[event.streamingFeedClose](closeEvent, source);
+            });
+            source.streamingFeed();
+        }
+    }
+
     function dataInterface(granularity, product) {
         invalidate();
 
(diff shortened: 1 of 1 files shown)


src/assets/js/data/coinbase/streaming/webSocket.js
```diff
@@ -9,6 +9,7 @@ export default function() {
     var dispatch = d3.dispatch('open', 'close', 'error', 'message');
     var messageType = 'match';
     var socket;
+    var historicDataLoaded = false;
```
jleft: I don't think the WebSocket should be concerned about whether historic data has been loaded or not - this module would have higher cohesion and be more loosely coupled if wasn't concerned with this.

jleft:
Thanks @rjmcneill - I'd consider looking to make this change in `dataInterface`, it feels like its responsibility to me. Would something like this work?

```js
source.historicFeed(callbackGenerator(function(historicFeedError, newData) {
    if (!historicFeedError) {
        data = dateSortAscending(newData);
        dispatch[event.historicDataLoaded](data, source);

        // Could pull this out into a function...
        if (source.streamingFeed != null) {
            source.streamingFeed.on('message', function(trade) {
                _collectOhlc(data, trade);
                dispatch[event.newTrade](data, source);
            })
            .on('error', function(streamingFeedError) {
                dispatch[event.streamingFeedError](streamingFeedError, source);
            })
            .on('close', function(closeEvent) {
                dispatch[event.streamingFeedClose](closeEvent, source);
            });

            source.streamingFeed();
        }
    } else {
        dispatch[event.historicFeedError](historicFeedError, source);
    }
}));
```

rjmcneill:
@jleft yup, that looks much easier!
I agree with you that this is a better approach as well, keeping them loosely coupled

jleft:
:+1: LGTM

PR #509, opened 2016-01-04
Text no longer selectable

### LINKED ISSUE #504: Can we disable text selection?
See the following:

capture

rjmcneill:
Closes #504

### DIFF
--- a/src/assets/styles/charts.less
+++ b/src/assets/styles/charts.less
@@ -2,3 +2,7 @@
   stroke: #06c;
   fill: #9cf;
 }
+
+#charts {
+  user-select: none;
+}
--- a/src/assets/styles/head.less
+++ b/src/assets/styles/head.less
@@ -1,4 +1,5 @@
 .head-row {
+  user-select: none;
   background-color: @background-colour;
   height: @head-height;
 


src/assets/styles/base.less
```diff
@@ -1,4 +1,12 @@
 html, body {
```
jleft: I'm not sure that we'd want to apply this to everything.

For example, I'd find it pretty frustrating if I couldn't select an error message/notification to copy it.

Perhaps we should only target the chart/axes etc.?

src/assets/styles/base.less
```diff
@@ -1,4 +1,12 @@
...
+  // Disable text selection
```
jleft: Remove comment

src/assets/styles/base.less
```diff
@@ -1,4 +1,12 @@
...
+  -webkit-touch-callout: none;
```
jleft: Do we need this - it only seems to affect links in mobile Safari?

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout

src/assets/styles/base.less
```diff
@@ -1,4 +1,12 @@
...
+  -webkit-user-select: none;
```
jleft: Seems like a good time to use autoprefixer! #510

src/assets/styles/axis.less
```diff
@@ -1,4 +1,5 @@
 .x-axis-row {
+  user-select: none;
```
jleft: Just wondering if there's a reason to apply to these several classes/ids, rather than just to `#charts` or `#chart-container`?

rjmcneill: I wasn't sure if there was a case where we'd want something within the chart container to be selectable to be honest.
That's why I kept it granular.
Happy enough to change it though if you want :)

jleft: > I wasn't sure if there was a case where we'd want something within the chart container to be selectable to be honest.

Likewise, I can't think of a case where we'd want the select something within the chart at the moment.

I'd be tempted to apply it once at a higher level, then if there if there is an exception, we can override it. The granularity is still there, as we can specify that user select is enabled on an ancestor to take precedence over the inheritance, we're just defaulting to `user-select: none`

rjmcneill: @jleft what about the head menu?
The periods and selected product are currently selectable

jleft: Let's go with this for now :)

We can always update later

jleft:
:+1: thanks @rjmcneill

PR #513, opened 2016-01-05
Only display the legend when the crosshair is displayed.

### LINKED ISSUE #498: tooltip labels are still present when the crosshair is not rendered
capture

I'd expect the labels to disappear when the crosshair is not rendered.

jdhodges:
Fixes #498.


djmiley:
:+1:

jleft:
As this is a change from the designs, I think it's worth getting confirmation from a UX/product perspective, to confirm if this is something which we should be changing. I'll chase this up.

Labelling as blocked until we have a decision.

jleft:
It has been confirmed that we will make this change.

PR #517, opened 2016-01-05
Fix streaming feed close and error message formatting

### LINKED ISSUE #515: Console error shown when switching from source with streaming feed to source with no streaming feed
**Steps to reproduce**
1. Open Developer Tools
2. Select `BTC-USD` product
3. Select `Data Generator` product 

**Observed behaviour**
Console error is shown:
```
Uncaught TypeError: b.streamingNotificationFormatter is not a function
```

**Expected behaviour**
Console error should not be shown.

In this case, a WebSocket close event is being fired - which is what we expect and we don't need to show a message to the user.

---

_This has no functional impact and isn't noticeable unless dev tools are open - as such, treat this as minor._

jleft:
Fixes #515

Still worth investigating data interface and the Coinbase WebSocket further, for a more elegant solution.

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -10,6 +10,7 @@ import coinbaseProducts from './data/coinbase/getProducts';
 import formatProducts from './data/coinbase/formatProducts';
 import notification from './notification/notification';
 import messageModel from './model/notification/message';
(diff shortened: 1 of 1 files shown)


rjmcneill:
Fixes the issue, LGTM :+1:

PR #518, opened 2016-01-05
Displays moving average line in bollinger bands indicator.

### LINKED ISSUE #500: Bollinger bands should include the central moving average line
This was probably removed for cosmetic reasons, however it is an important component of the indicator.

jdhodges431:
Fixes #500.

### DIFF
--- a/src/assets/styles/primary.less
+++ b/src/assets/styles/primary.less
@@ -34,6 +34,11 @@
     fill: @bollinger-fill-colour;
   }
 
+  .average > path.line {
+    stroke: @bollinger-average-colour;
+    stroke-dasharray: 5;
+  }
+
   .area {
     fill: @area-fill;
   }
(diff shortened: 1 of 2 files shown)


djmiley:
Is there UX thoughts on the color of this? i'm guessing it's going to default to d3fc.css styling. 
LGTM other than that.

jdhodges431:
It's definitely worth considering. There isn't anything specific in the UX spec, but some of the images show the colour as being the same as the moving average indicator. Don't think it would make sense for them to be exactly the same colour though.

jleft:
I think it'll default to `@secondary-orange`?

As orange is used for the primary series, I think it could be pretty confusing, especially if the line series is selected. If this is the case, then I'd consider changing its colour to a tertiary (maybe purple or pink) colour.

Using a tertiary colour would fit with the designs too:

> The Tertiary colours are generally used for the indicators.

jleft:
Another option could be to use a dashed line.

jleft:
:+1:

PR #520, opened 2016-01-06
Crosshair hidden and legend emptied on zoom. Crosshair shown again on…

### LINKED ISSUE #475: Crosshair does not snap to point on zoom
Zooming should snap the crosshair to a point.
Current behaviour is shown below:

zoomproblem

jdhodges:
… mouse move.

Fixes #475.

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -55,9 +55,6 @@ export default function() {
         .xLabel('')
         .yLabel('')
         .on('trackingmove', function(updatedCrosshairData) {
-            d3.select('.crosshair')
-                .classed('hidden', false);
-
             if (updatedCrosshairData.length > 0) {
                 dispatch.crosshairChange(updatedCrosshairData[0].datum);
             } else {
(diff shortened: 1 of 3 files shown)


djmiley:
Have you got different settings on your IDE/text editor? It's a bit hard to read :(
But looks like a good fix to me.

jleft:
> Have you got different settings on your IDE/text editor? It's a bit hard to read :(
But looks like a good fix to me.

https://github.com/ScottLogic/d3fc-showcase/pull/520/files?w=1

(?w=1 ignores whitespace changes in the diff :smiley:)

djmiley:
Thanks!

rjmcneill:
LGTM :+1:

src/assets/js/chart/primary.js
```diff
@@ -86,27 +86,30 @@ export default function() {
...
-              dispatch.crosshairChange(undefined);
-          }
-      })
-      .on('trackingend', function() {
-          dispatch.crosshairChange(undefined);
-      });
+        .xLabel('')
+        .yLabel('')
+        .on('trackingmove', function(updatedCrosshairData) {
+            d3.select('.crosshair')
```
jleft: Within event handlers, `this` is usually a reference to the element which the event came from.

```js
d3.select(this)
    .select('.crosshair')
    .classed('hidden', false);.
```

Wherever possible we should avoid using a global `d3.select`:
- It's less efficient than selecting a node or a subset of the DOM (as it has to traverse (more of) the DOM; potentially the entire DOM)
- A component may start to manipulate elements which aren't it's concern. The state of the DOM would not necessarily match the state of its associated model and it can become very hard to maintain and potentially lead to some unpredictable behaviour.

src/assets/js/chart/primary.js
```diff
@@ -222,29 +225,32 @@ export default function() {
...
         var zoom = zoomBehavior(zoomWidth)
-          .scale(xScale)
-          .trackingLatest(model.trackingLatest)
-          .on('zoom', function(domain) {
-              dispatch[event.viewChange](domain);
-          });
+            .scale(xScale)
+            .trackingLatest(model.trackingLatest)
+            .on('zoom', function(domain) {
+                d3.select('.crosshair')
```
jleft: (Similar comment to above - avoid global `d3.select`. It's not ideal, but `selection` is more relevant here.)

However, it doesn't feel right to me that zoom is selecting elements which belong to the crosshair and is manipulating them. It the crosshair's responsibility, not zoom's.

Would it be possible to dispatch a relevant event to app, manipulate the model appropriately and render. Or, are there complication with this approach?

src/assets/js/chart/primary.js
```diff
@@ -209,14 +196,15 @@ export default function() {
...
         var zoom = zoomBehavior(zoomWidth)
-          .scale(xScale)
-          .trackingLatest(model.trackingLatest)
-          .on('zoom', function(domain) {
-              dispatch[event.viewChange](domain);
-          });
+            .scale(xScale)
+            .trackingLatest(model.trackingLatest)
+            .on('zoom', function(domain) {
+                dispatch.crosshairChange(undefined);
```
djmiley: When we dispatch `event.viewChange`, do we ever want to see the crosshair? i.e; can they be thought of as the same event?

jleft:
https://github.com/ScottLogic/d3fc-showcase/issues/475#issuecomment-172587636:

> Looks like we'll be getting a fix for this for free in the next d3fc release

jleft:
> Looks like we'll be getting a fix for this for free in the next d3fc release

Thanks for this @jdhodges, but we should now have this with 6768f60545ad1889f72262ee937886499991189e (d3fc 5.3.0), as such I'm closing as I think this is now unnecessary. If anyone disagrees, please reopen.

PR #525, opened 2016-01-08
Limited zoom to 5 periods

### LINKED ISSUE #168: Zoomed in extent is unlimited (Restrict zoom to 5 periods width)

### LINKED ISSUE #499: Limit the zoom levels
Currently you can zoom in to a very small time range - can this be limited?

### LINKED ISSUE #521: Navigator brush should stop event propagating to zoom
When interacting with the brush, the event is also propagated to the zoom component. The brush should take priority in this case and only the brush should handle this interaction. 

https://github.com/mbostock/d3/wiki/Selections#d3_event

rjmcneill:
Relates to #499 
Relates to #168

### DIFF
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -2,8 +2,7 @@ import d3 from 'd3';
 import fc from 'd3fc';
 import util from '../util/util';
 
-export default function(width) {
-
+export default function(containerWidth) {
     var dispatch = d3.dispatch('zoom');
 
     var zoomBehavior = d3.behavior.zoom();
@@ -12,6 +11,8 @@ export default function(width) {
     var allowPan = true;
     var allowZoom = true;
     var trackingLatest = true;
+
+    var minimumPeriods = 5;
 
     function controlPan(zoomExtent) {
         // Don't pan off sides
@@ -39,43 +40,47 @@ export default function(width) {
         zoomBehavior.scale(1);
     }
 
-    function zoom(selection) {
+    function checkMinPeriods(domain) {
+        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
+        return ((domain[1].getTime() - domain[0].getTime()) >= (selectedPeriod * minimumPeriods));
+    }
 
+    function zoom(selection) {
         var xExtent = fc.util.extent()
-          .fields('date')(selection.datum().data);
+            .fields('date')(selection.datum().data);
 
         zoomBehavior.x(scale)
-          .on('zoom', function() {
-              var min = scale(xExtent[0]);
-              var max = scale(xExtent[1]);
+            .on('zoom', function() {
+                var min = scale(xExtent[0]);
+                var max = scale(xExtent[1]);
 
-              var maxDomainViewed = controlZoom([min, max - width]);
-              var panningRestriction = controlPan([min, max - width]);
-              translateXZoom(panningRestriction);
+                var maxDomainViewed = controlZoom([min, max - containerWidth]);
+                var panningRestriction = controlPan([min, max - containerWidth]);
+                translateXZoom(panningRestriction);
 
-              var panned = (zoomBehavior.scale() === 1);
-              var zoomed = (zoomBehavior.scale() !== 1);
+                var panned = (zoomBehavior.scale() === 1);
+                var zoomed = (zoomBehavior.scale() !== 1);
 
-              if ((panned && allowPan) || (zoomed && allowZoom)) {
-                  var domain = scale.domain();
-                  if (maxDomainViewed) {
-                      domain = xExtent;
-                  } else if (zoomed && trackingLatest) {
-                      domain = util.domain.moveToLatest(domain, selection.datum().data);
-                  }
+                if ((panned && allowPan) || (zoomed && allowZoom)) {
+                    var domain = scale.domain();
 
-                  if (domain[0].getTime() !== domain[1].getTime()) {
-                      dispatch.zoom(domain);
-                  } else {
-                      // Ensure the user can't zoom-in infinitely, causing the chart to fail to render
-                      // #168, #411
-                      resetBehaviour();
-                  }
-              } else {
-                  resetBehaviour();
-              }
-          });
+                    if (maxDomainViewed) {
+                        domain = xExtent;
+                    } else if (zoomed && trackingLatest) {
+                        domain = util.domain.moveToLatest(domain, selection.datum().data);
+                    }
 
+                    if (checkMinPeriods(domain)) {
+                        dispatch.zoom(domain);
+                    } else {
+                        // Ensure the user can't zoom-in infinitely, causing the chart to fail to render
+                        // #168, #411, #499
+                        resetBehaviour();
+                    }
+                } else {
+                    resetBehaviour();
+                }
+            });
         selection.call(zoomBehavior);
     }
 
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -49,8 +49,11 @@ export default function(containerWidth) {
                 var min = scale(xExtent[0]);
                 var max = scale(xExtent[1]);
 
-                var minimumPeriodMilliseconds = selection.datum().period.seconds *
-                    minimumVisiblePeriods * 1000;
+                var modelMinimumMilliseconds = selection.datum().period.seconds *
+                    selection.datum().minimumVisiblePeriods * 1000;
+                var xExtentMilliseconds = util.domain.domainMilliseconds(xExtent);
+
+                var minimumPeriodMilliseconds = xExtentMilliseconds < modelMinimumMilliseconds ? xExtentMilliseconds : modelMinimumMilliseconds;
 
                 var maxDomainViewed = controlZoom([min, max - containerWidth]);
                 var panningRestriction = controlPan([min, max - containerWidth]);
@@ -66,10 +69,12 @@ export default function(containerWidth) {
                         domain = xExtent;
                     } else if (zoomed && trackingLatest) {
                         domain = util.domain.moveToLatest(domain,
-                            selection.datum().data);
+                            selection.datum().data,
+                            selection.datum().minimumVisiblePeriods,
+                            selection.datum().period.seconds);
                     }
 
-                    if ((domain[1].getTime() - domain[0].getTime()) >= minimumPeriodMilliseconds) {
+                    if (util.domain.domainMilliseconds(domain) >= modelMinimumMilliseconds) {
                         dispatch.zoom(domain);
                     } else {
                         // Ensure the user can't zoom-in infinitely, causing
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -38,26 +38,21 @@ export default function(containerWidth) {
         zoomBehavior.scale(1);
     }
 
-    function checkMinPeriods(domain, selectedPeriod, minimumPeriods) {
-        return ((domain[1].getTime() - domain[0].getTime()) >= (selectedPeriod * minimumPeriods));
+    function extentLessThanMinimumPeriods(domain, minimumPeriodMilliSeconds) {
+        return ((domain[1].getTime() - domain[0].getTime()) >= minimumPeriodMilliSeconds);
     }
 
     function zoom(selection) {
         var xExtent = fc.util.extent()
             .fields('date')(selection.datum().data);
 
-        var selectedPeriod;
-        var minimumPeriods;
-
         zoomBehavior.x(scale)
             .on('zoom', function() {
                 var min = scale(xExtent[0]);
                 var max = scale(xExtent[1]);
 
-                if (selection.datum().period && selection.datum().minimumPeriods) {
-                    selectedPeriod = selection.datum().period.seconds * 1000;
-                    minimumPeriods = selection.datum().minimumPeriods;
-                }
+                var minimumPeriodMilliSeconds = selection.datum().period.seconds *
+                    selection.datum().minimumPeriods * 1000;
 
                 var maxDomainViewed = controlZoom([min, max - containerWidth]);
                 var panningRestriction = controlPan([min, max - containerWidth]);
@@ -72,13 +67,15 @@ export default function(containerWidth) {
                     if (maxDomainViewed) {
                         domain = xExtent;
                     } else if (zoomed && trackingLatest) {
-                        domain = util.domain.moveToLatest(domain, selection.datum().data);
+                        domain = util.domain.moveToLatest(domain,
+                            selection.datum().data);
                     }
 
-                    if (checkMinPeriods(domain, selectedPeriod, minimumPeriods)) {
+                    if (extentLessThanMinimumPeriods(domain, minimumPeriodMilliSeconds)) {
                         dispatch.zoom(domain);
                     } else {
-                        // Ensure the user can't zoom-in infinitely, causing the chart to fail to render
+                        // Ensure the user can't zoom-in infinitely, causing
+                        // the chart to fail to render
                         // #168, #411, #499
                         resetBehaviour();
                     }
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -12,6 +12,8 @@ export default function(containerWidth) {
     var allowZoom = true;
     var trackingLatest = true;
 
+    var minimumVisiblePeriods;
+
     function controlPan(zoomExtent) {
         // Don't pan off sides
         if (zoomExtent[0] >= 0) {
@@ -38,10 +40,6 @@ export default function(containerWidth) {
         zoomBehavior.scale(1);
     }
 
-    function extentLessThanMinimumPeriods(domain, minimumPeriodMilliSeconds) {
-        return ((domain[1].getTime() - domain[0].getTime()) >= minimumPeriodMilliSeconds);
-    }
-
     function zoom(selection) {
         var xExtent = fc.util.extent()
             .fields('date')(selection.datum().data);
@@ -51,8 +49,8 @@ export default function(containerWidth) {
                 var min = scale(xExtent[0]);
                 var max = scale(xExtent[1]);
 
-                var minimumPeriodMilliSeconds = selection.datum().period.seconds *
-                    selection.datum().minimumPeriods * 1000;
+                var minimumPeriodMilliseconds = selection.datum().period.seconds *
+                    minimumVisiblePeriods * 1000;
 
                 var maxDomainViewed = controlZoom([min, max - containerWidth]);
                 var panningRestriction = controlPan([min, max - containerWidth]);
@@ -71,7 +69,7 @@ export default function(containerWidth) {
                             selection.datum().data);
                     }
 
-                    if (extentLessThanMinimumPeriods(domain, minimumPeriodMilliSeconds)) {
+                    if ((domain[1].getTime() - domain[0].getTime()) >= minimumPeriodMilliseconds) {
                         dispatch.zoom(domain);
                     } else {
                         // Ensure the user can't zoom-in infinitely, causing
@@ -115,6 +113,14 @@ export default function(containerWidth) {
             return scale;
         }
         scale = x;
+        return zoom;
+    };
+
+    zoom.minimumVisiblePeriods = function(x) {
+        if (!arguments.length) {
+            return minimumVisiblePeriods;
+        }
+        minimumVisiblePeriods = x;
         return zoom;
     };
 
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -12,8 +12,6 @@ export default function(containerWidth) {
     var allowZoom = true;
     var trackingLatest = true;
 
-    var minimumPeriods = 5;
-
     function controlPan(zoomExtent) {
         // Don't pan off sides
         if (zoomExtent[0] >= 0) {
@@ -40,8 +38,7 @@ export default function(containerWidth) {
         zoomBehavior.scale(1);
     }
 
-    function checkMinPeriods(domain) {
-        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
+    function checkMinPeriods(domain, selectedPeriod, minimumPeriods) {
         return ((domain[1].getTime() - domain[0].getTime()) >= (selectedPeriod * minimumPeriods));
     }
 
@@ -49,10 +46,18 @@ export default function(containerWidth) {
         var xExtent = fc.util.extent()
             .fields('date')(selection.datum().data);
 
+        var selectedPeriod;
+        var minimumPeriods;
+
         zoomBehavior.x(scale)
             .on('zoom', function() {
                 var min = scale(xExtent[0]);
                 var max = scale(xExtent[1]);
+
+                if (selection.datum().period && selection.datum().minimumPeriods) {
+                    selectedPeriod = selection.datum().period.seconds * 1000;
+                    minimumPeriods = selection.datum().minimumPeriods;
+                }
 
                 var maxDomainViewed = controlZoom([min, max - containerWidth]);
                 var panningRestriction = controlPan([min, max - containerWidth]);
@@ -70,7 +75,7 @@ export default function(containerWidth) {
                         domain = util.domain.moveToLatest(domain, selection.datum().data);
                     }
 
-                    if (checkMinPeriods(domain)) {
+                    if (checkMinPeriods(domain, selectedPeriod, minimumPeriods)) {
                         dispatch.zoom(domain);
                     } else {
                         // Ensure the user can't zoom-in infinitely, causing the chart to fail to render
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -156,27 +156,10 @@ export default function() {
         } else {
             var centrePoint = (brush.extent()[0][0].getTime() +
                 brush.extent()[1][0].getTime()) / 2;
-
             var minPeriodDistance = minimumPeriodMilliSeconds / 2;
 
-            var beforeFirstDataPoint = (centrePoint - minPeriodDistance) <
-                navChart.xDomain()[0].getTime();
-
-            var afterLastDataPoint = (centrePoint + minPeriodDistance) >
-                navChart.xDomain()[1].getTime();
-
-            if (beforeFirstDataPoint) {
-                overShoot = centrePoint - minPeriodDistance - navChart.xDomain()[0].getTime();
-                newBrush[0] = navChart.xDomain()[0];
-                newBrush[1] = new Date(centrePoint + minPeriodDistance - overShoot);
-            } else if (afterLastDataPoint) {
-                overShoot = centrePoint + minPeriodDistance - navChart.xDomain()[1].getTime();
-                newBrush[0] = new Date(centrePoint - minPeriodDistance - overShoot);
-                newBrush[1] = navChart.xDomain()[1];
-            } else {
-                newBrush[0] = new Date(centrePoint - minPeriodDistance);
-                newBrush[1] = new Date(centrePoint + minPeriodDistance);
-            }
+            newBrush[0] = new Date(centrePoint - minPeriodDistance);
+            newBrush[1] = new Date(centrePoint + minPeriodDistance);
         }
 
         return newBrush;
@@ -220,18 +203,18 @@ export default function() {
         }).on('brushend', function() {
             d3.event.sourceEvent.stopPropagation();
             var brushExtentIsEmpty = xEmpty(brush);
-            var newBrush;
+            var minimumBrush;
             setHide(selection, false);
 
             if (brushExtentIsEmpty) {
                 dispatch[event.viewChange](centerOnDate([originalExtent[0][0],
                     originalExtent[1][0]], model.data, brush.extent()[0][0]));
             } else if (!extentLessThanMinimumPeriods(minimumPeriodMilliSeconds)) {
-                newBrush = setToMinPeriods(minimumPeriodMilliSeconds);
-                var centreDate = new Date((newBrush[1].getTime() + newBrush[0].getTime()) / 2);
-                dispatch[event.viewChange](centerOnDate([newBrush[0], newBrush[1]],
+                minimumBrush = setToMinPeriods(minimumPeriodMilliSeconds);
+                var centreDate = new Date((minimumBrush[1].getTime() + minimumBrush[0].getTime()) / 2);
+
+                dispatch[event.viewChange](centerOnDate(minimumBrush,
                     model.data, centreDate));
-                // dispatch[event.viewChange]([newBrush[0], newBrush[1]]);
             }
         });
 
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -130,46 +130,29 @@ export default function() {
             .call(brushMask);
     }
 
-    function extentLessThanMinimumPeriods(minimumPeriodMilliSeconds) {
-        return (brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime() >=
-            minimumPeriodMilliSeconds);
-    }
+    function setBrushExtentToMinPeriods(brushExtent, originalBrushExtent, minimumPeriodMilliseconds) {
+        var leftHandleMoved = brushExtent[0][0].getTime() !== originalBrushExtent[0].getTime();
+        var rightHandleMoved = brushExtent[1][0].getTime() !== originalBrushExtent[1].getTime();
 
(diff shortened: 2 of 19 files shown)


src/assets/js/chart/nav.js
```diff
@@ -85,6 +88,47 @@ export default function() {
...
+        if (brush.extent()[0][0].getTime() > originalExtent[0][0].getTime() && brush.extent()[1][0].getTime() === originalExtent[1][0].getTime()) {
+            newBrush[0] = new Date(brush.extent()[1][0].getTime() - selectedPeriod * minimumPeriods);
+            newBrush[1] = brush.extent()[1][0];
+        // If they have just moved the right handle
+        } else if (brush.extent()[1][0].getTime() < originalExtent[1][0].getTime() && brush.extent()[0][0].getTime() === originalExtent[0][0].getTime()) {
+            newBrush[0] = brush.extent()[0][0];
+            newBrush[1] = new Date(brush.extent()[0][0].getTime() + selectedPeriod * minimumPeriods);
+        // If they have moved both handles
+        } else {
+            var centrePoint = (((brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime()) / 2) + brush.extent()[0][0].getTime());
```
djmiley: Can you use `util.domain.centerOnDate` here?

src/assets/js/chart/nav.js
```diff
@@ -109,15 +158,21 @@ export default function() {
...
-                        model.data, brush.extent()[0][0]));
-                }
-            });
+        }).on('brushend', function() {
+            d3.event.sourceEvent.stopPropagation();
+            var brushExtentIsEmpty = xEmpty(brush);
+            setHide(selection, false);
+
+            if (brushExtentIsEmpty) {
+                var previousWidth = originalExtent[1][0].getTime() - originalExtent[0][0].getTime();
```
djmiley: Could possibly use `util.domain.centerOnDate` aswell?

src/assets/js/chart/nav.js
```diff
@@ -85,6 +88,47 @@ export default function() {
...
+        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
+        return ((brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime() >= (selectedPeriod * minimumPeriods)));
+    }
+
+    function setMinPeriods() {
+        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
+        var newBrush = [];
+        var overShoot = 0;
+
+        // If they have just moved the left handle
```
jdhodges431: Could the content of these comments be communicated by other means? Perhaps something like this:
```
var movedLeftHandle = brush.extent()[0][0].getTime() > originalExtent[0][0].getTime() &&
    brush.extent()[1][0].getTime() === originalExtent[1][0].getTime();
```

Then:
` if (movedLeftHandle) { ...`

src/assets/js/behavior/zoom.js
```diff
@@ -39,43 +40,47 @@ export default function(width) {
         zoomBehavior.scale(1);
     }
 
-    function zoom(selection) {
+    function checkMinPeriods(domain) {
+        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
```
djmiley: You can pass period into zoom on the model

src/assets/js/chart/nav.js
```diff
@@ -85,6 +88,47 @@ export default function() {
         return ((navBrush.extent()[0][0] - navBrush.extent()[1][0]) === 0);
     }
 
+    function checkMinPeriods() {
+        var selectedPeriod = d3.select('.head-menu').datum().selectedPeriod.seconds * 1000;
```
djmiley: You can pass period into zoom on the model.

src/assets/js/chart/nav.js
```diff
@@ -109,15 +158,21 @@ export default function() {
...
+            setHide(selection, false);
+
+            if (brushExtentIsEmpty) {
+                var previousWidth = originalExtent[1][0].getTime() - originalExtent[0][0].getTime();
+                var newBrush = [new Date(brush.extent()[0][0].getTime() - previousWidth / 2),
+                    new Date(brush.extent()[1][0].getTime() + previousWidth / 2)];
+
+                dispatch[event.viewChange]([newBrush[0], newBrush[1]]);
+            } else if (!checkMinPeriods()) {
+                setMinPeriods();
```
djmiley: Feels better to return a domain array from a calculating function and then dispatch it.

src/assets/js/chart/nav.js
```diff
@@ -85,6 +88,47 @@ export default function() {
...
+            } else if ((centrePoint + (selectedPeriod * minimumPeriods / 2)) > navChart.xDomain()[1]) {
+                overShoot = (centrePoint + (selectedPeriod * minimumPeriods / 2)) - navChart.xDomain()[1].getTime();
+                newBrush[0] = new Date(centrePoint - (selectedPeriod * minimumPeriods / 2) - overShoot);
+                newBrush[1] = navChart.xDomain()[1];
+            } else {
+                newBrush[0] = new Date(centrePoint - (selectedPeriod * minimumPeriods / 2));
+                newBrush[1] = new Date(centrePoint + (selectedPeriod * minimumPeriods / 2));
+            }
+        }
+
```
jdhodges431: What about extracting out `selectedPeriod * minimumPeriods / 2` since it seems to be used a lot?

This would make it a bit more readable:
```
var distanceFromCentreToEnd = selectedPeriod * minimumPeriods / 2;

var snapsBeforeFirstDataPoint = centrePoint - distanceFromCentreToEnd < navChart.xDomain()[0].getTime();
var snapsAfterLastDataPoint = centrePoint + distanceFromCentreToEnd > navChart.xDomain()[1];

if (snapsBeforeFirstDataPoint) {
    overShoot = (centrePoint - distanceFromCentreToEnd) - navChart.xDomain()[0].getTime();
    newBrush[0] = navChart.xDomain()[0];
    newBrush[1] = new Date(centrePoint + distanceFromCentreToEnd - overShoot);
} else if (snapsAfterLastDataPoint) {
    overShoot = (centrePoint + distanceFromCentreToEnd) - navChart.xDomain()[1].getTime();
    newBrush[0] = new Date(centrePoint - distanceFromCentreToEnd - overShoot);
    newBrush[1] = navChart.xDomain()[1];
} else {
    newBrush[0] = new Date(centrePoint - distanceFromCentreToEnd);
    newBrush[1] = new Date(centrePoint + distanceFromCentreToEnd);
}
```

src/assets/js/chart/nav.js
rjmcneill: @djmiley The problem with that is that the centrePoint still needs to  be calculated and passed into the `util.domain.centerOnDate` and the brush extents (that we could pass in as the domain) are whatever the user has just attempted to zoom to.  If we're hitting this section of code, it means that they've tried to zoom in too far and this extent will need updated, which leads to the `newBrush` array that I'm calculating the values for.

So I'm not sure `util.domain.centerOnDate` is entirely appropriate in this case?
Unless I'm missing something?

src/assets/js/initialiseModel.js
```diff
@@ -119,16 +119,17 @@ export default function() {
...
     var sources = initSources();
     var products = initProducts();
+    var minPeriods = 5;
 
     return {
         periods: periods,
         sources: sources,
         products: products,
-        primaryChart: model.chart.primary(products.generated),
+        primaryChart: model.chart.primary(products.generated, periods.day1, minPeriods),
```
djmiley: Do we need to pass `minPeriods` into secondary charts model for zooming purposes?

src/assets/js/chart/nav.js
```diff
@@ -101,23 +160,40 @@ export default function() {
...
-                }
-            });
+        }).on('brushend', function() {
+            d3.event.sourceEvent.stopPropagation();
+            var brushExtentIsEmpty = xEmpty(brush);
+            var newBrush;
+            setHide(selection, false);
+
+            if (brushExtentIsEmpty) {
+                var previousWidth = originalExtent[1][0].getTime() -
```
djmiley: Does `centerOnDate` not work in this condition case?

src/assets/js/behavior/zoom.js
```diff
@@ -39,43 +38,54 @@ export default function(width) {
...
-          });
+        var selectedPeriod;
+        var minimumPeriods;
 
+        zoomBehavior.x(scale)
+            .on('zoom', function() {
+                var min = scale(xExtent[0]);
+                var max = scale(xExtent[1]);
+
+                if (selection.datum().period && selection.datum().minimumPeriods) {
```
djmiley: Pass `minimumPeriods` in as a parameter on the zoom component?

src/assets/js/behavior/zoom.js
```diff
@@ -39,43 +38,54 @@ export default function(width) {
...
+                if ((panned && allowPan) || (zoomed && allowZoom)) {
+                    var domain = scale.domain();
+
+                    if (maxDomainViewed) {
+                        domain = xExtent;
+                    } else if (zoomed && trackingLatest) {
+                        domain = util.domain.moveToLatest(domain, selection.datum().data);
+                    }
+
+                    if (checkMinPeriods(domain, selectedPeriod, minimumPeriods)) {
```
djmiley: More explicit function name? `extentLessThanMinimumPeriods` possibly?

src/assets/js/chart/nav.js
```diff
@@ -85,8 +87,65 @@ export default function() {
         return ((navBrush.extent()[0][0] - navBrush.extent()[1][0]) === 0);
     }
 
-    function nav(selection) {
+    function checkMinPeriods(selectedPeriod, minimumPeriods) {
```
djmiley: More explicit function name? extentLessThanMinimumPeriods possibly?

djmiley:
In `nav.js` and `zoom.js`, can we just work with the calculated quantity `selectedPeriod * minimumPeriods`?

src/assets/js/chart/nav.js
```diff
@@ -85,8 +87,65 @@ export default function() {
...
+        if (leftHandleMoved) {
+            newBrush[0] = new Date(brush.extent()[1][0].getTime() -
+                selectedPeriod * minimumPeriods);
+            newBrush[1] = brush.extent()[1][0];
+        } else if (rightHandleMoved) {
+            newBrush[0] = brush.extent()[0][0];
+            newBrush[1] = new Date(brush.extent()[0][0].getTime() +
+                selectedPeriod * minimumPeriods);
+        } else {
+            var centrePoint = (((brush.extent()[1][0].getTime() -
```
djmiley: `(brush.extent()[0][0].getTime() + brush.extent()[1][0].getTime()) / 2`

src/assets/js/chart/nav.js
```diff
@@ -85,8 +87,65 @@ export default function() {
...
+            var minPeriodDistance = (selectedPeriod * minimumPeriods / 2);
+
+            var beforeFirstDataPoint = (centrePoint - minPeriodDistance) <
+                navChart.xDomain()[0].getTime();
+
+            var afterLastDataPoint = (centrePoint + minPeriodDistance) >
+                navChart.xDomain()[1].getTime();
+
+            if (beforeFirstDataPoint) {
+                overShoot = (centrePoint - minPeriodDistance) - navChart.xDomain()[0].getTime();
```
djmiley: I think `centerOnDate` should work well here? Or should be modified to accommodate the two cases. There is a lot of similarities between the function and what is going on in here.

test/util/domain/centerOnDateSpec.js
```diff
@@ -97,14 +97,14 @@ describe('util/domain/centerOnDate', function() {
         var centerOnDateDomain = centerOnDate(domain, data, centerDate);
```
rjmcneill: This was the main change in functionality for the centerOnDate function.

Previously, if you attempted to center on a date outside of the range, it would reset to the previous domain.
However, with the updated functionality, it returns the domain as close as possible to the center date provided while keeping the overall width.
I felt this was the best way to handle this.
This will be hit if the user attempts to use the brush to zoom to less than the minimum number of periods right beside either dataExtents.

Hope this makes sense to everyone!

jdhodges431:
LGTM :+1:

src/assets/js/chart/secondary/base.js
```diff
@@ -9,34 +9,40 @@ export default function() {
...
+                .on('zoom', function(domain) {
+                    dispatch[event.viewChange](domain);
+                });
 
             container.select('.plot-area-container')
-              .datum({data: selection.datum()})
+                .datum({
+                    data: selection.datum(),
+                    period: period,
+                    minimumPeriods: minimumPeriods
```
djmiley: I would be more inclined to pass the `minimumPeriods` in as a get/set-able quantity on the zoom component. I'd leave the period in `selection.datum()`.

src/assets/js/chart/nav.js
```diff
@@ -85,9 +87,47 @@ export default function() {
...
-    function nav(selection) {
+    function extentLessThanMinimumPeriods(minimumPeriodMilliSeconds) {
+        return (brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime() >=
+            minimumPeriodMilliSeconds);
+    }
+
+    function setToMinPeriods(minimumPeriodMilliSeconds) {
+        var newBrush = [];
+        var overShoot = 0;
+
```
djmiley: It would read easier to me if the `leftHandleMoved` condition was only a check on that and also `!==` rather than an `<` or `>`. The condition below would be `if(leftHandleMoved & !rightHandleMoved)` etc.

test/util/domain/centerOnDateSpec.js
djmiley: Sounds good, and much more sensible than my original implementation! :)

src/assets/js/chart/nav.js
```diff
@@ -85,9 +87,47 @@ export default function() {
         return ((navBrush.extent()[0][0] - navBrush.extent()[1][0]) === 0);
     }
 
-    function nav(selection) {
+    function extentLessThanMinimumPeriods(minimumPeriodMilliSeconds) {
+        return (brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime() >=
+            minimumPeriodMilliSeconds);
+    }
+
+    function setToMinPeriods(minimumPeriodMilliSeconds) {
```
djmiley: You could pass `minimumViewableTime` into `centerOnDate` as an optional 4th parameter defaulted to 0?

djmiley: And move this into that function.

djmiley:
Definitely like the look of this so far :)

src/assets/js/chart/nav.js
djmiley: Ignore me :)

rjmcneill:
This PR is getting a little out of hand commits wise, happy to squash to one commit when the comments are finished :grinning:

src/assets/js/behavior/zoom.js
```diff
@@ -39,43 +40,53 @@ export default function(width) {
...
-              } else {
-                  resetBehaviour();
-              }
-          });
-
+            .on('zoom', function() {
+                var min = scale(xExtent[0]);
+                var max = scale(xExtent[1]);
+
+                // console.log(selection.datum());
```
djmiley: Remove console.log

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,42 @@ export default function() {
...
+        if (leftHandleMoved && !rightHandleMoved) {
+            newBrush[0] = new Date(brush.extent()[1][0].getTime() - minimumPeriodMilliSeconds);
+            newBrush[1] = brush.extent()[1][0];
+        } else if (rightHandleMoved && !leftHandleMoved) {
+            newBrush[0] = brush.extent()[0][0];
+            newBrush[1] = new Date(brush.extent()[0][0].getTime() + minimumPeriodMilliSeconds);
+        } else {
+            var centrePoint = (brush.extent()[0][0].getTime() +
+                brush.extent()[1][0].getTime()) / 2;
+            var minPeriodDistance = minimumPeriodMilliSeconds / 2;
```
djmiley: Would probably look better with just `minimumPeriodMilliseconds / 2`, rather than declaring a new variable.

src/assets/js/behavior/zoom.js
```diff
@@ -39,43 +40,51 @@ export default function(width) {
         zoomBehavior.scale(1);
     }
 
-    function zoom(selection) {
+    function extentLessThanMinimumPeriods(domain, minimumPeriodMilliSeconds) {
```
djmiley: We only use this once in this file, so I would probably just put this in the `if` statement below.

src/assets/js/chart/nav.js
djmiley: Change to `setBrushToMinimumPeriods(brushExtent, minimumPeriodMilliseconds)`, where `brushExtent = [brush.extent()[0][0], brush.extent()[1][0]]`

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,41 @@ export default function() {
...
+        var rightHandleMoved = brush.extent()[1][0].getTime() !== originalExtent[1][0].getTime();
+
+        if (leftHandleMoved && !rightHandleMoved) {
+            newBrush[0] = new Date(brush.extent()[1][0].getTime() - minimumPeriodMilliSeconds);
+            newBrush[1] = brush.extent()[1][0];
+        } else if (rightHandleMoved && !leftHandleMoved) {
+            newBrush[0] = brush.extent()[0][0];
+            newBrush[1] = new Date(brush.extent()[0][0].getTime() + minimumPeriodMilliSeconds);
+        } else {
+            var centrePoint = (brush.extent()[0][0].getTime() +
```
djmiley: Does this need to be on two lines?

src/assets/js/chart/nav.js
```diff
@@ -142,35 +176,51 @@ export default function() {
         navChart.xDomain(fc.util.extent().fields('date')(model.data))
           .yDomain(yExtent);
 
-        brush.on('brush', function() {
+        brush.on('brushstart', function() {
+            d3.event.sourceEvent.stopPropagation();
+            originalExtent = brush.extent();
```
djmiley: `originalExtent = [brush.extent()[0][0], brush.extent()[1][0]]`

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,41 @@ export default function() {
             .call(brushMask);
     }
 
-    function nav(selection) {
+    function extentLessThanMinimumPeriods(minimumPeriodMilliSeconds) {
```
djmiley: We only use this once in this file, so I would probably just put this in the `if` statement below.

src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,23 +3,29 @@ import fc from 'd3fc';
...
-      .fields('date')(data);
-    var domainTimes = domain.map(function(d) { return d.getTime(); });
-    var domainTimeDifference = (d3.max(domainTimes) - d3.min(domainTimes)) / 1000;
+        .fields('date')(data);
+    var newBrush = [];
+    var overShoot = 0;
 
-    if (centerDate.getTime() < dataExtent[0] || centerDate.getTime() > dataExtent[1]) {
-        return [new Date(d3.min(domainTimes)), new Date(d3.max(domainTimes))];
+    if (domain[0].getTime() > domain[1].getTime()) {
```
djmiley: Comment above this saying why this functionality is here. I've got a good idea why, but I think only because I spent some time on this problem earlier :)

djmiley:
Once these are fixed, it will be looking good to merge imo :)

src/assets/js/chart/nav.js
```diff
@@ -142,35 +170,52 @@ export default function() {
...
-            });
+        .on('brushend', function() {
+            d3.event.sourceEvent.stopPropagation();
+            var brushExtentIsEmpty = xEmpty(brush);
+            var minimumBrush;
+            var brushExtentDelta = brush.extent()[1][0].getTime() - brush.extent()[0][0].getTime();
+            setHide(selection, false);
+
+            if (brushExtentIsEmpty) {
+                dispatch[event.viewChange](centerOnDate([originalExtent[0],
```
djmiley: just `originalExtent' ?

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,35 @@ export default function() {
             .call(brushMask);
     }
 
-    function nav(selection) {
+    function setToMinPeriods(minimumPeriodMilliSeconds) {
```
djmiley: Change method signature to `setBrushExtentToMinPeriods([brush.extent[0][0], brush.extent()[1][0]], originalExtent, minimumPeriodMilliseconds)`

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,35 @@ export default function() {
...
+        var newBrush = [];
+        var overShoot = 0;
```
djmiley: Do you use overshoot anymore?

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,35 @@ export default function() {
             .call(brushMask);
     }
 
-    function nav(selection) {
+    function setToMinPeriods(minimumPeriodMilliSeconds) {
+        var newBrush = [];
```
djmiley: Do we need to use `newBrush` or would it look like with return statements in the if conditions?

src/assets/js/chart/nav.js
djmiley: I mean passing in those parameters as sensibly named variables `(brushExtent, originalBrushExtent, minimumPeriodMilliseconds)`

djmiley:
This look good to me :+1:

src/assets/js/chart/nav.js
```diff
@@ -142,35 +165,52 @@ export default function() {
         navChart.xDomain(fc.util.extent().fields('date')(model.data))
           .yDomain(yExtent);
 
-        brush.on('brush', function() {
+        brush.on('brushstart', function() {
+            d3.event.sourceEvent.stopPropagation();
```
tlsim: Is this required to get the expected behaviour? This [stops propagation](http://javascript.info/tutorial/bubbling-and-capturing) of the underlying `mousedown` DOM event. In general, I would tend to avoid this so as to [avoid breaking](https://css-tricks.com/dangers-stopping-event-propagation/) other `mousedown` listeners in the tree. Same for `mousemove` and `mouseup` for `brush` and `brushend` below.

src/assets/js/behavior/zoom.js
```diff
@@ -40,42 +41,46 @@ export default function(width) {
...
+                        domain = util.domain.moveToLatest(domain,
+                            selection.datum().data);
+                    }
+
+                    if ((domain[1].getTime() - domain[0].getTime()) >= minimumPeriodMilliSeconds) {
+                        dispatch.zoom(domain);
+                    } else {
+                        // Ensure the user can't zoom-in infinitely, causing
+                        // the chart to fail to render
+                        // #168, #411, #499
```
tlsim: Would be great to have more testing around this, then we could remove these comments. But possibly not a simple task, so it's probably OK to leave for now.

src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,23 +3,30 @@ import fc from 'd3fc';
...
-    var timeShift = 0;
-    if (centeredDataDomain[1].getTime() > dataExtent[1].getTime()) {
-        timeShift = (dataExtent[1].getTime() - centeredDataDomain[1].getTime()) / 1000;
-    } else if (centeredDataDomain[0].getTime() < dataExtent[0].getTime()) {
-        timeShift = (dataExtent[0].getTime() - centeredDataDomain[0].getTime()) / 1000;
+    var distanceAroundCenterDate = (domain[1].getTime() - domain[0].getTime()) / 2;
+    newBrush[0] = new Date(centerDate.getTime() - distanceAroundCenterDate);
+    newBrush[1] = new Date(centerDate.getTime() + distanceAroundCenterDate);
+
+    if (newBrush[0].getTime() < dataExtent[0].getTime()) {
```
tlsim: Can compare date objects using `<` and `>` without `getTime`, i.e. `newBrush[0] < dataExtent[0]` would work.

src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,23 +3,30 @@ import fc from 'd3fc';
...
-    var domainTimes = domain.map(function(d) { return d.getTime(); });
-    var domainTimeDifference = (d3.max(domainTimes) - d3.min(domainTimes)) / 1000;
+        .fields('date')(data);
+    var newBrush = [];
+    var overShoot = 0;
 
-    if (centerDate.getTime() < dataExtent[0] || centerDate.getTime() > dataExtent[1]) {
-        return [new Date(d3.min(domainTimes)), new Date(d3.max(domainTimes))];
+    // If the domains are not in chronological order, reorder them
+    if (domain[0].getTime() > domain[1].getTime()) {
```
tlsim: Consider array.reverse()

src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,23 +3,30 @@ import fc from 'd3fc';
...
-      .fields('date')(data);
-    var domainTimes = domain.map(function(d) { return d.getTime(); });
-    var domainTimeDifference = (d3.max(domainTimes) - d3.min(domainTimes)) / 1000;
+        .fields('date')(data);
+    var newBrush = [];
+    var overShoot = 0;
 
-    if (centerDate.getTime() < dataExtent[0] || centerDate.getTime() > dataExtent[1]) {
-        return [new Date(d3.min(domainTimes)), new Date(d3.max(domainTimes))];
+    // If the domains are not in chronological order, reorder them
```
tlsim: Can remove comment.

src/assets/js/behavior/zoom.js
rjmcneill: I'm happy enough to look at writing some tests for this once I finish up what I'm currently working on Tom :)
Could be something nice to do before the next sprint starts

src/assets/js/chart/nav.js
rjmcneill: This isn't needed for the behaviour, but @jleft suggested this might be worth added to stop the multiple events propagating.
He might be best suited to answer!

src/assets/js/behavior/zoom.js
tlsim: Sounds good :smile:

src/assets/js/util/domain/centerOnDate.js
djmiley: I should say I thought a comment would be good here, so asked Rob for one.

tlsim: Ah ok, if it's not totally clear from the code, then keep, changing 'domains' to 'left and right hand limits of domain', or similar.

src/assets/js/chart/nav.js
jleft: The issue was that some interactions (e.g. click and drag on a navigator handle) on the navigator were being handled by both zoom and the brush - I think this was causing some occasionally weird behaviour and was making it pretty hard to debug (correct me if I'm wrong @rjmcneill!). I think zoom and brush shouldn't be handling the same interaction simultaneously - stopping the event propagating is one way to prevent this.

It is also something which is [mentioned in the D3 docs](https://github.com/mbostock/d3/wiki/Drag-Behavior) (although a slightly different scenario, for the drag behaviour rather than brush):

> When combining drag behaviors with other event listeners for interaction events (such as having drag take precedence over zoom), you may also consider stopping propagation on the source event to prevent multiple actions:
>
> ```js
> drag.on("dragstart", function() {
>   d3.event.sourceEvent.stopPropagation(); // silence other listeners
> });
> ```

However, I agree with @tlsim - `stopPropagation` is something best avoided. If we don't need to do this, then let's not. If we do need it, then there might be a way to achieve this without using `stopPropagation`?

rjmcneill: Yea, when we first started looking into it, there was strange behaviour and the multiple events being triggered were making it very difficult to track down any issues.

I'll give it a proper test without the ```stopPropagation()``` and push up changes in a few

tlsim: Yeah, I think a separate PR to resolve this makes sense, especially if the weird behaviour is pre-existing, and not introduced by limiting the zoom (the only open issue related to this I can find is #573). For #521 - It's not clear to me exactly which events need special attention when combining brush with zoom. Do we still want double tap, pinch zoom, and mousewheel on the brush extent to be handled by zoom? There's already logic in our zoom behavior to disable d3 zoom behavior's panning interaction in the navbar. Might be worth looking at that too.

tlsim:
:+1:

src/assets/js/chart/nav.js
```diff
@@ -124,9 +126,30 @@ export default function() {
...
+                new Date(brushExtent[0][0].getTime() + minimumPeriodMilliSeconds)];
+        } else {
+            var centrePoint = (brushExtent[0][0].getTime() + brushExtent[1][0].getTime()) / 2;
+
+            return [new Date(centrePoint - minimumPeriodMilliSeconds / 2),
+                new Date(centrePoint + minimumPeriodMilliSeconds / 2)];
+        }
+    }
+
+    function nav(selection, period) {
```
jleft: `period` isn't used?

`brushHide` on line 163 is also unused?

src/assets/js/util/domain/centerOnDate.js
```diff
@@ -3,23 +3,28 @@ import fc from 'd3fc';
 
 export default function(domain, data, centerDate) {
     var dataExtent = fc.util.extent()
-      .fields('date')(data);
-    var domainTimes = domain.map(function(d) { return d.getTime(); });
-    var domainTimeDifference = (d3.max(domainTimes) - d3.min(domainTimes)) / 1000;
+        .fields('date')(data);
+    var newBrush = [];
```
jleft: As a name `newBrush` seems out of context here, maybe `centeredDomain` would work?

src/assets/js/behavior/zoom.js
```diff
@@ -40,42 +41,46 @@ export default function(width) {
...
-              } else {
-                  resetBehaviour();
-              }
-          });
-
+            .on('zoom', function() {
+                var min = scale(xExtent[0]);
+                var max = scale(xExtent[1]);
+
+                var minimumPeriodMilliSeconds = selection.datum().period.seconds *
```
jleft: MilliSeconds => milli**s**econds

(grep for this, should be changed in a few places)

src/assets/js/chart/nav.js
```diff
@@ -142,35 +165,49 @@ export default function() {
         navChart.xDomain(fc.util.extent().fields('date')(model.data))
           .yDomain(yExtent);
 
-        brush.on('brush', function() {
+        brush.on('brushstart', function() {
+            originalExtent = [brush.extent()[0][0], brush.extent()[1][0]];
```
jleft: It's a shame we have to deal with this 2-dimensional array so frequently. I think one of the dimensions (corresponding to the y-scale) is irrelevant to us.

As we're using cartesian chart, it will set the y-scale on the brush, which I don't think we need to do.

Not something to do now, but as a future enhancement I'd be interested to see if this can be simplified by removing cartesian chart and rolling our own.

jleft:
Looks really good to me, just a few comments above and a couple of questions:

- Would this implementation work if data.length < minimumPeriods?
- I'm not sure about the naming of `minimumPeriods`, would `minimumVisiblePeriods`  or `minimumVisibleDatums` make sense?

rjmcneill:
Changed the ```minimumPeriods ``` to ```minimumVisiblePeriods ```
The data.length question is a different problem.
The issue is that domain is too small.
I'll add some checking around it to make sure the domain we're setting it to is less than the full domain

jleft:
> Changed the minimumPeriods to minimumVisiblePeriods

Thank you :)

> The data.length question is a different problem.
The issue is that domain is too small.
I'll add some checking around it to make sure the domain we're setting it to is less than the full domain

If this is a big change, I'd be open to adding a guard to make sure we never get into this situation, otherwise that sounds good to me.

jackarnstein:
When you double click on the candlesticks to zoom in, once only 5 sticks are in view, continuing to click will allow you to zoom in a bit more. At no point do less than 5 candlesticks appear. I think its purely an aesthetics issue which just looks a little wrong, it doesnt seem to have any negative side effects.

[candlestickzoom.avi](https://files.zenhub.io/56b0e0b6c4f4fa554c4ff90d)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

#573

PR #526, opened 2016-01-08
Navbar brush highlight using Mask

### LINKED ISSUE #343: Navigator gradient colour

WPFerg:
Fixes #343 

Adds an off-screen mask, which the brush uses to add a highlight to the chart

### DIFF
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -72,6 +72,16 @@ export default function() {
               return this.data;
           }
       });
+
+    var maskXScale = fc.scale.dateTime();
+    var maskYScale = d3.scale.linear()
+      .range([navChartHeight, 0]);
+
+    var brushMask = fc.series.area()
+      .yValue(function(d) { return d.close; })
+      .xScale(maskXScale)
+      .yScale(maskYScale);
+
     var layoutWidth;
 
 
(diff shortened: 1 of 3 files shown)


src/assets/styles/variables.less
```diff
@@ -90,7 +91,8 @@
 // Changes to height or bottom-margin must also be carried over in nav.js variables
 @navigator-height: 100px;
 @navigator-bottom-margin-from-javascript: 40px;   //Temporary variable until fully ported to flex
-@navigator-extent-colour: @primary-white;
+@navigator-extent-colour-top: @secondary-orange;
+@navigator-extent-colour-bottom: @primary-white;
```
jdhodges431: Rather than white I'd make this a lighter shade of orange.

`@navigator-extent-colour-bottom: lighten(@secondary-orange, 30%);`

jdhodges431:
Looks good to me, nice work :+1:

src/assets/styles/variables.less
WPFerg: Done, shouldn't make too much difference to using a `stop-opacity`

djmiley:
Definitely a much neater approach! :)

src/assets/js/chart/nav.js
```diff
@@ -85,9 +95,35 @@ export default function() {
         return ((navBrush.extent()[0][0] - navBrush.extent()[1][0]) === 0);
     }
 
+    function createDefs(selection, data) {
+        var defsEnter = selection.selectAll('defs').data([0]).enter().append('defs');
```
jleft: Minor style comment - one method call per line, please:

```js
selection.selectAll('defs')
    .data([0])
    .enter()
    .append('defs');
```

src/assets/js/chart/nav.js
```diff
@@ -85,9 +95,35 @@ export default function() {
...
+              <rect class="mask-background"></rect> \
+          </mask>');
+
+        selection.select('.mask-background').attr({
+            width: layoutWidth,
+            height: navChartHeight
+        });
+
+        maskXScale.domain(fc.util.extent().fields('date')(data))
+          .range([0, layoutWidth]);
```
jleft: Can we set the range in `nav.dimensionChanged`?

src/assets/styles/navigator.less
```diff
@@ -4,10 +4,10 @@
     margin-right: 0 !important;
 
     .extent {
-      fill: @navigator-extent-colour;
       stroke: none;
-      fill-opacity: 0.125;
       shape-rendering: crispEdges;
+      mask: url('/#brush-mask');
+      fill: url('/#brush-gradient');
```
jleft: See https://github.com/ScottLogic/d3fc-showcase/pull/418#discussion_r46283126

> If we use an absolute link then it won't work if the site is run from a path which isn't /; e.g. when deploying to gh-pages, it's /d3fc-showcase/

A relative path would be preferable, but is doesn't seem to work in Firefox.

I expect that inlining the styles will work. A good place to do this would be in `navMulti`'s `decorate`:

```js
// Something like this...
enter.select('.extent')
    .attr('mask', 'url("#brush-mask")')
    .attr('fill', 'url("#brush-gradient")');
```

jleft:
Thanks for this @WPFerg - really nice solution :+1: 

Just a few comments above, then it should be good to merge :)

src/assets/styles/navigator.less
WPFerg: Inlining the styles did work, I moved it to plain CSS on the refactor -- I hadn't tried running on non `/` though. I'll swap it back :)

WPFerg:
Changes made, hopefully good to merge now

jleft:
> Changes made, hopefully good to merge now

Nice work, looks great :) thanks again!

PR #530, opened 2016-01-09
Fix reset button to be clickable when tracking latest price

### LINKED ISSUE #529: Reset/tracking latest button
**Steps to reproduce**

1. Open the application in a web browser
2. Ensure the visible domain includes the latest price (so the viewport is tracking latest price) - this is the default
3. Adjust the zoom level

screen shot 2016-01-09 at 21 13 25

**Observed behaviour**

- The button appears not to be clickable, as a default cursor is shown on hover
- The button will only respond to clicks on the button stroke/edge, the main part of the button is not clickable
- Clicking on the button's stroke resets to the default visible domain

**Expected behaviour**

- A `pointer` cursor should be shown on hover, in both active and inactive states
- The entire button should be clickable in both states

jleft:
Also ensures pointer cursor is only shown on path element not entire container

Fixes #529

### DIFF
--- a/src/assets/styles/navigator.less
+++ b/src/assets/styles/navigator.less
@@ -53,15 +53,21 @@
   }
 
   #navbar-reset {
-    stroke: @primary-white;
     flex: 0 0 24px;
     height: 40px;
     opacity: 0.2;
     margin: 10px 21px 0 15px;
-    fill: none;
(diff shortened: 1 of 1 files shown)


djmiley:
:+1:

PR #535, opened 2016-01-11
Transforms text in close price callout

### LINKED ISSUE #497: (minor) padding around the current price callout is not consistent
capture

Looks like the text baseline needs to be altered a little?

djmiley:
Fixes #497

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -191,13 +191,17 @@ export default function() {
         var tickValues = produceAnnotatedTickValues(yScale, [latestPrice]);
         primaryChart.yTickValues(tickValues)
           .yDecorate(function(s) {
-              s.selectAll('.tick')
+              var closePriceTick = s.selectAll('.tick')
                 .filter(function(d) { return d === latestPrice; })
-                .classed('closeLine', true)
-                .select('path')
+                .classed('closeLine', true);
+
+              var calloutHeight = 18;
+              closePriceTick.select('path')
                 .attr('d', function(d) {
-                    return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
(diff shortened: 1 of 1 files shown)


src/assets/js/chart/primary.js
```diff
@@ -223,13 +223,16 @@ export default function() {
...
-              s.selectAll('.tick')
+              var closePriceTick = s.selectAll('.tick')
                 .filter(function(d) { return d === latestPrice; })
-                .classed('closeLine', true)
-                .select('path')
+                .classed('closeLine', true);
+
+              closePriceTick.select('path')
                 .attr('d', function(d) {
                     return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
```
jleft: `14` is the height of the callout, which made sense when the font was smaller, but now the the font-size is `16px`, is it worth increasing this callout's height? Maybe `18` or `20` would look better?

src/assets/js/chart/primary.js
```diff
@@ -223,13 +223,16 @@ export default function() {
...
-                .select('path')
+                .classed('closeLine', true);
+
+              closePriceTick.select('path')
                 .attr('d', function(d) {
-                    return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 14));
+                    return d3.svg.area()(calculateCloseAxisTagPath(yAxisWidth, 18));
                 });
+              closePriceTick.select('text')
+                .attr('transform', 'translate(9,1)');
```
jleft: Sorry, one final comment - I think `9` works as it's half the height of the callout?

If this is the case, please make we use meaningful vars, so it's not quite such a 'magic' number?

```js
var calloutHeight = 18;
// ...
```

Or is it just a coincidence?

Issue #523, opened 2016-01-08
Investigate inclusion in OpenFinD3FC


jleft:
Investigate and identify if the showcase's structure and API can be improved for easier inclusion in the OpenFin integration.

https://github.com/owennw/OpenFinD3FC

/cc @owennw @bjedrzejewski @ColinEberhardt

jleft:
I expect we may have to revisit the current integration and make some (hopefully small) changes as the projects progress, but as #537 and https://github.com/owennw/OpenFinD3FC/pull/256 have been merged, I'll close this.

PR #536, opened 2016-01-12
[WIP] Updates for easier integration in Openfin

jleft:
Needs to tidied and there are a few issues, but I think it's worth seeing if this is heading in the right direction.

- [ ] Consider bundling module with deps
- [ ] Remove `.updateModel()` from app?
- [ ] Seems pointless making the request for Coinbase products, when they aren't needed - add some way to disable this?

Longer-term, publishing to npm would make consumption much easier.

#523

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -254,7 +254,15 @@ module.exports = function(grunt) {
         rollup: {
             options: {
                 format: 'umd',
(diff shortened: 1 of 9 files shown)


jleft:
Closing this for now as OpenFinD3FC depends on this branch.

This work is continued in #537

PR #537, opened 2016-01-12
Updates for easier integration in Openfin

jleft:
Needs to tidied and there are a few issues, but I think it's worth seeing if this is heading in the right direction.

- [x] Remove `.updateModel()` from app?
- [x] Remove initial model param from app?
- [x] Seems pointless making the request for Coinbase products, when they aren't needed - add some way to disable this?

Longer-term, publishing to npm would make consumption much easier.

#523

TODO: bump d3fc version in OpenFin

The API exposed is intentionally very simple and inflexible, but with the intention of making it as easy-to-use as possible for the OpenFin project.


src/assets/js/app.js
```diff
@@ -6,55 +6,80 @@ import menu from './menu/menu';
...
-import formatProducts from './data/coinbase/formatProducts';
 import notification from './notification/notification';
 import messageModel from './model/notification/message';
+import dataModel from './model/data/data';
 import coinbaseStreamingErrorResponseFormatter from './data/coinbase/streaming/errorResponseFormatter';
-
-export default function(initialModel) {
+import initialiseModel from './initialiseModel';
+import coinbaseGetProducts from './data/coinbase/getProducts';
+import coninbaseFormatProducts from './data/coinbase/formatProducts';
```
rjmcneill: Just a small thing, but would the above 2 lines make more sense as:
```Javascript
import getCoinbaseProducts from './data/coinbase/getProducts';
import formatCoinbaseProducts from './data/coinbase/formatProducts';
```

Also, small typo in line 15 there.
coninbase > coinbase

src/assets/js/main.js
```diff
@@ -1,5 +1,5 @@
-import app from './app';
-import initialiseModel from './initialiseModel';
+import showcase from './sc';
 
-var model = initialiseModel();
-app(model).run();
+showcase.app()
```
tlsim: Would it be worth moving this to index.html, using the rollup-ed `sc` module there? We could then remove the extra rollup step, and it might help confirm that the module works as expected for openFin.

tlsim:
:+1:

src/assets/js/app.js
jleft: :+1: Good spot on the typo

We don't have a convention for naming our imports - something to consider in the future - but as the suggestion is easier to read (IMO), I'm happy to make the change :)

src/assets/js/app.js
```diff
@@ -388,13 +405,80 @@ export default function(initialModel) {
...
+    function addCoinbaseProducts(error, bitcoinProducts) {
+        if (error) {
+            var statusText = error.statusText || 'Unknown reason.';
+            var message = 'Error retrieving Coinbase products: ' + statusText;
+            model.notificationMessages.messages.unshift(messageModel(message));
+        } else {
+            var defaultPeriods = [model.periods.hour1, model.periods.day1];
+            var productPeriodOverrides = d3.map();
+            productPeriodOverrides.set('BTC-USD', [model.periods.minute1, model.periods.minute5, model.periods.hour1, model.periods.day1]);
+            var formattedProducts = coninbaseFormatProducts(bitcoinProducts, model.sources.bitcoin, defaultPeriods, productPeriodOverrides);
```
rjmcneill: Same typo here

rjmcneill:
LGTM @jleft :+1:

src/assets/js/main.js
jleft: We could do that, but as we're not bundling the other dependencies (d3, d3fc etc) with the `sc` module, I think it would mean that'd we'd lose the `rollup-plugin-npm` magic and have to concat our dependencies for the production build of our app?

I guess there are trade-offs either way... I don't have a strong preference. I'm happy to make a change now, but I'd suggest that we create another issues and look into the build separately - if that sounds reasonable?

jleft: >  I think it would mean that'd we'd lose the rollup-plugin-npm magic and have to concat our dependencies for the production build of our app?

Actually, I think I'm wrong here... I think I'm getting my self confused with something else!

tlsim: > I guess there are trade-offs either way... I don't have a strong preference. I'm happy to make a change now, but I'd suggest that we create another issues and look into the build separately - if that sounds reasonable?

Yeah that sounds reasonable :smile:

jleft: :smile:

jleft:
Thanks for the reviews guys :)

PR #540, opened 2016-01-13
Chart text elements no longer have text selection cursor

### LINKED ISSUE #538: Chart text element (axis labels) have text selection cursor, but aren't selectable
As the text is no longer selectable (#509), a normal/default cursor should be shown.

rjmcneill:
Fixes #538


jleft:
:+1:

PR #543, opened 2016-01-13
Zoom behaviour removed for tracking latest data

### LINKED ISSUE #539: Question regarding the navigator brush extent and scroll zooming
If you use the scroll wheel to zoom in on a selected element and your navigator brush extent is not extended to its full width, then you zoom in on the selected element.

If you use the scroll wheel to zoom in on a selected element and your navigator brush extent IS extended to its full width, then you dont zoom in on the selected element, instead its the furthest right.

This behavior is not consistent and is somewhat confusing to a user. Is it intentional?
For example:
[bandicam 2016-01-13 10-46-12-485.zip](https://github.com/ScottLogic/d3fc-showcase/files/88723/bandicam.2016-01-13.10-46-12-485.zip)

rjmcneill:
Quick fix to resolve #539 if we want to remove that functionality


jleft:
Thanks @rjmcneill, however the functionality is intentional and working as expected.

PR #547, opened 2016-01-15
Navbar mask only added when nav is visible

### LINKED ISSUE #544: *error when the navbar is not displayed
The latest dev build throws a large error (into the console) when you run it on a screen which is small enough that the navbar is removed.

[Error.zip](https://github.com/ScottLogic/d3fc-showcase/files/90261/Error.zip)

rjmcneill:
Relates to #544

### DIFF
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -3,7 +3,6 @@ import fc from 'd3fc';
 import util from '../util/util';
 import event from '../event';
 import zoomBehavior from '../behavior/zoom';
-import width from '../util/width';
 
 export default function() {
     var navHeight = 100; // Also maintain in variables.less
@@ -180,7 +179,7 @@ export default function() {
     d3.rebind(nav, dispatch, 'on');
 
     nav.dimensionChanged = function(container) {
-        layoutWidth = width(container.node());
+        layoutWidth = util.width(container.node());
         viewScale.range([0, layoutWidth]);
         maskXScale.range([0, layoutWidth]);
         maskYScale.range([navChartHeight, 0]);
--- a/src/assets/js/chart/nav.js
+++ b/src/assets/js/chart/nav.js
@@ -3,6 +3,7 @@ import fc from 'd3fc';
 import util from '../util/util';
 import event from '../event';
 import zoomBehavior from '../behavior/zoom';
+import width from '../util/width';
 
 export default function() {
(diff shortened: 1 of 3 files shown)


djmiley:
LGTM

jdhodges431:
Should this be done via the model instead?

jleft:
@rjmcneill - did you find the root cause of this error? Just wondering which code and conditions are causing this.

rjmcneill:
@jleft The problem is that the ```layoutWidth``` is not a number when it attempts to set the width of the ```mask-background.```
It only happens in mobile.

I can look at a different solution if you like?

jleft:
> @jleft The problem is that the layoutWidth is not a number when it attempts to set the width of the mask-background.
It only happens in mobile.

If it's not a number, what is it? `auto` I guess?

> I can look at a different solution if you like?

Just wondering if we can fix the root cause of this, as I could potentially see this happening in other places too.

jleft:
> If it's not a number, what is it? `auto` I guess?

Just to clarify, I mean the actual width style on the element (not `layoutWidth`, which I assume is `NaN`).

jleft:
FWIW, if this is to do with dimension calculations going wrong, then I'd be tempted to create a util which wraps the relevant [jQuery width function](http://api.jquery.com/width/).

rjmcneill:
It looks like it has nothing set for the width

width

src/assets/js/chart/nav.js
```diff
@@ -3,6 +3,7 @@ import fc from 'd3fc';
 import util from '../util/util';
 import event from '../event';
 import zoomBehavior from '../behavior/zoom';
+import width from '../util/width';
```
jleft: `util` is already imported, so if `src/assets/js/util/util.js` is updated to include `width`:
```js
// ...
import width from './width';

export default {
    domain: domain,
    layout: layout,
    uid: uid,
    width: width
};
```


 then we can remove this line and refer to `util.width` below

(Throttle and debounce probably should have been added too, but was missed in review!)

src/assets/js/util/width.js
```diff
@@ -0,0 +1,6 @@
+import $ from 'jquery';
+
+// Returns the width of a passed element
```
jleft: I'd remove the comment - the code is self-documenting

jdhodges431:
:+1:

jleft:
:+1:

PR #551, opened 2016-01-17
Update loading state to show spinner and translucent background over chart

### LINKED ISSUE #502: The loading experience is quite 'jarring'
When you switch instruments or time period all of the charts disappear, being replaced with a 'loading ...' text. This is visually quite jarring.

Could it perhaps render the charts (potentially without data) but at a reduced opacity?

Might be worth asking Rui or one of our other UXers for ideas.

jleft:
Fixes #502

recording

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -18,6 +18,7 @@ export default function() {
 
     var appTemplate = '<div class="container-fluid"> \
         <div id="notifications"></div> \
+        <div id="loading-status-message"></div> \
         <div class="row head-menu head-row"> \
             <div class="col-md-12 head-sub-row"> \
                 <div id="product-dropdown" class="dropdown product-dropdown"></div> \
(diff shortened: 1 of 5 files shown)


src/assets/js/app.js
```diff
@@ -232,12 +232,12 @@ export default function() {
     }
 
     function loading(isLoading, error) {
+        var spinner = '<div class="spinner"></div>';
```
djmiley: Do we need both declarations of spinner html?

src/assets/js/app.js
```diff
@@ -18,6 +18,9 @@ export default function() {
 
     var appTemplate = '<div class="container-fluid"> \
         <div id="notifications"></div> \
+        <div id="loading-status-message"> \
+            <div class="spinner"></div> \
```
djmiley: Do we need both declarations of spinner html?

djmiley: As in not totally sure line 22 is needed.

djmiley:
Looks good. :)

jdhodges431:
:+1:

src/assets/js/app.js
jleft: Removed :)

djmiley:
:+1:

PR #553, opened 2016-01-18
Added render call into coinbase products callback

### LINKED ISSUE #552: Coinbase products not added to dropdown until another render is called

rjmcneill:
Fixes #552

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -417,6 +417,8 @@ export default function() {
             var formattedProducts = formatCoinbaseProducts(bitcoinProducts, model.sources.bitcoin, defaultPeriods, productPeriodOverrides);
             model.headMenu.products = model.headMenu.products.concat(formattedProducts);
         }
+
+        render();
     }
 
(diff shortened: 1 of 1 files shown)


src/assets/js/app.js
```diff
@@ -416,6 +416,7 @@ export default function() {
             productPeriodOverrides.set('BTC-USD', [model.periods.minute1, model.periods.minute5, model.periods.hour1, model.periods.day1]);
             var formattedProducts = formatCoinbaseProducts(bitcoinProducts, model.sources.bitcoin, defaultPeriods, productPeriodOverrides);
             model.headMenu.products = model.headMenu.products.concat(formattedProducts);
+            render();
```
jleft: We still need to render if there is an error (so the notification message is shown). This could be moved outside and after the if/else, as the last line in the function.

rjmcneill: Ah yea, of course!
Update incoming

jleft:
:+1:

PR #555, opened 2016-01-18
Added error handling for CloudFlare specific errors

### LINKED ISSUE #549: Add more informative error messages for Coinbase historic API
Look into unofficial HTTP codes and their associated messages (Coinbase has timed-out a few times and appears to use Cloudflare). Consider using a common error message formatter, as this may apply to other APIs too.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

rjmcneill:
Relates to #549 

Added on the historic data and the retrieval of the Coinbase products.

List of CloudFlare error codes:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#CloudFlare

CloudFlare API responses:
https://api.cloudflare.com/#responses

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -320,7 +320,8 @@ export default function() {
                 } catch (e) {
                     responseText = '';
                 }
-                var statusText = err.statusText || 'Unknown reason.';
+
+                var statusText = handleHTTPErrors(err);
                 var message = 'Error getting historic data: ' + statusText + responseText;
 
                 addNotification(message);
@@ -423,7 +424,7 @@ export default function() {
 
     function addCoinbaseProducts(error, bitcoinProducts) {
         if (error) {
-            var statusText = error.statusText || 'Unknown reason.';
+            var statusText = handleHTTPErrors(error);
             var message = 'Error retrieving Coinbase products: ' + statusText;
             model.notificationMessages.messages.unshift(messageModel(message));
         } else {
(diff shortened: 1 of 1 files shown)


src/assets/js/app.js
```diff
@@ -421,6 +422,22 @@ export default function() {
         render();
     }
 
+    function handleHTTPErrors(error) {
+        var statusText;
+
+        // Range of CloudFlare specific error codes
+        if (error.status >= 520 && error.status <= 526) {
+            statusText = 'CloudFlare error -';
+            for (var i = 0; i < error.response.errors.length; i++) {
```
jdhodges431: How about:

```
error.response.errors.forEach(function(_error) {
    statusText += ' ' + _error.message;
});
```

jdhodges431:
:+1:

src/assets/js/app.js
rjmcneill: Agreed!
Keep forgetting about forEach!

src/assets/js/app.js
```diff
@@ -419,6 +420,22 @@ export default function() {
         render();
     }
 
+    function handleHTTPErrors(error) {
```
djmiley: More of a CloudFlare specific naming maybe, if this is all the method is handling?

djmiley: Or, if not, make the method more general? I may have missed something here though :(

rjmcneill: @djmiley it actually handles both.
Previously error handling was just 
```JavaScript
statusText = error.statusText || 'Unknown reason.';
```
This was fine for the normal error codes, but the CloudFlare specific errors are not returned in the same way.
So to handle this problem, I added this function which checks for the specific CloudFlare error codes (520 - 526) and handles them appropriately.
If an error code falls outside this range, the error is just handled the same way as the previous implemented

djmiley: Right, OK, sorry my brain didn't quite take things in before.

PR #556, opened 2016-01-19
Errors timeout after 5 seconds.

### LINKED ISSUE #542: The message about API key does not fade
When using the d3fc-showcase, you may get the message about running out of requests. This message will never fade. If you try calling the API again (multiple times) you can get large number of messages that you need to manually close.

jdhodges431:
Closes #542.


jdhodges431:
We may want to think about an exit animation at some point in the future.

tlsim:
:+1:

src/assets/js/app.js
```diff
@@ -167,8 +167,11 @@ export default function() {
         });
     }
 
-    function addNotification(message) {
-        model.notificationMessages.messages.unshift(messageModel(message));
+    function addNotification(messageBody) {
+        var message = messageModel(messageBody);
+        model.notificationMessages.messages.unshift(message);
+
+        setTimeout(onNotificationClose, message.timeout, message.id);
```
jleft: I'd be tempted to move this logic into toast.js - adding add a getter/setter for the visibility duration to the component and remove the `timeout` property from the model.

It would be nice if `null` could be set as the visibility duration to disable this functionality, so the toast message could remain open indefinitely.

Issue #508, opened 2016-01-04
Should we allow axis labels to overflow?


ColinEberhardt:
![capture](https://cloud.githubusercontent.com/assets/1098110/12095082/5a3b6886-b304-11e5-9c4a-b30574529276.PNG)

In most cases permitting an overflow would look better.

tlsim:
#325

jleft:
Reopening given #577 reverted the changes in #558

PR #558, opened 2016-01-20
Axis labels allowed to overflow.

jdhodges431:
#508

### DIFF
--- a/src/assets/styles/base.less
+++ b/src/assets/styles/base.less
@@ -43,3 +43,7 @@ text {
     fill: @background-colour;
   }
 }
+
+svg:not(:root) {
+  overflow: visible;
+}
--- a/src/assets/styles/charts.less
+++ b/src/assets/styles/charts.less
@@ -7,3 +7,7 @@
   user-select: none;
   cursor: default;
 }
+
+.plot-area-container > .plot-area {
+  overflow: hidden;
+}


rjmcneill:
LGTM :+1: 

overflow

rjmcneill:
LGTM :+1: 
overflow

src/assets/styles/charts.less
```diff
@@ -7,3 +7,7 @@
   user-select: none;
   cursor: default;
 }
+
+.plot-area {
+  overflow: hidden !important;
```
jleft: > Using `!important` is **bad practice** and should be avoided because it makes debugging more difficult by breaking the natural cascading in your stylesheets. 

https://developer.mozilla.org/en/docs/Web/CSS/Specificity

Is it possible to be more specific? Something like: `.plot-area-container > .plot-area` or `svg.plot-area`?

src/assets/styles/base.less
```diff
@@ -43,3 +43,7 @@ text {
     fill: @background-colour;
   }
 }
+
+svg:not(:root) {
+  overflow: visible;
+}
```
jleft: Add new line to EOF please.

src/assets/styles/charts.less
```diff
@@ -7,3 +7,7 @@
...
+}
```
jleft: Add new line to EOF please.

jleft:
Looks good :+1: Thanks for the updates.

There seems to be 3 commits with identical messages - please could these be squashed into 1 commit?

PR #561, opened 2016-01-20
Update jQuery to be a dependency (rather than dev dep); update shrinkwrap; configure globals for rollup

jleft:
- As #547 makes use of jQuery, it's now included as a dependency, rather than dev dependency
- Excluded jQuery from the rollup production build, as it's also needed for Bootstrap's JS
- Bumped a few other dep and devDep version - updated shrinkwrap to reflect this

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -254,7 +254,12 @@ module.exports = function(grunt) {
         rollup: {
             options: {
                 format: 'umd',
-                moduleName: 'sc'
+                moduleName: 'sc',
+                globals: {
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -254,7 +254,12 @@ module.exports = function(grunt) {
         rollup: {
```
jleft: @tlsim - would really appreciate it if you could have a look over this, especially the rollup changes :)

tlsim:
LGTM :+1:

Issue #563, opened 2016-01-22
X axis labels are being "clipped"


jackarnstein:
![clipped](https://cloud.githubusercontent.com/assets/5921987/12509170/864c267a-c0f8-11e5-9483-0fe71363f74f.png)

letters such as g and p which extend lower on the line are being clipped. Just a little more space between the axis labels and the bar would fix this.

PR #564, opened 2016-01-22
Changed x axis and nav height so labels aren't clipped.

jdhodges431:
#563.

### DIFF
--- a/src/assets/styles/variables.less
+++ b/src/assets/styles/variables.less
@@ -85,7 +85,7 @@
 @y-axis-width: 60px;
 
 // Shared X Axis
-@x-axis-height: 23px;
+@x-axis-height: 27px;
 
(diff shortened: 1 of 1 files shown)


src/assets/js/chart/nav.js
```diff
@@ -5,7 +5,7 @@ import event from '../event';
 import zoomBehavior from '../behavior/zoom';
 
 export default function() {
-    var navHeight = 100; // Also maintain in variables.less
+    var navHeight = 96; // Also maintain in variables.less
```
jleft: Do we need to change this? Can't we just adjust the x-axis height and let the primary chart automatically adjust itself?

jleft:
Looks good, cheers @jdhodges431 :+1:

Issue #571, opened 2016-01-25
Navigation bar chart does not stay within its lines


jackarnstein:
When the navigation chart size gets quite large, it spills out over its top edge into its border box. It makes it look a bit messy.

It also happens with some of the main charts where they make contact with the Y axis line

![navibar not clipping](https://cloud.githubusercontent.com/assets/5921987/12549963/a89cdf5a-c358-11e5-8146-f0b30b1c8aa5.png)

![navibar not clipping2](https://cloud.githubusercontent.com/assets/5921987/12550069/483abef6-c359-11e5-8de6-37fbc4250033.png)

![navibar not clipping3](https://cloud.githubusercontent.com/assets/5921987/12550155/c94a4ba6-c359-11e5-9a3a-5babaf371c95.png)

jackarnstein:
![naviborder](https://cloud.githubusercontent.com/assets/5921987/12920823/2012dcf2-cf42-11e5-9913-739463c765b0.png)

Problem still occurs on the side of your navi bar

PR #572, opened 2016-01-25
Padded nav y extent.

jdhodges:
Relates to #571.


djmiley:
Possibly go for `padding = [0, 0.04]`, to be consistent with the primary chart, and you don't need to declare it as a variable? (0.08 / 2 = 0.04, and we don't want to pad the bottom value?)

PR #583, opened 2016-01-29
Add coverage reporter for Karma

### LINKED ISSUE #459: Add coverage reporter for Karma
Add an additional Karma task which generates code coverage metrics (both HTML and text/text-summary would be useful):

https://github.com/karma-runner/karma-coverage

tlsim:
Resolves #459


jdhodges431:
What happened to shrinkwrap?

tlsim:
@jdhodges431 Not sure. I could try to make one with less churn.

jleft:
Will probably need to rebase with latest develop - `version` in the shrinkwrap has dropped from `1.0.0` to `0.11.0`.

I usually run this when generating the shrinkwrap:

```bash
rm -rf node_modules/ && npm cache clean && npm i && npm shrinkwrap --dev
```

Not sure if it will reduce the churn; no problem if not - as shrinkwrap is non-deterministic it's hard to avoid!

tlsim:
Rebased. I updated npm to 2.14.17, and running the above command still rewrites the "from" property of every dep in the shrinkwrap. Any suggestions?

jleft:
> Rebased. I updated npm to 2.14.17, and running the above command still rewrites the "from" property of every dep in the shrinkwrap. Any suggestions?

I can't think of anything. Same has happened to me previously too.

As long as it works, then this looks good to me :+1: thanks @tlsim

Issue #579, opened 2016-01-27
Give visual indication of which indicators have been added


owennw:
If there are already indicators added to the chart it would be nice to see in the indicators drop down which have already been selected. Possibly some highlighting or a checked icon of some sort. Same goes for the selected type (candlestick, OHLC etc.)

jleft:
:+1: Definitely agree with this - this has be bugging quite a few people for a while.

I'll organise some UX involvement, then look at getting this implemented.

jackarnstein:
Likewise, I totally agree with this one :) I mentioned it to james and we made a note of it so we could discuss it with the UX lot

jleft:
Desired behaviour:

![selected-indicator-styles](https://cloud.githubusercontent.com/assets/2017615/13216643/f1d0cf3a-d954-11e5-8814-2df093365a73.PNG)

jleft:
Duplicated by #718

PR #584, opened 2016-01-29
Indicators removed from list when selected. Button disappears when al…

jdhodges:
…l selected. Relates to #579.

### DIFF
--- a/src/assets/js/menu/generator/dropdown.js
+++ b/src/assets/js/menu/generator/dropdown.js
@@ -33,6 +33,13 @@ export default function() {
         var selectedIndex = model.selectedIndex || 0;
         var config = model.config;
 
+        if (model.options.length) {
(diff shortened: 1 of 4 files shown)


jleft:
Thanks @jdhodges, this seems like a reasonable approach, however - as mentioned in the comments for the related issue - we should have some input from a UX perspective before implementing anything for this issue. I'll mark this as blocked until we have the UX feedback. (The issue was in the _New Issues_ pipeline (not _To Do_), so is not ready to be worked on yet.)

jleft:
Uploaded design, specifying desired behaviour in https://github.com/ScottLogic/BitFlux/issues/579#issuecomment-187127029.

Given the difference in the design and the implemented behaviour here, I'm closing this PR. Please feel free to reopen if the implementation is updated to match the designs :smile:

Issue #589, opened 2016-01-29
1.0.0 broken in IE


tlsim:
IE11. Chart is not visible.

```
SCRIPT5022: Error parsing XML string
File: app.min.js, Line: 462, Column: 218
SCRIPT5007: Unable to get property 'getAttribute' of undefined or null reference
File: app.min.js, Line: 304, Column: 49
```

tlsim:
Non-minified build works locally.

jackarnstein:
![clipping](https://cloud.githubusercontent.com/assets/5921987/12715525/b3aa4cb4-c8d2-11e5-8a9f-f3f641f30b54.png)

I can post this as a seperate issue if you like, but the clipping is way off on the x axis for the navigation bar

jleft:
> I can post this as a seperate issue if you like, but the clipping is way off on the x axis for the navigation bar

Yes please, a separate issue would be great, thank you.

PR #590, opened 2016-01-29
Fix svg-innerhtml in IE

jleft:
#589


jleft:
Used approach in #591

Issue #503, opened 2016-01-04
navigator labels colliding


ColinEberhardt:
![capture](https://cloud.githubusercontent.com/assets/1098110/12094890/373034b2-b303-11e5-8d80-ac7453e148f7.PNG)

Instrument BTC-CAD Daily time period

jleft:
One solution is to adjust the number of ticks to the available space. E.g. https://github.com/ScottLogic/d3fc/commit/79b8ac7ee79d135954947bf54958d4dd2101b0f5?diff=unified#diff-1cdfafdab5c0a9c35924fd66089f0337R143

ColinEberhardt:
That sounds like a good solution to me @jleft

PR #599, opened 2016-02-02
Manual xAxis labels implemented

rjmcneill:
Relates to #503

### DIFF
--- a/src/assets/js/chart/xAxis.js
+++ b/src/assets/js/chart/xAxis.js
@@ -1,8 +1,6 @@
 import d3 from 'd3';
 import fc from 'd3fc';
 import util from '../util/util';
-import zoom from '../behavior/zoom';
-import domain from '../util/domain/domain';
 
 export default function() {
     var xScale = fc.scale.dateTime();
@@ -16,9 +14,26 @@ export default function() {
     var tickValues = [];
     var originalDomain;
     var tickSpacing;
+    var tickPixelSpacing = 150;
 
     function calculateNumberOfTicks() {
-        return Math.max(Math.ceil(containerWidth / 150), 2);
+        return Math.max(Math.ceil(containerWidth / tickPixelSpacing), 2);
+    }
+
+    function removeTickValuesOutsideDomain(viewDomain) {
+        // Remove ticks lower than low view domain
+        tickValues.forEach(function(tick) {
+            if (tick < viewDomain[0]) {
+                tickValues.splice(0, 1);
+            }
+        });
+
+        // Remove ticks higher than high view domain
+        for (var i = 0; i < tickValues.length; i++) {
+            if (tickValues[i] > viewDomain[1]) {
+                tickValues.splice(i, tickValues.length - i);
+            }
+        }
     }
 
     function calculateTickValues(model, numberOfTicks) {
@@ -26,27 +41,22 @@ export default function() {
         var upperRoundedViewDomain = model.period.d3TimeInterval.unit.floor(model.viewDomain[1]).getTime();
         var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain);
         var zoomLevelChanged;
-
-        if (originalDomain) {
-            zoomLevelChanged = (originalDomain[1].getTime() - originalDomain[0].getTime() !== model.viewDomain[1].getTime() - model.viewDomain[0].getTime());
-            if (zoomLevelChanged || containerWidthChanged) {
-                // If they have zoomed, recalculate the tick spacing
-                tickSpacing = util.domain.roundTimestampUp((viewDomainDifference / numberOfTicks), model.period.seconds * 1000);
-                containerWidthChanged = false;
-            }
-        }
-
-        if (!tickValues.length) {
-            tickSpacing = util.domain.roundTimestampUp((viewDomainDifference / numberOfTicks), model.period.seconds * 1000);
-        }
-
-        var lowerRounded = util.domain.roundTimestampDown(lowerRoundedViewDomain, tickSpacing);
         var viewDomainChanged = false;
 
         if (originalDomain) {
+            zoomLevelChanged = (originalDomain[1].getTime() - originalDomain[0].getTime() !==
+                model.viewDomain[1].getTime() - model.viewDomain[0].getTime());
+
             viewDomainChanged = (originalDomain[0].getTime() !== model.viewDomain[0].getTime()) ||
-            (originalDomain[1].getTime() !== model.viewDomain[1].getTime());
+                (originalDomain[1].getTime() !== model.viewDomain[1].getTime());
         }
+
+        if (!tickValues.length || zoomLevelChanged || containerWidthChanged) {
+            tickSpacing = util.roundUpToNearestMultiple((viewDomainDifference / numberOfTicks), model.period.seconds * 1000);
+            containerWidthChanged = false;
+        }
+
+        var lowerRounded = util.roundDownToNearestMultiple(lowerRoundedViewDomain, tickSpacing);
 
         if (tickValues.length !== numberOfTicks || viewDomainChanged || zoomLevelChanged) {
             tickValues = [];
@@ -56,21 +66,7 @@ export default function() {
             }
         }
 
-        // Remove ticks lower than low view domain
-        tickValues.forEach(function(tick) {
-            if (tick < model.viewDomain[0]) {
-                tickValues.splice(0, 1);
-            }
-        });
-
-        // Remove ticks higher than high view domain
-        for (var j = 0; j < tickValues.length; j++) {
-            if (tickValues[j] > model.viewDomain[1]) {
-                tickValues.splice(j, tickValues.length - j);
-            }
-        }
-
-        xAxis.tickValues(tickValues);
+        removeTickValuesOutsideDomain(model.viewDomain);
 
         if (viewDomainChanged || !originalDomain) {
             originalDomain = model.viewDomain;
@@ -83,60 +79,16 @@ export default function() {
         var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain) / 1000;
 
         var minimumPeriod = model.period.seconds;
-        var formatting;
+        var timePeriods = [3600, 86400, 2592000];
+        var tickFormats = [d3.time.format('%_I:%M %p'), d3.time.format('%a %_I %p'), d3.time.format('%b %e'), d3.time.format('%B %Y')];
 
-        var yearInSeconds = 31557600;
-        var monthInSeconds = 2592000;
-        var dayInSeconds = 86400;
-        var hourInSeconds = 3600;
-
-        var withinYear = viewDomainDifference < yearInSeconds;
-        var withinMonth = viewDomainDifference < monthInSeconds;
-        var withinDay = viewDomainDifference < dayInSeconds;
-        var withinHour = viewDomainDifference < hourInSeconds;
-
-        var monthsDifference = Math.floor(viewDomainDifference / monthInSeconds);
-        var daysDifference = Math.floor(viewDomainDifference / dayInSeconds);
-        var hoursDifference = Math.floor(viewDomainDifference / hourInSeconds);
-
-        if (withinHour) {
-            formatting = 'minutes';
-        } else if (withinDay) {
-            if (minimumPeriod < hourInSeconds && hoursDifference < numberOfTicks) {
-                formatting = 'minutes';
-            } else {
-                formatting = 'hours';
+        for (var i = 0; i < timePeriods.length; i++) {
+            if (viewDomainDifference < timePeriods[i] ||
+                (minimumPeriod < timePeriods[i] && Math.floor(viewDomainDifference / timePeriods[i]) < numberOfTicks)) {
+                return tickFormats[i];
             }
-        } else if (withinMonth) {
-            if (minimumPeriod < dayInSeconds && daysDifference < numberOfTicks) {
-                formatting = 'hours';
-            } else {
-                formatting = 'days';
-            }
-        } else if (withinYear) {
-            if (monthsDifference >= numberOfTicks) {
-                formatting = 'months';
-            } else {
-                formatting = 'days';
-            }
-        } else {
-            xAxis.tickFormat(d3.time.format('%b %Y'));
         }
-
-        switch (formatting) {
-        case 'minutes':
-            xAxis.tickFormat(d3.time.format('%_I:%M %p'));
-            break;
-        case 'hours':
-            xAxis.tickFormat(d3.time.format('%a %_I %p'));
-            break;
-        case 'days':
-            xAxis.tickFormat(d3.time.format('%b %e'));
-            break;
-        case 'months':
-            xAxis.tickFormat(d3.time.format('%B'));
-            break;
-        }
+        return tickFormats[tickFormats.length - 1];
     }
 
     function xAxisChart(selection) {
@@ -147,7 +99,8 @@ export default function() {
         var numberOfTicks = calculateNumberOfTicks(model);
         xScale.domain(model.viewDomain);
         calculateTickValues(model, numberOfTicks);
-        adaptAxisFormat(model, numberOfTicks);
+        xAxis.tickValues(tickValues);
+        xAxis.tickFormat(adaptAxisFormat(model, numberOfTicks));
         selection.call(xAxis);
     }
 
--- a/src/assets/js/chart/xAxis.js
+++ b/src/assets/js/chart/xAxis.js
@@ -22,8 +22,8 @@ export default function() {
     }
 
     function calculateTickValues(model, numberOfTicks) {
-        var lowerRoundedViewDomain = domain.roundTimestampUp(model.viewDomain[0].getTime(), model.period.seconds * 1000);
-        var upperRoundedViewDomain = domain.roundTimestampDown(model.viewDomain[1].getTime(), model.period.seconds * 1000);
+        var lowerRoundedViewDomain = model.period.d3TimeInterval.unit.ceil(model.viewDomain[0]).getTime();
+        var upperRoundedViewDomain = model.period.d3TimeInterval.unit.floor(model.viewDomain[1]).getTime();
         var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain);
         var zoomLevelChanged;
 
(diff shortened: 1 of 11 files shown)


rjmcneill:
Need to investigate some of the API docs to simplify the rounding of the intervals.

https://github.com/mbostock/d3/wiki/Time-Intervals

rjmcneill:
There isn't any overflow stuff included in this, but I have that in progress in the yAxis work.  Might be worth pulling that functionality out into a separate component once both issue are resolved

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+        var withinYear = viewDomainDifference < yearInSeconds;
+        var withinMonth = viewDomainDifference < monthInSeconds;
+        var withinDay = viewDomainDifference < dayInSeconds;
+        var withinHour = viewDomainDifference < hourInSeconds;
+
+        var monthsDifference = Math.floor(viewDomainDifference / monthInSeconds);
+        var daysDifference = Math.floor(viewDomainDifference / dayInSeconds);
+        var hoursDifference = Math.floor(viewDomainDifference / hourInSeconds);
+
+        if (withinHour) {
```
djmiley: You could do a recursive check here possibly, rather than chaining if statements? Using [365, 12, 31, 24] etc.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+        var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain) / 1000;
+
+        var minimumPeriod = model.period.seconds;
+        var formatting;
+
+        var yearInSeconds = 31557600;
+        var monthInSeconds = 2592000;
+        var dayInSeconds = 86400;
+        var hourInSeconds = 3600;
+
```
djmiley: What about leap years and months with different number of days in?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+        var withinYear = viewDomainDifference < yearInSeconds;
+        var withinMonth = viewDomainDifference < monthInSeconds;
+        var withinDay = viewDomainDifference < dayInSeconds;
+        var withinHour = viewDomainDifference < hourInSeconds;
```
djmiley: Do you need to declare these?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
         xScale.domain(model.viewDomain);
-        preventTicksMoreFrequentThanPeriod(model.period);
+        calculateTickValues(model, numberOfTicks);
+        adaptAxisFormat(model, numberOfTicks);
         selection.call(xAxis);
     }
 
     xAxisChart.dimensionChanged = function(container) {
-        xScale.range([0, parseInt(container.style('width'), 10)]);
+        containerWidth = parseInt(container.style('width'), 10);
```
djmiley: Is there a way around increasing the scope here?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+        return Math.max(Math.ceil(containerWidth / 150), 2);
+    }
+
+    function calculateTickValues(model, numberOfTicks) {
+        var lowerRoundedViewDomain = model.period.d3TimeInterval.unit.ceil(model.viewDomain[0]).getTime();
+        var upperRoundedViewDomain = model.period.d3TimeInterval.unit.floor(model.viewDomain[1]).getTime();
+        var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain);
+        var zoomLevelChanged;
+
+        if (originalDomain) {
```
djmiley: Is there a way to call `if(originalDomain)` only once, the repetition looks a bit off.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+        if (tickValues.length !== numberOfTicks || viewDomainChanged || zoomLevelChanged) {
+            tickValues = [];
+            for (var i = 0; i <= numberOfTicks; i++) {
+                var tickDate = (model.period.d3TimeInterval.unit.ceil(lowerRounded + (tickSpacing * i)));
+                tickValues.push(tickDate);
+            }
+        }
+
+        // Remove ticks lower than low view domain
+        tickValues.forEach(function(tick) {
```
djmiley: Pull out into function

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+            if (tick < model.viewDomain[0]) {
+                tickValues.splice(0, 1);
+            }
+        });
+
+        // Remove ticks higher than high view domain
```
djmiley: Pull out into function

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,154 @@
...
+                formatting = 'months';
+            } else {
+                formatting = 'days';
+            }
         } else {
-            xAxis.ticks(6);
+            xAxis.tickFormat(d3.time.format('%b %Y'));
+        }
+
+        switch (formatting) {
```
djmiley: Set formatting to the tick format and then call `xAxis.tickFormat(formatting)` ? Although change name to `tickFormat` or similar

src/assets/js/chart/xAxis.js
rjmcneill: It's just a rough number to trigger the change of tick formatting.
Being exact with this isn't a big deal really, so I just picked standard values

src/assets/js/chart/xAxis.js
rjmcneill: Do you mean iterative checking @djmiley?
I'm not sure that would be any better in this scenario.
I agree it's messy problem though, I can't think of a tidy solution.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,139 @@
...
+                tickSpacing = domain.roundTimestampUp((viewDomainDifference / numberOfTicks), model.period.seconds * 1000);
+                containerWidthChanged = false;
+            }
+        }
+
+        if (!tickValues.length) {
+            tickSpacing = domain.roundTimestampUp((viewDomainDifference / numberOfTicks), model.period.seconds * 1000);
+        }
+
+        var lowerRounded = domain.roundTimestampDown(lowerRoundedViewDomain, tickSpacing);
```
djmiley: Move into if statement

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,139 @@
...
-        if (scaleTickSeconds < period.seconds) {
-            xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value);
+    var containerWidth;
+    var containerWidthChanged;
+    var tickValues = [];
+    var originalDomain;
+    var tickSpacing;
+
+    function calculateNumberOfTicks() {
+        return Math.max(Math.ceil(containerWidth / 150), 2);
```
djmiley: 150?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,139 @@
...
+        var viewDomainChanged = false;
+
+        if (originalDomain) {
+            zoomLevelChanged = (originalDomain[1].getTime() - originalDomain[0].getTime() !==
+                model.viewDomain[1].getTime() - model.viewDomain[0].getTime());
+
+            viewDomainChanged = (originalDomain[0].getTime() !== model.viewDomain[0].getTime()) ||
+            (originalDomain[1].getTime() !== model.viewDomain[1].getTime());
+
+            if (zoomLevelChanged || containerWidthChanged) {
```
djmiley: Is it possible to only set `tickSpacing` once? Combine this if statement with the one below?

src/assets/js/chart/xAxis.js
djmiley: I think I may mean iterative checking? It could possible remove the if-else within the if statement?

src/assets/js/chart/xAxis.js
djmiley: As in where does it come from?

src/assets/js/chart/xAxis.js
rjmcneill: This can't go into the if statement because the tickSpacing is calculated in different places depending on the situation.
So if I was to move this into the if statement about, it would also need to be in the if statement on line 53.
This way we just avoid duplicating the code.

src/assets/js/chart/xAxis.js
rjmcneill: Just means we roughly have a tick value every 150px.
It was just from observation.
Do you think I should make it a variable?

djmiley: If you could find out where this comes from, it would be good. Looks a bit too much of a magic number to me

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+        var yearInSeconds = 31557600;
+        var monthInSeconds = 2592000;
+        var dayInSeconds = 86400;
+        var hourInSeconds = 3600;
+
+        var monthsDifference = Math.floor(viewDomainDifference / monthInSeconds);
+        var daysDifference = Math.floor(viewDomainDifference / dayInSeconds);
+        var hoursDifference = Math.floor(viewDomainDifference / hourInSeconds);
+
+        if (viewDomainDifference < hourInSeconds) {
```
rjmcneill: Still need to investigate a cleaner way of doing this

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
 
-    function preventTicksMoreFrequentThanPeriod(period) {
-        var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
-        if (scaleTickSeconds < period.seconds) {
-            xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value);
+    var containerWidth;
+    var containerWidthChanged;
+    var tickValues = [];
+    var originalDomain;
+    var tickSpacing;
```
djmiley: Can reduce scope for `tickSpacing`

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+        .scale(xScale)
+        .orient('bottom');
 
-    function preventTicksMoreFrequentThanPeriod(period) {
-        var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
-        if (scaleTickSeconds < period.seconds) {
-            xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value);
+    var containerWidth;
+    var containerWidthChanged;
+    var tickValues = [];
```
djmiley: Can reduce scope for `tickValues`

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+
+        if (tickValues.length !== numberOfTicks || viewDomainChanged || zoomLevelChanged) {
+            tickValues = [];
+            for (var i = 0; i <= numberOfTicks; i++) {
+                var tickDate = (model.period.d3TimeInterval.unit.ceil(lowerRounded + (tickSpacing * i)));
+                tickValues.push(tickDate);
+            }
+        }
+
+        removeTickValuesOutsideDomain(model);
```
djmiley: Pass in `viewDomain` and `tickValues`, and return new `tickValues`

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
-            xAxis.ticks(6);
+            tickFormat = d3.time.format('%b %Y');
         }
+
+        xAxis.tickFormat(tickFormat);
     }
 
     function xAxisChart(selection) {
         var model = selection.datum();
+        var numberOfTicks = calculateNumberOfTicks(model);
```
djmiley: Don't need to pass in `model`.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
-      .scale(xScale)
-      .orient('bottom');
+        .scale(xScale)
+        .orient('bottom');
 
-    function preventTicksMoreFrequentThanPeriod(period) {
-        var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
-        if (scaleTickSeconds < period.seconds) {
-            xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value);
+    var containerWidth;
```
djmiley: Could we remove this variable all together, and just call `parseInt(container.style('width'), 10)` instead? In the possible hope a different way of calculating the number of ticks comes along.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+            }
+        } else if (viewDomainDifference < monthInSeconds) {
+            if (minimumPeriod < dayInSeconds && daysDifference < numberOfTicks) {
+                tickFormat = d3.time.format('%a %_I %p');
+            } else {
+                tickFormat = d3.time.format('%b %e');
+            }
+        } else if (viewDomainDifference < yearInSeconds) {
+            if (monthsDifference >= numberOfTicks) {
+                tickFormat = d3.time.format('%B');
```
djmiley: Why are these two different?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+            } else {
+                tickFormat = d3.time.format('%b %e');
+            }
         } else {
-            xAxis.ticks(6);
+            tickFormat = d3.time.format('%b %Y');
```
djmiley: Why are these two different?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+
+        var yearInSeconds = 31557600;
+        var monthInSeconds = 2592000;
+        var dayInSeconds = 86400;
+        var hourInSeconds = 3600;
+
+        var monthsDifference = Math.floor(viewDomainDifference / monthInSeconds);
+        var daysDifference = Math.floor(viewDomainDifference / dayInSeconds);
+        var hoursDifference = Math.floor(viewDomainDifference / hourInSeconds);
+
```
djmiley: Could be neater if you don't declare these three variables, and just calculate them in the if.

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,135 @@
...
+        var upperRoundedViewDomain = model.period.d3TimeInterval.unit.floor(model.viewDomain[1]).getTime();
+        var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain) / 1000;
+
+        var minimumPeriod = model.period.seconds;
+        var tickFormat;
+
+        var yearInSeconds = 31557600;
+        var monthInSeconds = 2592000;
+        var dayInSeconds = 86400;
+        var hourInSeconds = 3600;
```
djmiley: Try putting these in an array, and using a while loop, to thin down the code below. You would need to declare the tick formats in an array though.

src/assets/js/chart/xAxis.js
rjmcneill: We only have the container passed into this section of the code though :/

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,120 @@
...
+        var viewDomainDifference = (upperRoundedViewDomain - lowerRoundedViewDomain) / 1000;
+
+        var minimumPeriod = model.period.seconds;
+        var tickFormat;
+
+        var monthInSeconds = 2592000;
+        var dayInSeconds = 86400;
+        var hourInSeconds = 3600;
+
+        if ((viewDomainDifference < hourInSeconds) ||
```
rjmcneill: @djmiley not quite what we talked about, but it's cleaner now.
I couldn't think of anyother way to get  the implementation we talked about working

src/assets/js/chart/xAxis.js
djmiley: Ok, sorry, my mistake, no problem :)

src/assets/js/chart/xAxis.js
djmiley: `var timePeriods = [2592000, 86400, 3600];`
`var tickFormats = [d3.time.format('%B'), d3.time.format('%b %e'), d3.time.format('%a %_I %p'), d3.time.format('%_I:%M %p')];`

`for (var i = 0; i < timePeriods.length; i++) {`
&nbsp;&nbsp;&nbsp;&nbsp;`if (viewDomainDifference < timePeriods[i] || (( i !== timePeriods.length - 1 && minimumPeriods < timePeriods[i + 1]) && Math.floor(viewDomainDifference / timePeriods[i]) < numberOfTicks)) {`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`return ticksFormats[i];`
&nbsp;&nbsp;&nbsp;&nbsp;`}`
`}`
`return ticksFormats[tickFormats.length - 1];`

djmiley: I think this might work?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,120 @@
...
+            tickFormat = d3.time.format('%a %_I %p');
+        } else if ((viewDomainDifference < monthInSeconds) ||
+                (Math.floor(viewDomainDifference / monthInSeconds) < numberOfTicks)) {
+            tickFormat = d3.time.format('%b %e');
         } else {
-            xAxis.ticks(6);
+            tickFormat = d3.time.format('%B');
         }
+
+        xAxis.tickFormat(tickFormat);
```
djmiley: return tickFormat instead?

src/assets/js/chart/xAxis.js
```diff
@@ -1,30 +1,120 @@
...
     }
 
     function xAxisChart(selection) {
         var model = selection.datum();
+        var numberOfTicks = calculateNumberOfTicks();
         xScale.domain(model.viewDomain);
-        preventTicksMoreFrequentThanPeriod(model.period);
+        calculateTickValues(model, numberOfTicks);
```
djmiley: I like the idea of returning the tickValues aswell, below line 74. I don't think that if statement should affect the tickValues that have been calculated before that.

src/assets/js/chart/xAxis.js
rjmcneill: No joy with this @djmiley sadly

rjmcneill:
Noticed some issues with this while rebasing, adding WIP tag for now

src/assets/js/util/domain/roundTimestampDown.js
```diff
@@ -0,0 +1,3 @@
+export default function(timeStamp, requiredRounding) {
```
djmiley: This is only called once, and is not strictly related to timestamps. Is there not an inbuilt maths method? Possibly remove?

src/assets/js/util/domain/roundTimestampUp.js
```diff
@@ -0,0 +1,3 @@
+export default function(timeStamp, requiredRounding) {
```
djmiley: This is only called once, and is not strictly related to timestamps. Is there not an inbuilt maths method? Possibly remove?

rjmcneill: Could do something like:
```JavaScript
Math.ceil((viewDomainDifference / numberOfTicks)/(model.period.seconds * 1000)) * model.period.seconds * 1000
```
I personally think that that's messier though, and if we implement manual tick values elsewhere, this util will be used in a few places.
I agree on the naming though

djmiley: Ok, I agree. A renaming would be make a lot more sense :)

src/assets/js/util/roundNumberDown.js
```diff
@@ -0,0 +1,3 @@
+export default function(number, requiredRounding) {
```
djmiley: Maybe `roundDownToNearestMultiple` or something similar, `roundNumberDown` doesn't seem quite right

src/assets/js/util/roundNumberUp.js
```diff
@@ -0,0 +1,3 @@
+export default function(number, requiredRounding) {
```
djmiley: Maybe `roundUpToNearestMultiple` or something similar, `roundNumberUp` doesn't seem quite right

PR #600, opened 2016-02-02
Browserify

tlsim:
Feel free to ignore this, but it possibly simplifies managing dependencies. There is a trade off with slower initial dev build and production minify, but dev rebuilds should be fast using watchify. It also makes the bootstrap dependency a bit more explicit.

This is similar to what we were doing before in production with the rollup commonjs plugin, but that caused some issues (such as bringing in 2 d3 versions with only patch level changes).

Shrinkwrap is not behaving. I followed the process from #583, but this time it rewrote the "from" properties back to the older format.


Gruntfile.js
```diff
@@ -280,28 +210,28 @@ module.exports = function(grunt) {
                 options: {
                     sourceMap: true
                 }
-            },
+            }
+        },
+
+        browserify: {
```
tlsim: Could also have browserify build the module (and not rollup), using similar config with `external` for d3, d3fc and jquery.

Issue #587, opened 2016-01-29
gridlines dont match up to ticks?


jackarnstein:
![gridlines](https://cloud.githubusercontent.com/assets/5921987/12677801/667a0056-c693-11e5-9826-fa8635edd9d1.png)

Should the gridlines not match up to the ticks at the edge of the screen?

jackarnstein:
![gridlinebug](https://cloud.githubusercontent.com/assets/5921987/12919562/946d3724-cf3c-11e5-97e4-ca9099995a57.png)

jackarnstein:
Ive attached a video which will help you replicate the bug which is occuring with your grid lines.

[bandicam 2016-02-09 14-50-36-197.zip](https://files.zenhub.io/56b9fd2ec7438c6402f4984c)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

PR #608, opened 2016-02-04
Gridline count now adapts to tick count

rjmcneill:
Relates to #587

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -67,7 +67,6 @@ export default function() {
     crosshair.id = util.uid();
 
     var gridlines = fc.annotation.gridline()
-      .yTicks(5)
       .xTicks(0);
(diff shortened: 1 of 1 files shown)


src/assets/js/chart/primary.js
```diff
@@ -204,6 +203,8 @@ export default function() {
                 .attr('transform', 'translate(' + calloutHeight / 2 + ',1)');
           });
 
+        gridlines.yTicks(tickValues.length);
```
jleft: Should this be `tickValues.length - 1`, as `tickValues` will also include the tick used for the latest price tick/callout?

rjmcneill: Yup!
Doesn't actually show as a problem, so it must not go above the ```tickValue.length``` value.
Update coming

Issue #603, opened 2016-02-04
Relative strength index gridline widths are inconsistent


jackarnstein:
The top gridlines of the RSI chart contain 2x the number of light grey pixels.
![widthissue](https://cloud.githubusercontent.com/assets/5921987/12812929/d516a328-cb2c-11e5-84af-715f9ef8de7c.png)

jackarnstein:
Does not always seem to happen, but if you add a MACD chart underneath it, you can see that the middle line is different to the upper/lower gridlines.
![gridlinestwo](https://cloud.githubusercontent.com/assets/5921987/12813666/5e8fbfbe-cb31-11e5-9333-3ed1af8001d4.png)

jleft:
Looks like it's probably happening because it'll be dealing with fractions of pixels.

This will probably give use a quick win, without having to round everything:

```css
shape-rendering: crispEdges;
```

jackarnstein:
They are also being rendered on top of the actual chart lines. Image one shows this really clearly.

rjmcneill:
Opened a new issue for @jackarnstein's comment
#610

jleft:
> They are also being rendered on top of the actual chart lines. Image one shows this really clearly.

Good spot - this is coming directly from a d3fc component so I've raised an issue against d3fc too.

PR #611, opened 2016-02-04
Changed shape rendering on rsi annotation line to crisp edges

rjmcneill:
Relates to #603


jleft:
Looks good to me.

I think it would be worth adding `shape-rendering: crispEdges` to target more element too - pretty much anything with straight lines: gridlines, candlestick, OHLC etc. I'll open a separate issue.

Issue #597, opened 2016-02-01
Multiple volume secondary containers


jackarnstein:
To reproduce:
Add all secondary indicators, remove all except volume, and then all secondary containers will contain class of secondary volume.
![image](https://cloud.githubusercontent.com/assets/5921987/12722796/7c23a264-c8fe-11e5-893f-d3f61b691703.png)

jleft:
Resolved in #749

Issue #331, opened 2015-10-27
Create component to manage secondary charts


jleft:

jdhodges:
Is the idea here to have the secondaries grouped in a container?

jleft:
Yes. Rather than having [three containers permanently](https://github.com/ScottLogic/BitFlux/blob/develop/src/assets/js/app.js#L34-L36) in the page and having app do [lots](https://github.com/ScottLogic/BitFlux/blob/develop/src/assets/js/app.js#L114-L121) [of](https://github.com/ScottLogic/BitFlux/blob/develop/src/assets/js/app.js#L379-L382) [logic](https://github.com/ScottLogic/BitFlux/blob/develop/src/assets/js/app.js#L385-L386) for the secondaries, the idea is to have a component which can manage this. (Complexity in layout.js should also be reduced.)

I expect it would be similar to how we use multi/Cartesian chart with the series/indicators, but rather than the renderers being rendered into the same container, each renderer would have its own container. It might look something like this:

```js
var secondaries = multiChart() // <-- new component
    .charts([rsi, volume]);

secondariesContainer.call(secondaries);
```

Give me a shout if you'd like to chat through this :)

PR #624, opened 2016-02-10
Charts and secondaries components.

jdhodges:
Relates to #390 #331 #597 #680. Supersedes #601.


jleft:
Really nice work 👍  Thank @jdhodges! Just made a couple of small change in #749.

PR #631, opened 2016-02-15
Import correct Roboto font weight (300)

jleft:


jdhodges:
I knew there was something off about the font! :+1:

Issue #576, opened 2016-01-27
Add crosshair to secondary charts


jleft:
Similar to: https://d3fc.io/examples/low-barrel/index.html

PR #635, opened 2016-02-17
Added crosshairs to the secondary charts

rjmcneill:
Relates to #576


src/assets/js/chart/secondary/base.js
```diff
@@ -1,6 +1,7 @@
 import d3 from 'd3';
 import fc from 'd3fc';
 import event from '../../event';
+// import util from '../../util/util';
```
jdhodges: Remove comment

src/assets/js/chart/secondary/volume.js
```diff
@@ -5,25 +5,75 @@ import event from '../../event';
...
+                crosshairDataPoint = updatedCrosshairData[0].datum;
+                dispatch.crosshairChange(updatedCrosshairData[0].datum);
+            } else {
+                crosshairDataPoint = undefined;
+                dispatch.crosshairChange(undefined);
+            }
+        })
+        .on('trackingend', function() {
+            dispatch.crosshairChange(undefined);
+        });
```
jdhodges: Any way we can avoid repeating this crosshair creation code? Maybe a util function or separate component?

src/assets/js/model/chart/primary.js
```diff
@@ -5,6 +5,7 @@ import util from '../../util/util';
 export default function(initialProduct) {
     var model = {
         data: [],
+        currentCrosshairData: null,
```
jdhodges: I'd be inclined to just call this `crosshairData`

src/assets/js/model/chart/secondary.js
```diff
@@ -1,8 +1,27 @@
+import candlestickSeries from '../../series/candlestick';
+import option from '../menu/option';
+import util from '../../util/util';
+
 export default function(initialProduct) {
-    return {
+    var model = {
         data: [],
+        currentCrosshairData: null,
```
jdhodges: Same as previous comment.

src/assets/styles/primary.less
```diff
@@ -62,34 +62,5 @@
...
-        > line {
-          stroke: @crosshair-colour;
-          stroke-dasharray: 0;
-        }
-      }
-    }
-  }
-
   background-color: @background-colour;
 }
```
jdhodges: Need new line at the end of the file, must have been removed at some point.

src/assets/js/chart/secondary/macd.js
```diff
@@ -1,37 +1,104 @@
...
+              .style('stroke-width', width);
+        });
+    }
+
+    var initialised = false;
+    var macdSeries;
+
+    function initialiseSeries(primarySeries) {
+        // Initialise the series to allow the secondaries to snap correctly
+        macdSeries = jquery.extend({}, primarySeries);
```
rjmcneill: This is needed to ensure we can snap to data before the bar chart part of the macd indicator starts.

src/assets/js/app.js
```diff
@@ -213,6 +217,8 @@ export default function() {
 
     function onCrosshairChange(dataPoint) {
         model.legend.data = dataPoint;
+        model.primaryChart.crosshairData = dataPoint;
+        model.secondaryChart.crosshairData = dataPoint;
```
jleft: Are either of these used anywhere?

src/assets/js/util/crosshair.js
```diff
@@ -0,0 +1,28 @@
+import d3 from 'd3';
+import fc from 'd3fc';
+import event from '../event';
+import util from './util';
+
+export default function() {
```
jleft: I think it's a very good idea to separate this, but I don't think this is a util, I'd move this to a new `tool` namespace.

src/assets/js/model/chart/secondary.js
```diff
@@ -1,9 +1,28 @@
...
         data: [],
+        crosshairData: null,
         viewDomain: [],
         trackingLatest: true,
         product: initialProduct,
         discontinuityProvider: initialDiscontinuityProvider
     };
+
+    var candlestick = candlestickSeries();
+    candlestick.id = util.uid();
```
jleft: It feels pretty odd having the primary chart series on the secondary charts model, but I can see why - let's discuss if there's an alternative.

If this was to stay, then until an indicator is changed, both the primary chart model and secondary chart model will be storing a reference to a different candlestick series, with a different ID. These should both be pulled up a level and passed into the models when we initialise them. (I'd leave this for now though until we discuss the above.)

src/assets/js/model/chart/secondary.js
```diff
@@ -1,9 +1,28 @@
...
+    var _primarySeries = option('Candlestick', 'candlestick', candlestick);
+    _primarySeries.option.extentAccessor = ['high', 'low'];
+    Object.defineProperty(model, 'primarySeries', {
+        get: function() { return _primarySeries; },
+        set: function(newValue) {
+            _primarySeries = newValue;
+            model.selectorsChanged = true;
```
jleft: This is unused by the secondary charts and is not initialised anywhere, I don't think we need to use `Object.defineProperty`

src/assets/js/chart/secondary/macd.js
```diff
@@ -1,37 +1,93 @@
 import d3 from 'd3';
```
jleft: There's quite a lot of duplication between all of the secondary charts, I think some should be able to be pulled out into the base component - let's discuss further.

PR #636, opened 2016-02-18
Added active and hover styles for buttons.

jdhodges:
#524


rjmcneill:
Do we want to add a permanent border around the product dropdown?
I think it looks a little strange with just the highlight border around it. What we have is consistent at least.

Also, on mobile, the hover styles stay on the buttons.
Not too experienced with mobile stuff, so not sure if this is avoidable?
Could just take the style off mobile if all else fails I guess...

rjmcneill:
Will +1 once we decide on what we're doing for those things.
Looks good to me other than those minor things though :)

jdhodges:
On mobile the hover action persists after you tap, so the styling stays in hover mode. We agreed to stay consistent with bootstrap, and the same thing happens on mobile with [bootstrap buttons](http://v4-alpha.getbootstrap.com/components/buttons/).

Issue #638, opened 2016-02-18
Navigator gradient style doesn't match the design


jleft:
Desired style:

![Navigator](https://cloud.githubusercontent.com/assets/2017615/13144900/2d1ece7e-d645-11e5-910a-d8b90e371877.png)

PR #639, opened 2016-02-18
Fixed nav gradient to match UX design.

jdhodges:
#638


src/assets/js/chart/nav.js
```diff
@@ -81,12 +83,21 @@ export default function() {
...
+        .decorate(function(selection) {
+            selection.attr('fill', 'url("#brush-gradient")');
+        });
+
+    var brushLine = fc.series.line()
+        .yValue(function(d) { return d.close; });
 
     var layoutWidth;
 
+    var multi = fc.series.multi()
```
jleft: Maybe `multi` => `brushMaskMulti`, so not to be confused with the other multi series?

djmiley: Good point.

src/assets/js/chart/nav.js
```diff
@@ -84,13 +86,24 @@ export default function() {
...
     var brushMask = fc.series.area()
-      .xValue(function(d) { return d.date; })
-      .yValue(function(d) { return d.close; })
-      .xScale(maskXScale)
-      .yScale(maskYScale);
+        .xValue(function(d) { return d.date; })
+        .y1Value(function(d) { return d.close; })
+        .y0Value(function(d) { return maskYScale.domain()[0]; })
+        .decorate(function(selection) {
+            selection.attr('fill', 'url("#brush-gradient")');
```
jleft: Can we do this on `selection.enter()`?

src/assets/styles/navigator.less
```diff
@@ -60,19 +61,21 @@
       .mask-background {
         fill: @mask-transparent;
       }
-      .area {
-        fill: @mask-opaque;
+
+      .line {
+        stroke-width: 1.2px;
```
jleft: Just wondering why `1.2px` and not `1px`?

jdhodges: `1.2px` looked closer to the ux designs, I can change to 1 if you think it'd be better.

jdhodges: 1px: image

1.2px: image

Which would we prefer?

jleft: Let's go with 1.2px then :)

jleft:
After https://github.com/ScottLogic/BitFlux/pull/639#discussion_r58212929 is adressed and this is rebased, this looks good to me :+1:

jleft:
Just needs the rebase to resolve conflicts then good to merge, thanks @jdhodges :+1:

Issue #640, opened 2016-02-18
Invalid Value for <path> on auto widths


WPFerg:
When trying to run Bitflux using the API on a container with width set to `auto`, the following errors are printed:
```
Error: Invalid value for <path> attribute d="MNaN,6V0H0V6"
Error: Invalid value for <g> attribute transform="translate(NaN,0)"
Error: Invalid value for <g> attribute transform="translate(NaN,0)"
```
All have `xAxis.js:23` as the line causing the error.

PR #641, opened 2016-02-18
X Axis dimension changed now uses util width function to set x scale …

jdhodges:
…range. #640.


jleft:
Thanks @jdhodges :+1: 

If you have a moment, please could you roll out this change to any other components which are measuring the width of an element in a similar way?

jdhodges:
Yeah sure :smile:

Issue #645, opened 2016-02-22
Cannot read property 'displayString' of undefined


WPFerg:
In version 1.1.0, attempting to use Quandl and `changeQuandlProduct` produces the error:
```
Uncaught TypeError: Cannot read property 'displayString' of undefined
(anonymous function)	@	dropdown.js:47
dropdown	@	dropdown.js:46
(anonymous function)	@	overlay.js:36
overlay	@	overlay.js:25
renderInternal	@	app.js:153
```
on every render. @jdhodges had an investigate on this and it seems adding in a `overlay.products.push` into `app.changeQuandlProduct` fixes it.

PR #646, opened 2016-02-22
Quandl products added to overlay model.

jdhodges:
#645.


src/assets/js/app.js
```diff
@@ -441,10 +441,11 @@ export default function() {
 
     app.changeQuandlProduct = function(productString) {
         var product = dataModel.product(productString, productString, [model.periods.day1], model.sources.quandl, '.3s');
-        var existsInHeadMenuProducts = model.headMenu.products.some(function(p) { return p.id === product.id; });
+        var existsInModel = model.headMenu.products.some(function(p) { return p.id === product.id; });
```
jleft: Please can we have two separate checks - one for `existsInHeadMenuProducts` and one for `existsInOverlayProducts`?

PR #647, opened 2016-02-22
Use ID to find selected product in overlay

jleft:
#645

@jdhodges - please could you a quick look over this? :smile:


jdhodges:
Data generator's id is null, is it worth giving it an id?

jleft:
> Data generator's id is null, is it worth giving it an id?

Definitely - nice find!

PR #648, opened 2016-02-22
Added hovered and selected behaviour to dropdown items.

jdhodges:
#579.

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -143,7 +143,7 @@ export default function() {
         var width = currentSeries.option.width(data);
 
         crosshair.decorate(function(selection) {
-            selection.classed('band hidden-xs hidden-sm', true);
+            selection.classed('band', true);
 
             selection.selectAll('.vertical > line')
               .style('stroke-width', width);
@@ -152,7 +152,6 @@ export default function() {
 
     function lineCrosshair(selection) {
         selection.classed('band', false)
-            .classed('hidden-xs hidden-sm', true)
             .selectAll('line')
             .style('stroke-width', null);
     }


src/assets/js/app.js
```diff
@@ -376,7 +376,7 @@ export default function() {
                 return option.isSelected && option.isPrimary;
             });
 
-        model.overlay.primaryIndicators = model.primaryChart.indicators;
+        model.overlay.primaryIndicators = model.headMenu.primaryIndicators = model.primaryChart.indicators;
```
djmiley: Is this best practice? (I don't actually know)

src/assets/js/app.js
```diff
@@ -389,7 +389,7 @@ export default function() {
             chartOption.option.on(event.viewChange, onViewChange);
         });
 
-        model.overlay.secondaryIndicators = charts.secondaries;
+        model.overlay.secondaryIndicators = model.headMenu.secondaryIndicators = charts.secondaries;
```
djmiley: Is this best practice? (I don't actually know)

src/assets/js/menu/selectors.js
```diff
@@ -44,7 +44,7 @@ export default function() {
                     return option.isSelected ? index : null;
                 })
                 .filter(function(option) {
-                    return option;
+                    return option !== null;
```
djmiley: Why this check? When do we get `undefined`?

src/assets/js/menu/generator/dropdown.js
```diff
@@ -63,6 +63,10 @@ export default function() {
         listItemAnchors.append('span')
             .attr('class', 'name');
 
+        listItems.classed('selected', function(d, i) {
+            return model.selected ? model.selected.indexOf(i) > -1 : i === selectedIndex;
```
djmiley: Can we change `selected` to `selectedIndexes` to make it clear that we are doing this when there may be multiple things selected?

djmiley:
Looks like an effective fix at the moment :)

src/assets/js/menu/selectors.js
jdhodges: Your comment made me come up with a much better way of doing this using reduce :smile:

src/assets/js/app.js
jdhodges: I think I'll split this into multiple lines to avoid confusion.

djmiley:
Once that's updated, I'll +1 it.

src/assets/js/app.js
```diff
@@ -389,8 +390,9 @@ export default function() {
             chartOption.option.on(event.viewChange, onViewChange);
         });
 
-        model.overlay.secondaryIndicators = charts.secondaries;
-        // TODO: Remove .remove! (could a secondary chart group component manage this?).
+        model.headMenu.secondaryIndicators = charts.secondaries;
+        model.overlay.secondaryIndicators = model.headMenu.secondaryIndicators;
```
djmiley: `model.overlay.secondaryIndicators = charts.secondaries`

Issue #650, opened 2016-02-23
Update d3fc dep to 6.0.0


jleft:
https://github.com/ScottLogic/d3fc/releases/tag/v6.0.0

rjmcneill:
After an initial look, these things look like they need fixed:

- [ ] Moving average
- [ ] Volume
- [ ] Line series
- [ ] Point series
- [ ] Area series

PR #652, opened 2016-02-23
Update to d3fc v6.0.0

rjmcneill:
Relates to #650

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -28,14 +28,21 @@
                       "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.1.tgz",
                       "dependencies": {
                         "ansi-styles": {
-                          "version": "2.1.0",
+                          "version": "2.2.0",
                           "from": "ansi-styles@>=2.1.0 <3.0.0",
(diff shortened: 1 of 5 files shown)


src/assets/js/chart/primary.js
```diff
@@ -121,6 +124,7 @@ export default function() {
         case 'line':
         case 'point':
         case 'area':
+            currentSeries.option.xValue(function(d) { return d.date; });
```
jleft: Can we not set this when we initialise the models? The function name - `updateYValueAccessorUsed` - is now misleading as it also affects the xValue accessor.

src/assets/js/chart/primary.js
```diff
@@ -109,7 +109,10 @@ export default function() {
 
     function updateMultiSeries() {
         var baseChart = [gridlines, currentSeries.option, closeLine];
-        var indicators = currentIndicators.map(function(indicator) { return indicator.option; });
+        var indicators = currentIndicators.map(function(indicator) {
```
jleft: Can we not set this when we initialise the models? Rather than every time we update the multi?

rjmcneill:
@jleft is this more what you had in mind?

jleft:
> @jleft is this more what you had in mind?

Yeah, exactly - thanks! :+1:

PR #653, opened 2016-02-23
Fixed problem with gridlines not accounting for the closePrice

rjmcneill:
Relates to #587

### DIFF
--- a/src/assets/js/chart/primary.js
+++ b/src/assets/js/chart/primary.js
@@ -191,8 +191,8 @@ export default function() {
 
         // Find current tick values and add close price to this list, then set it explicitly below
         var latestPrice = currentYValueAccessor(model.data[model.data.length - 1]);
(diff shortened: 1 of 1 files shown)


src/assets/js/chart/primary.js
```diff
@@ -207,7 +207,8 @@ export default function() {
                 .attr('transform', 'translate(' + calloutHeight / 2 + ',1)');
           });
 
-        gridlines.yTicks(tickValues.length - 1);
+        var gridlineCount = tickValues.indexOf(latestPrice) === -1 ? tickValues.length : tickValues.length - 1;
+        gridlines.yTicks(gridlineCount);
```
jleft: Could we use `gridlines.yTickValues()` to set the values explicitly? May need to refactor a bit above so there is an array of tick values, excluding the annotation value(s), and another array of the ticks values, including the annotation value(s).

Issue #651, opened 2016-02-23
Add tracking text when tracking latest price


jleft:
![live-price-tracking](https://cloud.githubusercontent.com/assets/2017615/13247299/af27a28c-da10-11e5-92e3-7342bed59df5.PNG)

---

[Design.pdf](https://github.com/ScottLogic/BitFlux/files/142225/Changes_for_initial_bitflux_meetings.pdf)

PR #654, opened 2016-02-24
Add tracking text when tracking latest price

rjmcneill:
Relates to #651

### DIFF
--- a/src/assets/styles/navigator.less
+++ b/src/assets/styles/navigator.less
@@ -81,24 +81,35 @@
     }
   }
 
-  #navbar-reset {
-    flex: 0 0 24px;
-    height: 40px;
-    opacity: 0.2;
-    margin: 10px 21px 0 15px;
+  #reset-container {
+    display: flex;
+    width: @y-axis-width;
+    height: 60px;
+    flex-direction: column;
 
-    .reset-button {
-      stroke: @primary-white;
-      fill: @primary-dark-blue;
-      shape-rendering: auto;
+    #navbar-reset {
+      opacity: 0.2;
+      margin: 5px 21px 0 15px;
+      
+      .reset-button {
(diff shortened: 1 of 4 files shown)


djmiley:
Looks good to me, is the css what is specified by UX?

rjmcneill:
@djmiley they showed that the actual label should be in line with the left of the arrow, but it didn't look right like that.
I made it aligned to the center of it.  Happy to change it if people disagree though!

jdhodges:
Do we need a bigger gap between the button and the text? 
image image
Gap seems to be a bit larger on the UX designs.

rjmcneill:
@jdhodges - updated.
That's a far as we can go I reckon

src/assets/styles/navigator.less
```diff
@@ -77,23 +77,35 @@
...
+
+      .reset-button > path {
+        cursor: pointer;
       }
     }
 
-    .reset-button > path {
-      cursor: pointer;
+    .tracking-text {
+      margin: 1px 0 0 7px;
```
jdhodges: How about:
```margin: 1px 0 -4px 7px;```
Not sure if having a negative margin value is good practice but it gives us a bigger space that's closer to the UX design.

rjmcneill:
Current PR implementation (zero bottom margin):
zeromargintrackingtext

Suggested implementation (negative bottom margin):
minusmargintrackingtext

I quite like the current implementation because it lines the text up with the axis, but I'll change it if we want to increase the gap.

rjmcneill:
@jleft can this be closed now with the #702 PR the new suggestion from UX?

jleft:
> @jleft can this be closed now with the #702 PR the new suggestion from UX?

I think we may still want this - even if it's something which we disable for now.

Issue #613, opened 2016-02-05
Add crisp edge shape rendering to specific elements


jleft:
I propose we change `shape-rendering: crispEdges` as the default to all top-level chart SVG elements, then override to `auto` for the line, area and point series (or anything that doesn't look so good with this setting).

Alternatively, we can just target specific elements/groups which should have  `shape-rendering: crispEdges`, but I expect this will be the majority.

Related to #603 and #611

tlsim:
Is it just me that likes the antialiasing? Horizontal and vertical lines might benefit from crispEdges / rounding pixel values, but only if the page is not zoomed (e.g grid lines start to disappear at < 100%).

jleft:
I think the antialiasing works well for elements without horizontal/vertical lines - like the line series, area series, point series etc.

I think I prefer the crisper edges for OHLC, candlestick, grid lines, axes etc. e.g.

#### `crispEdges`
![sr-ce](https://cloud.githubusercontent.com/assets/2017615/12882519/19f7d596-ce4a-11e5-983a-0430a3ca4a48.PNG)

#### `auto`
![sr-auto](https://cloud.githubusercontent.com/assets/2017615/12882520/1af5ec08-ce4a-11e5-9d79-9a7e7705ee48.PNG)

jdhodges:
Yeah looks good with crisp edges.

rjmcneill:
Closed in #661

PR #661, opened 2016-02-26
Added crispEdges shape rendering

rjmcneill:
Relates to #613 

CrispEdges is now the default.
Added auto to:
- Moving average
- Bollinger bands
- RSI line
- MACD lines
- Nav lines
- Nav handle circles
- Line series
- Crosshair circle
- Point series
- Area series
- Reset button

One thing I'm not sure about is how the icons look in the dropdown with them.
Happy to remove it from these depending on the feedback.

### DIFF
--- a/src/assets/icons/bar-series.svg
+++ b/src/assets/icons/bar-series.svg
@@ -1,5 +1,5 @@
 <svg viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg">
-  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd">
+  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd" shape-rendering="crispEdges">
     <rect x="2" y="15" height="5" width="5"></rect>
     <rect x="9" y="10" height="10" width="5"></rect>
     <rect x="16" y="5" height="15" width="5"></rect>
--- a/src/assets/icons/bar-series.svg
+++ b/src/assets/icons/bar-series.svg
@@ -1,5 +1,5 @@
 <svg viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg">
-  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd" shape-rendering="crispEdges">
+  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd">
     <rect x="2" y="15" height="5" width="5"></rect>
     <rect x="9" y="10" height="10" width="5"></rect>
     <rect x="16" y="5" height="15" width="5"></rect>
--- a/src/assets/styles/base.less
+++ b/src/assets/styles/base.less
@@ -8,13 +8,16 @@ html, body {
   line-height: @line-height-base;
   height: 100%;
   min-width: @min-width-supported;
-
-  svg {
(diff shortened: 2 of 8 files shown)


src/assets/styles/base.less
```diff
@@ -8,6 +8,7 @@ html, body {
   line-height: @line-height-base;
   height: 100%;
   min-width: @min-width-supported;
+  shape-rendering: crispEdges;
```
jleft: Not sure, if it's worth just targeting `svg`, as `shape-rendering` will only apply to the SVG elements?

src/assets/styles/secondary.less
```diff
@@ -28,10 +28,7 @@
   .rsi {
     stroke: @rsi-line-colour;
     stroke-width: @stroke-width;
-
-    .annotation line {
-      shape-rendering: crispEdges;
-    }
+    shape-rendering: auto;
```
jleft: Won't this also affect the annotations (which I think should still have `crispEdges`)?

src/assets/styles/secondary.less
```diff
@@ -28,6 +28,7 @@
   .rsi {
     stroke: @rsi-line-colour;
     stroke-width: @stroke-width;
+    shape-rendering: auto;
```
jleft: Can we follow the general rule of only setting `shape-rendering: crispEdges` at the top level and overriding with `shape-rendering: auto` for exceptions?

So, within `.rsi`, remove the `.annotation line` rule below and set `auto` on  `.indicator .line` (I think that's the right class)?

jleft:
Looks good overall, thanks @rjmcneill! There seems to be a fair bit of repetition with setting `shape-rendering: auto` to the `.line` class - is this necessary (so to be specific enough), or can this just be set once?

rjmcneill:
@jleft I moved the `shape-rendering: auto` to the top level in the 3rd commit there.
It looks good to me, and I tried to remove what I could in the lower levels but let me know if you can see anything I've missed and I'll sort it out! :smiley:

src/assets/icons/bar-series.svg
```diff
@@ -1,5 +1,5 @@
 <svg viewBox="0 0 25 25" xmlns="http://www.w3.org/2000/svg">
-  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd">
+  <g stroke="#7B9FA1" fill="none" fill-rule="evenodd" shape-rendering="crispEdges">
```
jleft: Having had a look at these, I think it's worth spending the time trying to pixel align the SVG manually, or using another optimsed image format (PNG?) for this - I'll open an separate issue.

Please can `shape-rendering="crispEdges"` be removed from the icons?

src/assets/styles/base.less
```diff
@@ -8,6 +8,15 @@ html, body {
   line-height: @line-height-base;
   height: 100%;
   min-width: @min-width-supported;
+
+  svg {
```
jleft: I'd move this and `path.line` out of `html, body { ...`

jleft:
Thanks for this @rjmcneill - looks good to me, just a couple more comments to address then I think this is good to go

jleft:
Rebased in #710

Issue #657, opened 2016-02-25
Add option to set which indicators and displayed when the application is initialised


jleft:

jdhodges:
Is this only aimed at setting the indicators upon initialisation? It will cause layout issues if someone tries to change the indicators while the app is running.

jleft:
> Is this only aimed at setting the indicators upon initialisation?

No, should work at any time.

> It will cause layout issues if someone tries to change the indicators while the app is running.

What sort of issues? Have you got an example/steps to reproduce?

jdhodges:
If I run
```
var app = BitFlux.app()
    .fetchCoinbaseProducts(true)
    .indicators(['rsi']);

app.run('#app-container');

app.indicators(['rsi', 'macd']);
```

I get:
![image](https://cloud.githubusercontent.com/assets/14218569/13636623/ba356d44-e5fa-11e5-8334-bdf1d2295f5b.png)

and for
```
var app = BitFlux.app()
    .fetchCoinbaseProducts(true)
    .indicators(['rsi', 'macd']);

app.run('#app-container');

app.indicators(['rsi']);
```
I get:
![image](https://cloud.githubusercontent.com/assets/14218569/13636644/db63689a-e5fa-11e5-9319-1d541bf16243.png)

PR #662, opened 2016-02-26
Added and exposed option to set which indicators to display.

jdhodges:
#657

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -104,6 +104,8 @@ export default function() {
 
     var proportionOfDataToDisplayByDefault = 0.2;
 
+    var indicators;
+
     var firstRender = true;
     function renderInternal() {
         if (firstRender) {
@@ -125,7 +127,7 @@ export default function() {
             .filter(function(d, i) { return i < charts.secondaries.length; })
             .each(function(d, i) {
                 d3.select(this)
-                    .attr('class', 'secondary-container ' + charts.secondaries[i].valueString)
+                    .attr('class', 'secondary-container secondary-' + charts.secondaries[i].valueString)
                     .call(charts.secondaries[i].option);
             });
 
@@ -199,12 +201,14 @@ export default function() {
 
     function onPrimaryIndicatorChange(indicator) {
         indicator.isSelected = !indicator.isSelected;
+        indicators[indicator.valueString] = indicator.isSelected;
         updatePrimaryChartIndicators();
         render();
     }
 
     function onSecondaryChartChange(_chart) {
         _chart.isSelected = !_chart.isSelected;
+        indicators[_chart.valueString] = _chart.isSelected;
         updateSecondaryCharts();
         render();
     }
@@ -433,6 +437,23 @@ export default function() {
         render();
     }
 
+    function initialiseIndicators() {
+        var _indicators = indicators || {};
+
+        model.selectors.indicatorSelector.options.forEach(function(indicator) {
+            if (_indicators[indicator.valueString]) {
+                indicator.isSelected = true;
+            } else {
+                _indicators[indicator.valueString] = false;
+            }
+        });
+
+        updatePrimaryChartIndicators();
+        updateSecondaryCharts();
+
+        return _indicators;
+    }
+
     app.fetchCoinbaseProducts = function(x) {
         if (!arguments.length) {
             return fetchCoinbaseProducts;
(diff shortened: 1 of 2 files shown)


jdhodges:
Can set the indicators to be displayed like this:
```
app().indicators({
    movingAverage: true,
    bollinger: true,
    rsi: true,
    macd: true,
    volume: true
});
```

src/assets/js/app.js
```diff
@@ -510,6 +541,12 @@ export default function() {
         overlay = initialiseOverlay();
         toastNotifications = initialiseNotifications();
 
+        if (indicators) {
```
djmiley: Can we not just call `selectIndicators(indicators)` with `indicators` falsy; the method may need to be slightly changed, but probably not much.

src/assets/js/initialiseModel.js
```diff
@@ -115,11 +115,11 @@ export default function() {
         var indicators = [
```
djmiley: Do we need to change the names of these indicators?

src/assets/js/app.js
```diff
@@ -125,7 +127,7 @@ export default function() {
             .filter(function(d, i) { return i < charts.secondaries.length; })
             .each(function(d, i) {
                 d3.select(this)
-                    .attr('class', 'secondary-container ' + charts.secondaries[i].valueString)
+                    .attr('class', 'secondary-container secondary-' + charts.secondaries[i].valueString)
```
djmiley: See below.

src/assets/js/app.js
```diff
@@ -199,12 +201,14 @@ export default function() {
 
     function onPrimaryIndicatorChange(indicator) {
         indicator.isSelected = !indicator.isSelected;
+        indicators[indicator.valueString] = indicator.isSelected;
```
djmiley: @jleft Do we need this? (I don't know, are we passing the config back out of the app?)

src/assets/js/app.js
```diff
@@ -199,12 +201,14 @@ export default function() {
...
         updatePrimaryChartIndicators();
         render();
     }
 
     function onSecondaryChartChange(_chart) {
         _chart.isSelected = !_chart.isSelected;
+        indicators[_chart.valueString] = _chart.isSelected;
```
djmiley: @jleft Do we need this? (I don't know, are we passing the config back out of the app?)

src/assets/js/initialiseModel.js
jdhodges: just thought it would be nicer when using the api:
```
rsi: true
```
rather than
```
secondary-rsi: true
```

Having 'secondary-' in the valueString also isn't required now we have ```isPrimary```

src/assets/js/app.js
jdhodges: If indicators hasn't been set, we need to create the object so it can be retrieved using the api. If it has been set we need to select the necessary indicators. With a single call we'd end up having to both these jobs in the same function. My intention was to keep these concerns separate, but I suppose this way would be more succinct. Any thoughts?

src/assets/js/app.js
jleft: The API should allow the user to inspect which indicators are visible and which are hidden.

src/assets/js/initialiseModel.js
djmiley: Makes sense, good idea. :)

src/assets/js/app.js
djmiley: Ok, great. :)

src/assets/js/app.js
djmiley: Forget this. :)

src/assets/js/app.js
djmiley: We could initialise the indicators on line 107, and then we can just call `selectedIndicators(indicators)` without the check on truthy/falsy.

jdhodges: That's what I initially went for, but then I thought it's probably a good idea to initialise the indicators at runtime. Someone might add an indicator to the model and then not realise the change isn't being reflected in the API. ```initialiseIndicators()``` gets the indicators from the model so if we were to add, remove, or edit any indicators, we wouldn't have to also reflect those changes at line 107.

djmiley: Would that not entail re-running app.js anyway?

rjmcneill:
Would it be worth adding this to main.js:

```JavaScript
BitFlux().app()
    .indicators({
        movingAverage: false,
        bollinger: false,
        rsi: false,
        macd: false,
        volume: false
    })
```

Just would make it clear to the openFinners and wouldn't cause us any issues.

Issue #575, opened 2016-01-27
Use skip weekends discontinuity provider for Quandl


jleft:
https://d3fc.io/components/scale/dateTime.html

PR #663, opened 2016-02-26
Added discontinuity provider for quandl data

rjmcneill:
Relates to #575


rjmcneill:
Could maybe pull the functionality out into a util file.
Happy enough to do that if everyone thinks it'll be easier to follow/tidier

jdhodges:
I reckon a util function would be a good idea since we're doing the same thing in 4 places.

rjmcneill:
Looks like this is causing the view domain to change for some reason.
Looking into it now

rjmcneill:
I've added a few things for us to help testing (and I ran out of quandl requests yesterday).  So I've made data generator exclude weekend data, and I've added skip weekends all sources.

So here's the problem:
We have our domains set via an upper and lower limit.
This means that it fills in all the dates for us (including weekends)
So what that means is that the domains are looking for the Sunday and Saturday times, then eventually snapping to Friday when they can't get that on the scale.

I have already tried setting the domain explicitly using the extent and filter code.
However, that means that the data dates are the only pieces of information for the domain, in turn meaning that the domain can only snap to dates contained in the data, meaning that the scrolling jumps between points of data.

rjmcneill:
[Alternative with discontinuity on the model] (https://github.com/rjmcneill/d3fc-showcase/tree/enhancement/discontinuity-provider-alternative)

jleft:
> Alternative with discontinuity on the model

I think I'd have a preference for this approach

src/assets/js/chart/nav.js
```diff
@@ -146,6 +151,7 @@ export default function() {
           .yDomain(yExtent);
 
         brush.on('brush', function() {
+            console.log('brushing');
```
rjmcneill: Remove

src/assets/js/model/chart/secondary.js
```diff
@@ -1,8 +1,9 @@
-export default function(initialProduct) {
+export default function(initialProduct, sources) {
```
jleft: Need to add `discontinuity` property and pass in `initialDiscontinuity`?

src/assets/js/behavior/zoom.js
```diff
@@ -65,6 +70,7 @@ export default function(width) {
                   }
 
                   if (domain[0].getTime() !== domain[1].getTime()) {
+                      // console.log(domain);
```
rjmcneill: Remove

Issue #675, opened 2016-03-07
Investigate only rendering visible data


jleft:

jdhodges:
The performance impact varies based on how zoomed in you are, but for the default view with all indicators added, the render takes 33% less time with only visible data. This has no detrimental effect on the app at all so is well worth adding in. I'll add a PR in a moment.

PR #676, opened 2016-03-07
Only render visible data for primary and secondaries.

jdhodges:
#675.


jleft:
Nice work, just made a couple of small changes and rebased in #748 :+1:

Issue #674, opened 2016-03-07
Investigate sampling data


jleft:
https://d3fc.io/components/sampler/largestTriangleOne.html
https://d3fc.io/components/sampler/largestTriangleThreeBucket.html
https://d3fc.io/components/sampler/modeMedian.html

jdhodges:
We can half the render time of the navigator if we use a sampler with a bucket size of 4. This may mean we lose a bit too much detail so 2 or 3 may be preferable, thoughts? 
![image](https://cloud.githubusercontent.com/assets/14218569/13573965/01e91e9c-e47a-11e5-9ebd-7008e286c9ce.png)

jdhodges:
May be worth varying the bucket size based on the number of data points.

jleft:
> May be worth varying the bucket size based on the number of data points.

Sounds like a good idea to me :+1:

How about varying based on the width of the chart too? I.e. Would it make sense to have more buckets with a wider chart? I'm not so sure about this though...

PR #679, opened 2016-03-09
Nav data sampled based on layout width and number of data points.

jdhodges:
#674.


jdhodges:
Fixed some indentation so it's a good idea to omit white space changes
https://github.com/ScottLogic/BitFlux/pull/679/files?w=1

src/assets/js/chart/nav.js
```diff
@@ -130,20 +134,21 @@ export default function() {
     function nav(selection) {
         var model = selection.datum();
 
-        createDefs(selection, model.data);
+        sampler.bucketSize((model.data.length * layoutWidth) / 1e5);
```
jleft: Why `1e5`? Feels like a magic number?

Would this be better expressed as a function where we can pass in a desired number of buckets and a width to generate an appropriate bucket size?

```js
// for example: bucketSizeForBucketsPerWidth(1, 10)
// could return a bucketSize which will give us 1 bucket every 10 pixels

```

src/assets/js/chart/nav.js
```diff
@@ -19,82 +19,88 @@ export default function() {
...
+                    [viewScale.domain()[0], navChart.yDomain()[0]],
+                    [viewScale.domain()[1], navChart.yDomain()[1]]
+                ]);
+            } else {
+                // This stops the brush data being overwritten by the point data
+                return this.data;
+            }
+        });
 
+    var maskXScale = fc.scale.dateTime();
```
jleft: Is this unused now?

src/assets/js/chart/nav.js
```diff
@@ -19,82 +19,88 @@ export default function() {
...
     var maskYScale = d3.scale.linear();
```
jleft: Is this unused now?

src/assets/js/chart/nav.js
```diff
@@ -120,35 +126,41 @@ export default function() {
...
 
+    function bucketSizeForBucketsPerWidth(buckets, width, numberOfPoints) {
+        var bucketSize = numberOfPoints / ((buckets / width) * layoutWidth);
+        return bucketSize > 1 ? bucketSize : 1;
+    }
+
     function nav(selection) {
         var model = selection.datum();
 
+        sampler.bucketSize(bucketSizeForBucketsPerWidth(1, 10, model.data.length));
```
jleft: I know this was my suggestion - but after trying it out, I'm not convinced that varying the bucket size based on width is such a good idea - sorry!

I think adjusting it based upon the number of data points is still a good idea though.

Let me know if you disagree :)

src/assets/js/chart/nav.js
```diff
@@ -166,25 +178,26 @@ export default function() {
                     dispatch[event.viewChange](util.domain.centerOnDate(
                         model.discontinuityProvider,
                         viewScale.domain(),
-                        model.data,
+                        sampledData,
```
jleft: Should we use sampled data or data here?

src/assets/js/chart/nav.js
```diff
@@ -166,25 +178,26 @@ export default function() {
...
+            .scale(viewScale)
+            .trackingLatest(model.trackingLatest)
+            .allowPan(false)
+            .on('zoom', function(domain) {
+                dispatch[event.viewChange](domain);
+            });
 
         selection.select('.plot-area')
-          .call(zoom);
+            .datum({data: sampledData})
```
jleft: Should we use sampled data or data here?

PR #680, opened 2016-03-09
Fixed setting secondary indicators from API while app is running.

jdhodges:
https://github.com/ScottLogic/BitFlux/issues/657#issuecomment-194296517
#681


jleft:
Thanks @jdhodges but I think this is no longer needed as a fix has been merged in #749 (please reopen if you disagree).

PR #684, opened 2016-03-15
Add skip weekend discontinuity provider

jleft:
Relates to #575, #686
Closes #663


src/assets/js/app.js
```diff
@@ -253,12 +257,23 @@ export default function() {
         model.nav.data = data;
     }
 
+    function updateDiscontinuity(productSource) {
+        var discontinuity = util.discontinuityProvider(productSource, model.discontinuousSources);
+
+        model.xAxis.discontinuity = discontinuity;
```
jleft: General point: would  `discontinuityProvider` be more appropriate than `discontinuity` as the name on the models?

rjmcneill: Yea, I think it would be more explicit

jleft: Cool, I'll update

src/assets/js/util/discontinuityProvider.js
```diff
@@ -0,0 +1,18 @@
...
+export default function(productSource, discontinuousSources) {
+    var skip = false;
+
+    if (!Array.isArray(discontinuousSources)) {
+        discontinuousSources = [discontinuousSources];
+    }
+
+    discontinuousSources.forEach(function(discontinuousSource) {
+        if (productSource === discontinuousSource) {
+            skip = true;
```
rjmcneill: My code I know, but what about this @jleft ?
```JavaScript
    discontinuousSources.forEach(function(discontinuousSource) {
        if (productSource === discontinuousSource) {
            return skipWeekends();
        }
    });

    return fc.scale.discontinuity.identity();
```
Would also leave us open to adding in future discontinuity providers if needed

jdhodges:
When panning all the way to the left it begins to zoom in.

[bandicam 2016-03-16 10-47-43-621.zip](https://github.com/ScottLogic/BitFlux/files/175769/bandicam.2016-03-16.10-47-43-621.zip)

src/assets/js/util/discontinuityProvider.js
jleft: I'm not convinced that a discontinuity provider has a direct relationship with a source - I can imagine a source providing data sets which are and are not discontinuous (depending on the product/period). I think a discontinuity provider could relate to a product/period pair. However, for now, this solves are problem, so let's roll with it :).

In 0a62c9c I've attempted to simplify this by adding the discontinuity provider to the source - what do you think?

rjmcneill: I guess quandl with the period set to weekly doesn't need the skipWeekends.
Yea, I like that approach for now though, definitely simplifies it.

In the future I guess we could set it on-the-fly though with testing the data for Saturday and Sunday information.

jleft:
Let's merge what we have :+1: then we can have another round of testing :) Thanks @rjmcneill!

rjmcneill:
Cheers @jleft 
We got there eventually!

Issue #688, opened 2016-03-16
Add option to public API to set callback to handle historic error events


jleft:
- If a callback is set externally, notification messages should not be added for this event
- The callback should be passed the `err` object from the request
- The callback can (optionally) return a string which would be shown in the error alert
- The chart should still be obscured if there's an error, even if there isn't an error message passed in
- The loading icon should be shown/hidden as before

PR #689, opened 2016-03-16
Expose option to set historic feed error callback handler

jleft:
#688

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -86,6 +86,8 @@ export default function() {
 
     var _dataInterface = initialiseDataInterface();
 
+    var externalHistoricFeedErrorCallback;
+
     var charts = {
         primary: undefined,
         secondaries: [],
@@ -240,10 +242,14 @@ export default function() {
 
     function loading(isLoading, error) {
         var spinner = '<div class="spinner"></div>';
-        var errorMessage = '<div class="content alert alert-info">' + error + '</div>';
+        var obscure = arguments.length > 1 || isLoading;
 
+        var errorMessage = '';
+        if (error && error.length) {
+            errorMessage = '<div class="content alert alert-info">' + error + '</div>';
+        }
         containers.app.select('#loading-status-message')
-            .classed('hidden', !(isLoading || error))
+            .classed('hidden', !obscure)
             .html(error ? errorMessage : spinner);
     }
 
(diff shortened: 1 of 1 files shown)


src/assets/js/app.js
```diff
@@ -309,21 +315,26 @@ export default function() {
...
-                loading(false, 'Error loading data. Please make your selection again, or refresh the page.');
-                var responseText = '';
-                try {
-                    var responseObject = JSON.parse(err.responseText);
-                    var formattedMessage = source.historicNotificationFormatter(responseObject);
-                    if (formattedMessage) {
-                        responseText = '. ' + formattedMessage;
+                if (externalHistoricFeedErrorCallback) {
+                    var error = externalHistoricFeedErrorCallback(err) || true;
+                    loading(false, error);
```
rjmcneill: If `externalHistoricFeedErrorCallback(err)` isn't set, will this then just display `true` in the notification?

jleft: No sure exactly what you mean by isn't set? If `externalHistoricFeedErrorCallback` is falsey, then we won't get to this code at all; if it's truthy then we will.

Line 319 is checking the return value of `externalHistoricFeedErrorCallback(err)` - it allows the user to return a string from their callback, which can be set as the error message - if they don't return anything (or return something falsey) then `error` is set to `true`, so the chart will still be obscured by the overlay when there is an error, but no message will be shown.

rjmcneill: > but no message will be shown

That's the part I wasn't sure about.
I wasn't sure if `true` would get cast to a string or what would happen.

rjmcneill:
Can't +1 on ZenHub atm for some reason, but LGTM

src/assets/js/app.js
jleft: This is the loading function:

```js
function loading(isLoading, error) {
    var spinner = '<div class="spinner"></div>';
    var obscure = arguments.length > 1 || isLoading;

    var errorMessage = '';
    if (error && error.length) {
        errorMessage = '<div class="content alert alert-info">' + error + '</div>';
    }
    containers.app.select('#loading-status-message')
        .classed('hidden', !obscure)
        .html(error ? errorMessage : spinner);
}
```

If `error` is `true`, then this `if (error && error.length)` will evaluate to false, so the message will remain as `''`;

jleft:
Thanks for the reviews guys :)

Issue #690, opened 2016-03-17
View domain doesn't quite reach the extents


rjmcneill:
1) Drag the viewDomain in the nav to the far left 
2) Slowly zoom out
3) At the point when the zoom looks like it completely zoomed out the tracking latest won't be active
4) Zooming out one more click will activate tracking latest

Doesn't work every time.

PR #695, opened 2016-03-18
Implemented rounding of the domains

rjmcneill:
Relates to #690

### DIFF
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -75,6 +75,16 @@ export default function(width) {
 
                   domain = clampDomain(domain, selection.datum().data, xExtent);
 
+                  // Round upper domain to upper extent
+                  if (zoomed && scale(domain[1]) === width) {
+                      domain[1] = xExtent[1];
+                  }
+
+                  // Round lower domain to lower extent
+                  if (zoomed && scale(domain[0]) === 0) {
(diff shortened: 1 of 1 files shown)


src/assets/js/behavior/zoom.js
```diff
@@ -75,6 +75,16 @@ export default function(width) {
 
                   domain = clampDomain(domain, selection.datum().data, xExtent);
 
+                  // Round upper domain to upper extent
+                  if (zoomed && scale(domain[1]) === width) {
```
jleft: Will this not round the domain to the xExtent on any 'normal' zoom event, I.E. when the following conditions are *not* met (or they do not change the domain):

- max domain viewed
- tracking latest
- domain clamped

As `scale.domain()` will be the same `domain`, so `scale(domain[0])` and `scale(domain[1])` will return `scale.range()[0]` (`0`) and `scale.range()[1]` (`width`), respectively, causing the rounding to occur.

Conversely, if any of the conditions in the list above are met, then the domain will have been changed, but the scale's domain will not have been updated, so the calculations here will not take this into account, leading to incorrect calculations.

---

Rather than comparing the corresponding pixel values of the domain to `0` and `width`, should they be compared to the corresponding pixel values of the `xExtent`? `scale(xExtent[0])` and `scale(xExtent[1])`? (And rounded to see if they're within a pixel of each other.)

PR #697, opened 2016-03-21
Slight performance enchancement for crosshair

rjmcneill:
Relates to #696 
Relates to #746

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -126,6 +126,11 @@ export default function() {
         }
     }
 
+    function renderLegend() {
+        containers.legend.datum(model.legend)
+            .call(charts.legend);
+    }
+
     var render = fc.util.render(renderInternal);
 
     var layoutRedrawnInNextRender = true;
@@ -181,7 +186,7 @@ export default function() {
 
     function onCrosshairChange(dataPoint) {
         model.charts.legend.data = dataPoint;
-        render();
+        renderLegend();
     }
 
     function onStreamingFeedCloseOrError(streamingEvent, source) {
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -126,12 +126,12 @@ export default function() {
         }
     }
 
-    function renderLegend() {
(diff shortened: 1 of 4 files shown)


jdhodges:
Seems to be a slight issue. Try reloading the page then changing the series to area and back to candlestick again.

rjmcneill:
@jdhodges should be resolved now :)

src/assets/js/app.js
```diff
@@ -162,6 +162,14 @@ export default function() {
         }
     }
 
+    function renderLegend() {
```
rjmcneill: Would `renderCrosshairChange` maybe be a more appropriate name for this?

src/assets/js/chart/primary.js
```diff
@@ -139,8 +139,16 @@ export default function() {
         model.selectorsChanged = false;
     }
 
-    function bandCrosshair(data) {
-        var width = currentSeries.option.width(data);
+    function bandCrosshair(model) {
+        var mapFunction = function(d, i) { return xScale(xValue(d, i)); };
+        var barWidth = currentSeries.option.barWidth();
+        var xValue = currentSeries.option.xValue();
+        var yValue = function(d) { return d.volume; };
```
rjmcneill: `d.volume` change this

src/assets/js/chart/primary.js
```diff
@@ -139,8 +139,16 @@ export default function() {
...
+        var y0Value = d3.functor(0);
+        var x0Value = d3.functor(0);
+
+        var filteredData = model.data.filter(fc.util.fn.defined(x0Value, y0Value, xValue, yValue));
+        var width = barWidth(filteredData.map(mapFunction));
```
jdhodges: Don't think we need all of this, how about just this:
```
var xValue = currentSeries.option.xValue();
var mapFunction = function(d, i) { return xScale(xValue(d, i)); };
var barWidth = currentSeries.option.barWidth();

var width = barWidth(model.data.map(mapFunction));
```

Issue #700, opened 2016-03-23
Add multi-line, clearer x-axis labels


jleft:
https://github.com/ScottLogic/StockFlux/issues/666

PR #703, opened 2016-03-24
Updated xAxis labelling to match new UX designs

rjmcneill:
Relates to #700 
https://github.com/ScottLogic/StockFlux/issues/666

### DIFF
--- a/src/assets/js/chart/xAxis.js
+++ b/src/assets/js/chart/xAxis.js
@@ -11,10 +11,25 @@ export default function() {
 
     function preventTicksMoreFrequentThanPeriod(period) {
         var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
+
         if (scaleTickSeconds < period.seconds) {
             xAxis.ticks(period.d3TimeInterval.unit, period.d3TimeInterval.value);
         } else {
             xAxis.ticks(6);
+        }
+    }
+
+    function adaptTickFormatting() {
+        var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
+
+        if (scaleTickSeconds <= 1800) {
+            xAxis.tickFormat(d3.time.format('%H:%M, %d %b'));
+        } else if (scaleTickSeconds <= 21600) {
+            xAxis.tickFormat(d3.time.format('%I %p, %a %d %b'));
+        } else if (scaleTickSeconds <= 608400) {
+            xAxis.tickFormat(d3.time.format('%a %d, %b %Y'));
+        } else if (scaleTickSeconds <= 2592000) {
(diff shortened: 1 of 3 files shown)


src/assets/js/chart/xAxis.js
```diff
@@ -18,6 +19,20 @@ export default function() {
         }
     }
 
+    function adaptTickFormatting() {
+        var scaleTickSeconds = (xScale.ticks()[1] - xScale.ticks()[0]) / 1000;
```
jleft: I know this is based on the existing code above, but do we need to take into account if we're using a discontinuous scale? I.e. skip weekends.

src/assets/js/chart/xAxis.js
```diff
@@ -18,6 +19,20 @@ export default function() {
...
+
+        if (scaleTickSeconds <= 1800) {
```
jleft: Feels like there's quite a few magic numbers and if/then/elses here - maybe _[d3.time.format.multi(formats)](https://github.com/mbostock/d3/wiki/Time-Formatting#format_multi)_ would help here?

src/assets/js/chart/xAxis.js
```diff
@@ -25,7 +40,28 @@ export default function() {
         xScale.discontinuityProvider(model.discontinuityProvider);
 
         preventTicksMoreFrequentThanPeriod(model.period);
+        adaptTickFormatting();
+
         selection.call(xAxis);
+
+        selection.selectAll('text')
```
jleft: If we don't use the d3fc axis (see comment below), then could chain selectAll to the call(xAxis) above.

Might be worth selecting the tick class too? `.tick text`

src/assets/js/chart/xAxis.js
```diff
@@ -25,7 +40,28 @@ export default function() {
...
+            .call(function(text) {
```
jleft: Maybe a good opportunity to use the d3fc axis and `decorate()`?

src/assets/js/chart/xAxis.js
```diff
@@ -25,7 +40,28 @@ export default function() {
...
+                text.each(function() {
+                    var _this = d3.select(this);
```
jleft: Stylistically, I'm not particularly keen on `_this` as the variable name - I don't think we do this anywhere else and I think it could be more descriptive - something like `text` or `tickText`?

src/assets/js/chart/xAxis.js
```diff
@@ -25,7 +40,28 @@ export default function() {
...
+                    var split = _this.text().split(',');
```
jleft: Based on the formats above, in `adaptTickFormatting()`, which always have a space after the `, `, won't `split[1]` have a space as the first character? Which I don't think we want.

jleft:
Thanks @rjmcneill - looks like a really good start to me :)

Issue #701, opened 2016-03-23
Navigation Bar is displaying inconsistent labels


jackarnstein:
![image](https://cloud.githubusercontent.com/assets/5921987/13981531/89c0f59c-f0dc-11e5-9880-5eae8b94ee9c.png)

Month, day of week, month, year, month, month

Seems a little inconsistent?

jleft:
Is this just with the GOOG product (i.e. when we're skipping weekends)?

jleft:
> Is this just with the GOOG product (i.e. when we're skipping weekends)?

Confirmed.

PR #709, opened 2016-03-29
Nav tick values ignore discontinuity

jdhodges:
#701


jdhodges:
This is a problem with the discontinuity provider not reconciling with the [default time format](https://github.com/mbostock/d3/wiki/Time-Formatting#format_multi). We're getting weird formats because the date is being clamped when a tick coincides with a weekend. In this PR I'm generating the tick values without discontinuity which seems to do what we want:

image

Issue #713, opened 2016-03-30
Do not render a component if it is hidden?


jleft:
Potential performance enhancement.

For example, the navigator is hidden on mobile displays, but we still render it - we may be able to improve performance by not rendering hidden components.

jdhodges:
I have managed to get this to work with the navigator and it seems to be hitting 60fps quite comfortably - my approach uses a util function to deduce whether a selection is visible before rendering:

```
function nav(selection) {
    if (util.isVisible(selection)) {
        var model = selection.datum();
        ....
```

Would this be ok or would we prefer to do something at app.js level?

jleft:
Great stuff, how about [extending d3.selection()](https://github.com/mbostock/d3/wiki/Selections#d3_selection) to add a convenience method for this?

```js
// Not sure about the name... any better ideas?
d3.selection.prototype.callIfVisible = function(callback) {
    // if selection is visible, call, otherwise don't
};
```

If this seems reasonable, I think it makes most sense to do it app.js.

PR #715, opened 2016-03-30
Only renders the nav if it is visible

jdhodges:
#713


jdhodges:
`isVisible` uses [`HTMLElement.offsetParent`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) which will return null if the element or any of its ancestors are hidden.

rjmcneill:
Only thing with this is according to chrome console:
`SVGElement.offsetParent' is deprecated and will be removed in M50, around April 2016.`

Same issue with `SVGElement.offsetWidth`
We might want to port these both over to an alternative at some point soon I'm guessing?

jdhodges:
[`Element.getBoundingClientRect()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) is recommended as an alternative. Looks like it'll do  what we want in both cases.

Issue #718, opened 2016-04-01
Selected indicators should be highlighted in the indicator dropdown


jleft:
![image](https://cloud.githubusercontent.com/assets/2017615/14205169/2df08eae-f801-11e5-8303-0c337e23a35b.png)

jleft:
Resolved in #719

Issue #717, opened 2016-04-01
Dropdowns should have new hover style


jleft:
Border, instead of background colour change:

![image](https://cloud.githubusercontent.com/assets/2017615/14205165/2838b900-f801-11e5-8532-24b4abb8377d.png)

jdhodges:
Implemented in #648.

jleft:
Resolved in #719

PR #719, opened 2016-04-01
Added hovered and selected behaviour to dropdown items.

jdhodges:
#717 #718 #579. Replaces #648.

### DIFF
--- a/src/assets/js/menu/selectors.js
+++ b/src/assets/js/menu/selectors.js
@@ -40,18 +40,15 @@ export default function() {
             var options = model.indicatorSelector.options;
 
             var selectedIndicatorIndexes = options
-                .map(function(option, index) {
-                    return option.isSelected ? index : null;
-                })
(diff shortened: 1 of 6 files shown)


src/assets/js/menu/selectors.js
```diff
@@ -40,18 +40,15 @@ export default function() {
...
 
             var selectedIndicatorIndexes = options
-                .map(function(option, index) {
-                    return option.isSelected ? index : null;
-                })
-                .filter(function(option) {
-                    return option;
-                });
+                .reduce(function(array, option, index) {
+                    return array.concat(option.isSelected ? index : []);
```
jleft: Could give `array` a more meaningful name - maybe `selectedIndexes`?

Would this work? Looks a bit neater to me and avoids unnecessary concatenation.

```js
return option.isSelected ? selectedIndexes.concat(index) : selectedIndexes;
```

PR #721, opened 2016-04-04
Nav data sampled to only show 200 points

jdhodges:
#674. Replaces #679.


src/assets/js/chart/nav.js
```diff
@@ -132,24 +135,24 @@ export default function() {
     function nav(selection) {
         var model = selection.datum();
 
+        sampler.bucketSize(model.data.length / numberOfSamples);
```
jleft: Worth being defensive here to ensure we don't have a bucket size less than 1 (i.e. when `model.data.length < numberOfSamples`)?

`Math.max(model.data.length / numberOfSamples, 1)`

jleft:
One final comment, then LGTM :+1:

PR #722, opened 2016-04-04
Limited zoom to 5 periods - now working with discontinued weekends

rjmcneill:
Relates to #499 
Relates to #168 
Replaces and closes #525

### DIFF
--- a/src/assets/js/behavior/zoom.js
+++ b/src/assets/js/behavior/zoom.js
@@ -2,7 +2,7 @@ import d3 from 'd3';
 import fc from 'd3fc';
 import util from '../util/util';
 
-export default function(width) {
+export default function(containerWidth) {
 
     var dispatch = d3.dispatch('zoom');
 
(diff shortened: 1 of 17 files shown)


src/assets/js/chart/secondary/base.js
```diff
@@ -61,5 +64,13 @@ export default function() {
         zoomWidth = util.width(container.node()) - yAxisWidth;
     };
 
+    secondary.minimumVisiblePeriods = function(x) {
```
jdhodges: Don't think this is needed, `minimumVisiblePeriods` should still be in `model`.

src/assets/js/chart/secondary/base.js
```diff
@@ -35,6 +37,7 @@ export default function() {
             var zoom = zoomBehavior(zoomWidth)
               .scale(xScale)
               .trackingLatest(trackingLatest)
+              .minimumVisiblePeriods(minimumVisiblePeriods)
```
jdhodges: `.minimumVisiblePeriods(model.minimumVisiblePeriods)` should work.

src/assets/js/behavior/zoom.js
```diff
@@ -123,6 +130,14 @@ export default function(width) {
         return zoom;
     };
 
+    zoom.minimumVisiblePeriods = function(x) {
```
jdhodges: Don't need this either

Issue #720, opened 2016-04-01
Secondaries border looks too thick


rjmcneill:
Looks like some of the borders are about twice as thick as they should be.
Primary -> RSI and RSI -> MACD
![image](https://cloud.githubusercontent.com/assets/10117342/14207851/67159554-f814-11e5-9a30-400abadcfcaa.png)

PR #723, opened 2016-04-04
Rounded secondaries height to prevent antialiasing issues

jdhodges:
#720.


src/assets/js/util/layout.js
```diff
@@ -4,26 +4,27 @@ import fc from 'd3fc';
 var renderedOnce = false;
 
 export default function(containers, charts) {
+    var primaryContainerFlex = 5;
```
jdhodges: We could get this from the DOM if that's preferable.

Issue #729, opened 2016-04-06
Add testing around zoom


rjmcneill:
Refactor zoom to allow better testability and write unit tests

PR #730, opened 2016-04-06
Refactoring and testing of zoom

rjmcneill:
Relates to #729


jdhodges:
Would it be worth moving the zoom utils to util? We could put them in their own directory like we've done with the domain stuff.

Issue #733, opened 2016-04-06
Firefox - Product dropdown button moves when clicked


rjmcneill:
No screenshot for this really.
Just clicking on the dropdown will cause it to move slightly

PR #741, opened 2016-04-07
Added 0 padding to dropdown toggles to prevent move on click on firefox

jdhodges:
#733


jleft:
Thanks @jdhodges, but this will also change the positioning of the button.

Replaced by #796.

Issue #762, opened 2016-05-04
Navigator brush handle should not be cut at extreme ends


jleft:
**Steps to reproduce**

Navigate to the application in a desktop web browser. The default view should show the issue.

**Observed Behaviour**

The brush handle is cut:

![handle-cut](https://cloud.githubusercontent.com/assets/2017615/15007977/c08836b2-11d4-11e6-8f2b-678326bdba02.PNG)

**Expected Behaviour**

The brush handle should be fully visible.

PR #763, opened 2016-05-04
Allow brush handles to overflow

jdhodges:
#762

### DIFF
--- a/src/assets/styles/navigator.less
+++ b/src/assets/styles/navigator.less
@@ -9,6 +9,7 @@
   #navbar-container {
     min-height: @navigator-height;
     margin-right: 0 !important;
+    overflow: visible;
 
     .extent {
       fill: @secondary-orange;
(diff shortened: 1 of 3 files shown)


src/assets/styles/navigator.less
```diff
@@ -21,36 +22,50 @@
...
-        .outer-handle {
-          stroke: @primary-white;
-          fill: @primary-dark-blue;
-          shape-rendering: auto;
+    .cartesian-chart {
+      overflow: visible;
+
+      .plot-area-container {
+        .domain {
+          display: none;
```
jleft: Is this because of the axes outer ticks? If so, would setting the outer tick size to `0` for the navigator chart fix this? If so, I think I'd lean to this, as ideally we shouldn't be removing the domain line here.

E.g.

```js
fc.chart.cartesian(xScale, yScale)
    .xOuterTickSize(0)
    .yOuterTickSize(0);
```

src/assets/styles/variables.less
```diff
@@ -97,7 +97,7 @@
 @navigator-extent-colour: @secondary-orange;
 @navigator-colour: @primary-blue;
 @navigator-background-colour: @secondary-dark-blue;
-@navigator-background-stroke-width: 2px;
+@navigator-background-stroke-width: 1px;
```
jleft: Should `backgroundStrokeWidth` in nav.js also be updated to reflect this change?

Perhaps this change will make this comment and the associated behaviour irrelevant now `// should have been 2 * borderWidth, but for unknown reason it is incorrect in practice.` (in nav,js)? I'm not sure, but might be worth checking...

jleft:
Looking good :+1: just one more minor comment

PR #789, opened 2016-05-24
Update to d3fc v8.0.0

rjmcneill:
Relates to #779


jleft:
Rebased and merged in #792

Issue #790, opened 2016-05-24
Data gen dates (/times) are off by one hour


rjmcneill:
This makes the axis and the legend look like they don't match up correctly.
![image](https://cloud.githubusercontent.com/assets/10117342/15510037/ac1bb916-21cd-11e6-9900-5f2a91079aab.png)

jleft:
Fixed by #791

PR #791, opened 2016-05-24
Normalised data generator dates/times

rjmcneill:
Relates to #790

### DIFF
--- a/src/assets/js/data/generator/historic/feedAdaptor.js
+++ b/src/assets/js/data/generator/historic/feedAdaptor.js
@@ -1,3 +1,4 @@
+import d3 from 'd3';
 import fc from 'd3fc';
 
 export default function() {
@@ -11,9 +12,8 @@ export default function() {
 
     var dataGeneratorAdaptor = function(cb) {
         end.setHours(0, 0, 0, 0);
-        var millisecondsPerDay = 24 * 60 * 60 * 1000;
-        dataGenerator.startDate(new Date(end - (candles - 1) * millisecondsPerDay));
-
+        var startDate = d3.time.day.offset(end, -(candles - 1));
+        dataGenerator.startDate(startDate);
         var data = dataGenerator(candles);
         cb(null, data);
     };


src/assets/js/data/generator/historic/feedAdaptor.js
```diff
@@ -12,7 +12,10 @@ export default function() {
     var dataGeneratorAdaptor = function(cb) {
         end.setHours(0, 0, 0, 0);
         var millisecondsPerDay = 24 * 60 * 60 * 1000;
-        dataGenerator.startDate(new Date(end - (candles - 1) * millisecondsPerDay));
+        var startDate = (new Date(end - (candles - 1) * millisecondsPerDay));
+
+        startDate.setHours(0, 0, 0, 0);
```
jleft: This can lead to inconsistent behaviour depending on the DST transition in the range between the start and end dates.

For example:

```js
// When crossing the spring DST transition,
// the date will be something like this (one hour too far)
var startDate = new Date('Wed Nov 02 2016 23:00:00 GMT+0000 (GMT Standard Time)');

// Offsets -23 hours; this is wrong, the end date will be one day earlier than specified
startDate.setHours(0, 0, 0, 0);  // -> Wed Nov 02 2016 00:00:00 GMT+0000 (GMT Standard Time)

// When crossing the autumn DST transition,
// the date will be something like this (short by one hour)
var startDate = new Date('Sat Aug 23 2014 01:00:00 GMT+0100 (GMT Daylight Time)');

// Offsets -1 hour; this is correct, the end date will be correct
startDate.setHours(0, 0, 0, 0);  // -> Sat Aug 23 2014 00:00:00 GMT+0100 (GMT Daylight Time)
```

I'd consider using `d3.time.day.offset` to calculate the start date, it will handle the DST issues.

rjmcneill: Ah of course...
I'll sort that in the morning :)

src/assets/js/data/generator/historic/feedAdaptor.js
```diff
@@ -11,9 +12,8 @@ export default function() {
 
     var dataGeneratorAdaptor = function(cb) {
         end.setHours(0, 0, 0, 0);
-        var millisecondsPerDay = 24 * 60 * 60 * 1000;
-        dataGenerator.startDate(new Date(end - (candles - 1) * millisecondsPerDay));
-
+        var startDate = d3.time.day.offset(end, -candles);
```
jleft: `-(candles - 1)`

PR #797, opened 2016-06-10
Bump jQuery to 3.0.0

jleft:
This appears to be incompatible with Bootstrap's JS, so may not merge.


jleft:
Closing as this change is not compatible with Bootstrap 3.3.6.

It should be compatible with the next release (3.3.7), so may consider this change again if/when this release happens.

